PR #1, opened 2015-03-05
Fix install command line

dspe:
There is a wrong command on INSTALL.md. Just replacing ezpublish-community by ezplatform


andrerom:
+1

bdunogier:
Thank you ;)

PR #5, opened 2015-03-19
Nginx : Allow to configure upload max size

yannickroger:
## Decription
See: http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

Ran into this issue while trying to upload a big file on the Platform UI

## Test
Manual test


andrerom:
+1

bdunogier:
+1

PR #6, opened 2015-03-25
Update INSTALL.md

technolabtips:
Add the case if EZP archive does not contain the required libraries in vendor directory.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -32,6 +32,14 @@
        php -d memory_limit=-1 composer.phar run-script post-install-cmd
        ```
 
+       *Note: The post install scripts will fail if the archive does not contain the required libraries in vendor directory.*
+       *Then execute install command*
+
+       ```bash
+       cd /<directory>/
+       php -d memory_limit=-1 composer.phar install --no-dev
+       ```
+
 
     B. **Composer**
 


andrerom:
hmm, you got this when downloading from github didn't you?

On one side I'm I would almost suggest we rather tag with vendors, but that would probably break create-project... Someone tried this? (cc @bdunogier @lolautruche )

Alternative we [talked about changing install doc to always do update on install](https://github.com/ezsystems/ezpublish-community/pull/232#issuecomment-76155059), however there are some small differences in post-install-cmd and post-update-cmd we should then re consider before we can do that.

technolabtips:
Yes, it's when i'm downloading from github.
I agree with you on the use of composer.
Maybe we can create a checker script (as ezpublish/checker.php) to ensure that all components is OK.
What do you think about  ?

andrerom:
Hi,

that is one option indeed, briefly discussed it with @bdunogier and came up with solution before we consider starting to abstract composer by:
- Improve doc, see #7 *(please review if you can)*
- improve text on tags to either not be github `releases`, or always refer to install text, see: https://github.com/ezsystems/ezplatform/releases/tag/v0.5.0

Would these improvements also solve the issue from your perspective?

PR #7, opened 2015-03-26
Improve install to clarify archive & composer use

andrerom:
- Specifies that archives are the ones from share.ez.no (as tags don't contain vendors, and are for composer use)
- Also improve & simplify the composer create-project instructions

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,7 +20,7 @@
 
 1. **Get eZ Platform**:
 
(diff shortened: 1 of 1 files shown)


bdunogier:
+1

andrerom:
@technolabtips you're ok with this as well?

technolabtips:
I'm OK

PR #8, opened 2015-04-08
Update INSTALL.md

mzbusiness:
I needed to set php memory_limit=256M , default value for php 5.4 is 128M and ezplatform::install runs out of memory

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,6 +15,7 @@
 
   These instructions assume you have strong technical knowledge and have already installed PHP, web server & a database server with a corresponding clean database needed for this software.
   For further information on requirements see: https://doc.ez.no/display/EZP/Requirements
+  *Note: set php.ini memory_limit=256M before running ezplatform::install command below
 
 ## Install
 


andrerom:
could be on the command instead, like other places here

bdunogier:
It is most likely caused by the container's compilation, not by the script itself (it doesn't do anything that really uses memory).

We usually add the option to the command line directly in the instructions, we might want to do it here as well.

joaoinacio:
Nitpick: all composer command examples use `php -d memory_limit=-1 ...`, IMHO it would probably make sense to use the same value for consistency.
(That and the possibility of 256M not being sufficient in the future for some unknown reason)

mzbusiness:
other issue :
after installing I got error :   request.CRITICAL: Uncaught PHP Exception eZ\Bundle\EzPublishCoreBundle\ApiLoader\Exception\InvalidRepositoryException: "Undefined repository ''. Did you forget to configure it in ezpublish_*.yml?" at ...

After googling I found this related issue , 
https://jira.ez.no/browse/EZP-24025

After running clear:cache --env prod , ezpublish worked

mzbusiness:
by the way , running php -d memory_limit=-1 ...
also runs out of memory, only manage to pass it after using 'none'  in the last question

bdunogier:
That's really odd... the only explanation I can see would be that the process that runs out of memory isn't the same one than the initial one. @lolautruche ?

andrerom:
@mzbusiness Which --env did you use for install command?

mzbusiness:
I use prod env

andrerom:
Ok, so we don't clear enough cache in the install command then.

andrerom:
Reproduced and hopefully fixed the cache issue: https://jira.ez.no/browse/EZP-24158

Remaining open question is why `-d memory_limit=-1` didn't work for you.

andrerom:
Returning to this question, I'm going to guess the issue is because we / symfony implicit call another php process via `static::executeCommand`, so the use of `-d memory_limit=-1` is not carried on to the sub process.

Two options:
- find a way to pass it on to sub process
- remove use of `-d memory_limit=-1` and rather accept this PR.

andrerom:
Opened PR against Symfony: https://github.com/symfony/symfony/pull/15153

However since this will take some time to trickle into Symfony releases, and also Composer releases which bundles this component, and thus blocks the local file in vendor form being used, I vote for also merging this for now.

+1

bdunogier:
Given this analysis, yes, this patch is a good workaround.

+1 then.

andrerom:
Thanks for the PR @mzbusiness, all clear, all merged :)

PR #10, opened 2015-04-14
EZP-24150: Implement privacy cookie law in demo bundle

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-24150

More info: https://github.com/ezsystems/DemoBundle/pull/149


andrerom:
+1

PR #14, opened 2015-05-01
Fix EZP-24158: Need to clear cache manually after platform install

andrerom:
Testing if https://github.com/ezsystems/ezpublish-kernel/pull/1244 also can remove need for manual cache clar on travis, hence be tested by behat tests.

*This cleans up some issues that was created when we introduced installer here in terms of having to do custom stuff for behat which with the kernel change here and the one in ezpublish-kernel is no longer needed.*

- [x] amend away composer.json change when kernel pr has been merged

### DIFF
--- a/bin/.travis/prepare_ezpublish.sh
+++ b/bin/.travis/prepare_ezpublish.sh
@@ -5,6 +5,10 @@
 echo "> Setup github auth key to not reach api limit"
 ./bin/.travis/install_composer_github_key.sh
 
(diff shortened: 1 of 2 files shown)


andrerom:
ready for review @crevillo, @lolautruche @bdunogier

crevillo:
+1

bdunogier:
+1

andrerom:
Thanks guys, enjoy your weekend :)

PR #17, opened 2015-05-20
Update INSTALL.md to install LegacyBridge after EZPlatform Install

ksalama:
Hi, 
I propose to move the optional step of legacy bridge because when composer add ezsystems/legacy-bridge, the process need run a ```bash cache:clear ```.

But at this time, the ezpublish.yml file is not created. This file appears after the execution of ```bash ezplatform:install ```

K.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -54,8 +54,6 @@
 
 2. *Only for *NIX users* **Setup folder rights**:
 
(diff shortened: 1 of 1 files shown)


andrerom:
well, depends on if user intends to use setup wizard.

andrerom:
As we don't technically support legacy bridge with platform anymore *(as planned)*, this is closed.

PR #20, opened 2015-05-27
Use global namespaces in EzPublishKernel

joaoinacio:
See ~~https://github.com/ezsystems/ezpublish-community/pull/172~~ [ezpublish-community/pull/243](https://github.com/ezsystems/ezpublish-community/pull/243)

### DIFF
--- a/ezpublish/EzPublishKernel.php
+++ b/ezpublish/EzPublishKernel.php
@@ -7,41 +7,8 @@
  * @version //autogentag//
  */
 
-use Egulias\ListenersDebugCommandBundle\EguliasListenersDebugCommandBundle;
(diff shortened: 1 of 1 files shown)


andrerom:
+1 from my side, just trying to think if there is anything but LegacyBridge install instructions we need to update. ping @bdunogier @lolautruche

joaoinacio:
@andrerom : actually, AFAICT the docs are already OK:
```
Edit ezpublish/EzPublishKernel.php, and add
 new eZ\Bundle\EzPublishLegacyBundle\EzPublishLegacyBundle( $this ) 
at the end of the $bundles array. 
```

bdunogier:
+1 :fire:

lolautruche:
+1

PR #25, opened 2015-06-24
Remove nelmio/cors-bundle composer dependency, as it is already a sub…

guillaumelecerf:
… requirement


guillaumelecerf:
BTW why is `composer.lock` versionned ?

andrerom:
+1

as for reasoning, see kernel PR description, and in addition faster CI builds, and faster composer create-project.

PR #27, opened 2015-07-01
[Travis/Composer] Remove composer.lock and prefer ip4 on travis to av…

andrerom:
…oid packagist issues

ref: https://github.com/composer/composer/issues/4142#issuecomment-117648869


michalpipa:
Big :+1: for removing composer.lock

andrerom:
Job just restarted, should pass now with fix from @lolautruche and @bdunogier so we can get rid of composer.lock so we can go back to testing master for behat tests.

lolautruche:
+1

bdunogier:
+1

andrerom:
Thanks, @michalpipa please cherry-pick this on studio as well :)

michalpipa:
@andrerom Great!

PR #32, opened 2015-08-03
Improve webserver setup instructions

gggeek:
english improvements; mention ExecCGI for Apache; move nginx instructions to .md format; format Varnish instructions same as other ones


gggeek:
PS: One thing which I feel is currently under documented is the creation of the database schema.
I think it could be detailed in install.md as a dedicated step - without going as far as including the sql for a given type of db, but mentioning that the character set has to be utf8.
The reasoning is that:
- currently it is easy to miss in the docs the mention that the db has to be created already
- it is more or less the only manual step which is not documented, if we assume that the user starts with a plain lamp stack

As an alternative (or as addition): put the sql for db schema creation in doc/, next to the vhost examples...

yannickroger:
Regarding the nginx documentation being in RST, it was because we were planning on using readthedocs at some point and it would only support RST.
ping @mere-teresa  @andrerom

doc/apache2/Readme.md
```diff
@@ -36,12 +36,12 @@ Configuration
 
 #### Pure legacy setup
 
-In eZ Publish 5.x you can optionally set it up to only use legacy, this is useful for 4.x upgrades which only uses legacy.
-*Warning: By doing this, absolutely no integrations between legacy and Platform works, so you can not use any of the Platform features (API, HttpCache, Symfony, ..) in this setup.*
+eZ Publish 5.x can optionally be set up to only use the legacy kernel. This is useful for 4.x upgrades which only use legacy.
```
yannickroger: Should we also rename eZ Publish 5.x to eZ Platform ?

yannickroger:
+1 (with my comments sorted out ;) )

doc/apache2/vhost.template
```diff
@@ -14,12 +14,13 @@
     # LogLevel debug
 
     <Directory %BASEDIR%/web>
+        # If using php configured in FastCGI mode, you will also need to add "ExecCGI" to the line below
```
andrerom: do we? this is used directly for Behat tests on travis, and seems to work w/o there.

andrerom:
@gggeek would prefer if it is kept as rst for now, and that we do that choice separately, it also possible affect links.

doc/apache2/vhost.template
gggeek: It did not work without it on my setup: apache 2.4 + mod_fcgid. I did not test using plain mod_fastcgi, but according to google it does need it as well (at least it says so in the Arch Linux wiki).
Are you sure about the Travis Apache config?

gggeek:
About rst: I'd personally prefer to change it now (for uniformity) and change it back when/if all of the doc changes format. But I see your point about links. Undoing

doc/apache2/vhost.template
andrerom: It's probably 2.2 and not 2.4 as travis uses old version of Ubuntu since they are afraid of breaking stuff for people.

you can see what is used here:
- https://github.com/ezsystems/ezplatform/blob/master/bin/.travis/configure_apache2.sh
- https://github.com/ezsystems/ezplatform/blob/master/bin/.travis/apache2/php5-fcgi

gggeek:
While at it, renamed eZ Publish to eZ Platform in all doc occurrences

doc/apache2/vhost.template
```diff
@@ -164,14 +162,14 @@
...
+        <LocationMatch "^/share/icons/.*">
+            # Icons as used by admin interface, barely change
             ExpiresActive on
             ExpiresDefault "now plus 7 days"
         </LocationMatch>
 
-        <LocationMatch "^/share/icons/.*">
-            # Icons as used by admin interface, barly change
+        <LocationMatch "^/design/[^/]+/(stylesheets|images|javascripts?|lib|flash)/.*">
+            # Same as above for elements bundled eZ Publish designs
```
andrerom: could actually remove all legacy stuff here

however you are starting to do to many things in same PR as usual, so if further legacy removal stuff is found, do it in separate PR.

gggeek:
PS: changed the comment mentioning need for ExecCgi to 'might need'. I can not test on windows using fastcgi as the dlls have vanished off the net, and do not have fcgid.dll for apache 2.2, either.

Will squash the commits when we agree that all is good.

doc/apache2/vhost.template
gggeek: It was in my todo list to test existance of legacy dirs.

Any idea about how to split this PR in a meaningful way?

andrerom: followup

andrerom:
+1, sorry, should have pinged you for squash, but ok.

doc/apache2/vhost.template
gggeek: will do

PR #34, opened 2015-08-11
Switch to Selenum2 driver for Behat to get newer packages

andrerom:
Sahi packages are mostly not maintained so it's the reason why we get old
packages for Mink and so on now, fix by attempting to switch to Selenium once and for all.

This change gives us the following package updates:

```bash
  - Removing fabpot/goutte (v1.0.7)
  - Installing fabpot/goutte (v2.0.4)

  - Removing behat/mink (v1.5.0)
  - Installing behat/mink (v1.6.1)

  - Removing behat/mink-browserkit-driver (v1.1.0)
  - Installing behat/mink-browserkit-driver (v1.2.0)

  - Removing behat/mink-goutte-driver (v1.0.9)
  - Installing behat/mink-goutte-driver (v1.1.0)

  - Removing instaclick/php-webdriver (1.0.17)
  - Installing instaclick/php-webdriver (1.4.3)

  - Removing behat/mink-selenium2-driver (v1.1.1)
  - Installing behat/mink-selenium2-driver (v1.2.0)
```

Todo
- [x] Remove Sahi and only use selenium2 to be able to get updated packages
- [x] Fix deprecated use of Mink API
- [x] Identify and fix reaming scenario failing, most likely because of the named => named_partical change in BehatBundle  / alternative: login link failing is loaded asynchronously, could be selenium is not waiting long enough for it.


andrerom:
@pcardiga / @miguelcleverti : One scenario still failing, any clue?
https://travis-ci.org/ezsystems/ezplatform/jobs/75128354#L902

andrerom:
I can confirm locally that I get the same error, and if I rerun the test it passes.
So it basically fails because this login menu is to slow on first load when it is not cached.

andrerom:
:fireworks:  passes :)

PR #35, opened 2015-08-30
Small update to directory permissions to align with Sf 2.8

andrerom:
Just small update now to align with symfony 2.8.

~~~Background: Recent discussions with dpobel on permissions, and PM request to make recording of install process made me think we should simplify directory permissions greatly, by using a script we can theoretically make it more predictable and less error prone. dpobel I didn't remember what you were voicing as your preference on this, I think it was making sure both command line and www user has rights but others have limited rights.~~~

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -59,49 +59,60 @@
        - For core development: Add '--prefer-source' to get full git clones, and remove '--no-dev' to get things like phpunit and behat installed.
        - Further reading: https://getcomposer.org/doc/03-cli.md#create-project
 
-2. *Only for *NIX users* **Setup folder rights**<a name="install-2-folder-rights"></a>:
+2. **Setup folder rights**<a name="install-2-folder-rights"></a>:
 
        Like most things, [Symfony documentation](http://symfony.com/doc/current/book/installation.html#checking-symfony-application-configuration-and-setup)
-       applies here, meaning `app/cache` and `app/logs` need to be writable by cli and web server user. Furthermore, future
-       files and directories created by these two users will need to inherit those access rights. *For security reasons,
-       there is no need for web server to have access to write to other directories.*
+       applies here, in this case meaning `app/cache`, `app/logs` and `web` need to be writable by cli and web server user.
+       Furthermore, future files and directories created by these two users will need to inherit those access rights. *For
+       security reasons, in production there is no need for web server to have access to write to other directories.*
 
-       Change `www-data` to your web server user:
+       For development setup you may change your web server config to use same user as owner of folder, what follows
+       below are mainly for production setup, and like Symfony we first and foremost recommend using an approach using ACL.
 
        A. **Using ACL on a *Linux/BSD* system that supports chmod +a**
 
+       Some systems, primarily Mac OS X, supports setting ACL using a `+a` flag on `chmod`. Example uses a command to
+       try to determine your web server user and set it as ``HTTPDUSER``, alternatively change to your actual web server
+       user if non standard:
+
        ```bash
        $ rm -rf app/cache/* app/logs/*
-       $ sudo chmod +a "www-data allow delete,write,append,file_inherit,directory_inherit" \
-         app/cache app/logs web
-       $ sudo chmod +a "`whoami` allow delete,write,append,file_inherit,directory_inherit" \
-         app/cache app/logs web
+       $ HTTPDUSER=`ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1`
+       $ sudo chmod +a "$HTTPDUSER allow delete,write,append,file_inherit,directory_inherit" app/cache app/logs  web
+       $ sudo chmod +a "`whoami` allow delete,write,append,file_inherit,directory_inherit" app/cache app/logs web
        ```
 
        B. **Using ACL on a *Linux/BSD* system that does not support chmod +a**
 
-       Some systems don't support chmod +a, but do support another utility called setfacl. You may need to enable ACL support on your partition and install setfacl before using it (as is the case with Ubuntu), like so:
+       Some systems don't support chmod +a, but do support another utility called setfacl. You may need to
+       [enable ACL support](https://help.ubuntu.com/community/FilePermissionsACLs) on your partition and install setfacl
+       before using it *(as is the case with Ubuntu)*. With it installed example below uses a command to try to determine
+       your web server user and set it as ``HTTPDUSER``, alternatively change to your actual web user if non standard:
 
        ```bash
-       $ sudo setfacl -R -m u:www-data:rwx -m u:`whoami`:rwx \
-         app/cache app/logs web
-       $ sudo setfacl -dR -m u:www-data:rwx -m u:`whoami`:rwx \
-         app/cache app/logs web
+        $ HTTPDUSER=`ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1`
+        $ sudo setfacl -R -m u:"$HTTPDUSER":rwX -m u:`whoami`:rwX app/cache app/logs web
+        $ sudo setfacl -dR -m u:"$HTTPDUSER":rwX -m u:`whoami`:rwX app/cache app/logs web
        ```
 
        C. **Using chown on *Linux/BSD/OS X* systems that don't support ACL**
 
-       Some systems don't support ACL at all. You will need to set your web server's user as the owner of the required directories.
+       Some systems don't support ACL at all. You will need to set your web server's user as the owner of the required
+       directories, in this setup further symfony console commands against installation should use the web server user
+       as well to avoid new files being created using another user.  Example uses a command to try to determine your
+       web server user and set it as ``HTTPDUSER``, alternatively change to your actual web server user if non standard:
 
        ```bash
-       $ sudo chown -R www-data:www-data app/cache app/logs web
+       $ HTTPDUSER=`ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1`
+       $ sudo chown -R "$HTTPDUSER":"$HTTPDUSER" app/cache app/logs web
        $ sudo find {app/{cache,logs},web} -type d | xargs sudo chmod -R 775
        $ sudo find {app/{cache,logs},web} -type f | xargs sudo chmod -R 664
        ```
 
        D. **Using chmod on a *Linux/BSD/OS X* system where you can't change owner**
 
-       If you can't use ACL and aren't allowed to change owner, you can use chmod, making the files writable by everybody. Note that this method really isn't recommended as it allows any user to do anything.
+       If you can't use ACL and aren't allowed to change owner, you can use chmod, making the files writable by everybody.
+       Note that this method really isn't recommended as it allows any user to do anything.
 
        ```bash
        $ sudo find {app/{cache,logs},web} -type d | xargs sudo chmod -R 777
@@ -122,6 +133,7 @@
        write access to the following directories:
        - app/cache
        - app/logs
+       - web
 
 
 3. **Configure a VirtualHost**<a name="install-3-vhost"></a>:
--- a/bin/.travis/prepare_ezpublish.sh
+++ b/bin/.travis/prepare_ezpublish.sh
@@ -9,8 +9,8 @@ echo "> Setup github auth key to not reach api limit"
 cp bin/.travis/composer-auth.json ~/.composer/auth.json
 
 echo "> Set folder permissions"
-sudo find {app/{cache,logs},web} -type d | sudo xargs chmod -R 777
-sudo find {app/{cache,logs},web} -type f | sudo xargs chmod -R 666
+sudo find {app/{cache,logs},web} -type d | xargs sudo chmod -R 777
+sudo find {app/{cache,logs},web} -type f | xargs sudo chmod -R 666
 
 echo "> Copy behat specific parameters.yml settings"
 cp bin/.travis/parameters.yml app/config/


bin/ezdirperm.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+# Script to fix folder and file permissions on places www-user needs write access
+
+[ ! -f "bin/ezdirperm.sh" ] && echo "\033[31mThis script has to be run the root of the installation\033[0m" && exit 1
```
glye: in/from the root...?

INSTALL.md
```diff
@@ -57,59 +57,35 @@
...
-       C. **Using chown on systems that don't support ACL**
-
-       Some systems don't support ACL at all. You will either need to set your web server's user as the owner of the required directories.
+       Example on Mac OS X:
 
        ```bash
-       $ sudo chown -R www-data:www-data ezpublish/{cache,logs,config,sessions} web
-       $ sudo find {ezpublish/{cache,logs,config,sessions},web} -type d | xargs sudo chmod -R 775
-       $ sudo find {ezpublish/{cache,logs,config},web} -type f | xargs sudo chmod -R 664
+       $ sudo bin/ezdirperm.sh -u _www
```
lolautruche: Maybe specify different possibilities for the user as it really depends on how you installed your webserver.

bin/ezdirperm.sh
```diff
@@ -0,0 +1,82 @@
...
+
+print_usage()
+{
+    echo ""
+    echo -e "\033[36mHelp text\033[0m"
+    echo "Fixes permissions on files folders that need write access on *nux systems by setting provided <www-user> as owner"
```
lolautruche: `*nix`

dpobel:
my point was more to get rid of the acl as this is complicating things while traditional Unix permission are more than enough. Then the script could just do:
```
chown -R $WWW_USER ezpublish/{cache,logs,config,sessions} web
chmod -R og+rwX ezpublish/{cache,logs,config,sessions} web
```

yannickroger:
To my point of view, ACL should not be documented (and scripted) as the default install as it is overkill on most installs.

andrerom:
@dpobel / @yannickroger Interesting, but then the console/cron user should be documented to impersonate www user, or? To issues when folders gets created with cron user as owner, while this *should* not be a problem in theory, with legacy Users frequently got into issues because folders where not owned by www user anymore.

@bdunogier / @glye / @lolautruche / @gggeek Your ok with @dpobel's proposal? If so we can skip the script, one less thing to learn. Point was anyway to simplify permissions.

dpobel:
indeed we should instruct devs to run the script as the same user as the web server or as a tip document how to run the web server under the dev user on a local dev environment.

glye:
Fine with me if this approach (console/cron impersonates www) is considered an acceptable practice from sysadmin point of view. Another alternative is to setup group permissions so that www/console/cron can edit each others files. ACL is more complex but perhaps "best practice"?

lolautruche:
Fine by me

bdunogier:
> indeed we should instruct devs to run the script as the same user as the web server or as a tip document 
I strongly second that. For development environments, it really, really makes things much easier.

I think that having to run the scripts as the www-user is what people do, and have to do anyway. Maybe we should clarify it in the doc as well ?

andrerom:
Second try, ping @dpobel @yannickroger @glye @bdunogier @lolautruche

lolautruche:
+1

INSTALL.md
```diff
@@ -57,54 +57,22 @@
        - For core development: Add '--prefer-source' to get full git clones, and remove '--no-dev' to get things like phpunit and behat installed.
        - Further reading: https://getcomposer.org/doc/03-cli.md#create-project
 
-2. *Only for *NIX users* **Setup folder rights**:
+2.  *Only for *NIX/OS X users* **Setup folder rights**:
 
-       *Skip to step #3 if you plan to install LegacyBridge and eZ Publish Legacy with this installation!*
+       *Skip to step #3 if you plan to install LegacyBridge and eZ Publish Legacy with this installation.*
```
glye: When I preview this, header count starts from 1, not 0, so this is step 3, and it should say "skip to step 4". Not sure if bug in richtext rendering on github.

INSTALL.md
```diff
@@ -57,54 +57,22 @@
...
+       Change `www-data` to your web server user, and `cmd-group` to a group command line users is part of:
 
        ```bash
-       $ rm -rf ezpublish/cache/* ezpublish/logs/* ezpublish/sessions/*
-       $ sudo chmod +a "www-data allow delete,write,append,file_inherit,directory_inherit" \
-         ezpublish/{cache,logs,config,sessions} web
-       $ sudo chmod +a "`whoami` allow delete,write,append,file_inherit,directory_inherit" \
-         ezpublish/{cache,logs,config,sessions} web
+       $ chown -R www-data:cmd-group ezpublish/{cache,logs,config,sessions} web
+       $ chmod -R ug+rwX ezpublish/{cache,logs,config,sessions} web
```
glye: Sorry if my CLI-fu is weak, but this doesn't change ownership or permissions on dirs created after this point, does it? Don't we have to change umask settings to solve that?

INSTALL.md
```diff
@@ -57,54 +57,22 @@
...
-       One common issue is that the `ezpublish/cache`, `ezpublish/logs` and `ezpublish/config` directories **must be writable both by the web server and the command line user**.
-       If your web server user is different from your command line user, you can run the following commands just once in your project to ensure that permissions will be set up properly.
+       One common issue is that the `ezpublish/{cache,logs,config},web` directories **must be writable both by the web server and the command line user**.
+       To make sure both have access, you can run the following commands in your `<root-dir>` to ensure that permissions will be set up properly.
 
-       Change `www-data` to your web server user:
+       *Before you begin, make sure you have a user group assigned to your command line user that can be safely used to set rights*
 
-       A. **Using ACL on a system that supports chmod +a**
+       Change `www-data` to your web server user, and `cmd-group` to a group command line users is part of:
```
andrerom: Ok, but maybe "group command line users" => "user group command line users"

INSTALL.md
andrerom: @dpobel ?

INSTALL.md
glye: Yes, that's good. Oh, and "are part of", not is.

INSTALL.md
andrerom: ok

INSTALL.md
dpobel: not sure about the `cmd-group`, maybe we can even omit that. Normally, those directories already belong to the current user, so we don't really have to fix that.

andrerom: but we are changing owner here aren't we? that is the nature of chown as opposed to setfacl.

INSTALL.md
```diff
@@ -57,67 +57,37 @@
...
 
-2. *Only for *NIX users* **Setup folder rights**:
+3. *Only for *NIX/OS X users* **Setup folder rights**:
 
-       *Skip to step #3 if you plan to install LegacyBridge and eZ Publish Legacy with this installation!*
+       *Skip to step #4 if you plan to install LegacyBridge and eZ Publish Legacy with this installation.*
 
-       One common issue is that the `ezpublish/cache`, `ezpublish/logs` and `ezpublish/config` directories **must be writable both by the web server and the command line user**.
-       If your web server user is different from your command line user, you can run the following commands just once in your project to ensure that permissions will be set up properly.
+       One common issue is that the `ezpublish/{cache,logs,config,sessions},web` directories **must be writable both by the web server and the command line user**.
```
andrerom: Actually, `config` directory does not currently need write access by web user does it? As there is currently no web based setup wizard. / @bdunogier

andrerom:
Third round ready for review, @glye looks ok? *besides the open questions for @bdunogier and @dpobel*

INSTALL.md
glye: Yes, umask can set the permissions on future dirs but not the ownership. So one way is to make sure there is a user group containing both the apache user and the cli/cron user, and set umask so that read/write is always set on the group level (which is default anyway, I believe). Then recursively set this user group as group owner of the ezp root directory, and use `setgid` on the ezp root to ensure all future dirs/files have the same group owner.
http://superuser.com/questions/437781/how-do-i-set-default-group-ownership-for-files-in-a-directory
http://www.omnisecu.com/gnu-linux/redhat-certified-engineer-rhce/linux-setuid-and-setgid.php
"when a user creates a file in a setgid directory, the group ownership of the file is set to the group owner of the directory"

glye:
+1 besides open questions, yes.

bdunogier:
> Actually, config directory does not currently need write access by web user does it?

Nope, it does not.

dpobel:
+1

andrerom:
@bdunogier removed that then

ALSO removed the whole group thing and back to what @dpobel proposed basically as I realized it was much more relevant to add inline doc that we assume existing group set on the directories is what you actually want *(the user that created the install)*. This further greatly simplifies the step :tada:

vidarl:
Honestly, I like the ACLs.... After we started using them, I have never had issues with permissions...
This PR basically reverts to how things where before ( and which only caused problems.... ) ?

bdunogier:
Note that unless I'm mistaken, we have seen recent exchanges about permission related issues. I'm not gonna say I agree with @vidarl on ACL are "likable", but a still part of me still believes there are situations where they are the best solution.

andrerom:
I'm ok with that, I have had same experience with ACL (setfacl) as @vidarl, never had any permission based issues after it was added and I started using it.

So then the current proposal will be the main recommended way, and `setfacl` is mentioned as a alternative advance example. Hence keeping it simple for beginners.

@bdunogier @vidarl ok?

vidarl:
IMO, we should not say that lazy-people/beginners/developers may do it this way ( without ACL ), but that the ACL method is the correct one and should be used in production.

As far as I can tell, the "chmod -R ug+rwX" method will not work flawlessly unless console user *always* prefix his commands with "sudo -u www-data  ...".
And that implies that sudo is configured correctly ( which I guess is more difficult than using ACLs in the first place )

gggeek:
Well, I have both my console user and cron users part of the www-data group, and I never have any trouble nor need for sudo. Of course, it takes some time before you get it right the first time...

andrerom:
@gggeek So then www-data group is primary group for both right? Otherwise you'll still have problems.

gggeek:
yes

bdunogier:
> So then the current proposal will be the main recommended way, and setfacl is mentioned as a alternative advance example. Hence keeping it simple for beginners.

It works for me.

INSTALL.md
```diff
@@ -66,67 +61,38 @@
...
+       $ chown -R www-data ezpublish/{cache,logs,sessions}
+       $ chmod -R ug+rwX ezpublish/{cache,logs,sessions}
        ```
 
-       D. **Using chmod**
+       Note: *User group is on purpose not specified as it is assumed this is set to current command line users group that will continue to need access.
+       To make this work flawlessly you'll need to add web server user group as primary group to your command line user, to make sure it has access to files created by web user and vica-versa.*
 
-       If you can't use ACL and aren't allowed to change owner, you can use chmod, making the files writable by everybody. Note that this method really isn't recommended as it allows any user to do anything.
+       Note2: *On Linux it is also possible to use ACL to setup more advance rights if you are more comfortable using that.*
```
bdunogier: "more advance*d*"

bdunogier:
+1

vidarl:
But in my opinion, the proposed install.MD is faulty.
It do not instruct the user to use explicit change UID to www user ( by sudo or other means) before issuing any console command.
So a sentence instructing the user to do so is lacking

INSTALL.md
vidarl: And here we should insert the "+a" and "setfacl" which has been removed IMO

andrerom:
@dpobel @vidarl can you bash guys bash it out and decide on this, my aim is just to simplify where possible, so I'll let you deal with the semantics.

INSTALL.md
andrerom: @vidarl then the whole point of this PR is moot, as there is no simplifications done.

andrerom:
Updated: https://github.com/ezsystems/ezplatform/blob/dir_permissions/INSTALL.md#install

Took some parts from Symfony doc:
- was not so keen on documenting the umask approach
- chown approach has some open questions, for instance is `chmod g+s` universally supported to handle new files/folder group inheritance?

bdunogier:
Since we are now using `app` instead of `ezpublish`, do we still need our own instructions ? Or maybe they can be trimmed down ?

This is what symfony uses: http://symfony.com/doc/2.7/book/installation.html#book-installation-permissions.

bdunogier:
Ping @andrerom

bdunogier:
Rebased to make it mergeable (adapted `ezpublish/` strings).

andrerom:
hmm, well:
- Point 1 seems insecure, shouldn't it mentioned that cli/www user should thus only have write access to the writable folders?
- Point 4 is rather hackish.

So, of course if we manage to contribute to their doc something that is a bit more bullet proof I 'm all for reusing their doc.

andrerom:
PS: your rebase is not really complete text wise :P
       And besides VL and DP never managed to agree on anything.

bdunogier:
> your rebase is not really complete text wise

What's missing ? I did check block by block that I had not lost text. I might have made a mistake of course.

andrerom:
The one I noticed is that Windows is twice ;)

bdunogier:
Removed the duplicate.

INSTALL.md
```diff
@@ -59,69 +59,51 @@
...
        ```
 
-       E. **Setup folder rights on Windows**
+       C. **Setup folder rights on Windows**
 
        For your choice of web server you'll need to make sure web server user has read access to `<root-dir>`, and
        write access to the following directories:
        - app/cache
        - app/logs
+       - app/sessions *Unless you choose to configure session to be stored by another system then file system*
```
andrerom: wrong here and above, afaik session folder was removed some commits ago.
Same goes for all the changes above, they should all prefer changes in master and reduce the diff heavily.

bdunogier:
Hmm ok. If you still have your branch, feel free to overwrite this one. No need to keep my messed up attempt to revive this PR.

andrerom:
Updated re syncing with master and re applying approach form symfony, aiming rather to do further  simplifications once we move to Symfony 3.0 structure.


For reference the old change can be seen on 5ed30d67ef9fd002bb0d1f1f0afefdb5a5a73047.

andrerom:
Small gotcha caused by my last update
```bash
> Set folder permissions
find: `web/var': No such file or directory
find: `web/var': No such file or directory
```

I'll change back to `web` for now.

vidarl:
It would be nice if the web/var folder could be part of dist ( either placed in git ( would require a web/var/.keep file or so ), or the directory could maybe be created by composer ? )

crevillo:
i'd vote for the web/var/.gitkeep option as we have with src. (even dunno if that .gitkeep is necessary since we had AppBundle there...)

andrerom:
@bdunogier Do you see any reason not to add it to dist?

andrerom:
> i'd vote for the web/var/.gitkeep option as we have with src. (even dunno if that .gitkeep is necessary since we had AppBundle there...)

AppBundle was not always there, and will maybe not always be there. In the base distribution that we might be able to distribute via docker, this folder will be empty, and content generated on install for instance or manually added by end user.

crevillo:
ok, then i said nothing.

andrerom:
Well, you did say you would also like us to add web/var/.[git]keep :)

crevillo:
true that :).

bdunogier:
> It would be nice if the web/var folder could be part of dist ( either placed in git ( would require a web/var/.keep file or so ), or the directory could maybe be created by composer ? )

I don't mind `web/var/.keep`.

bdunogier:
And about `src/.gitkeep`, yes, we could remove it. [symfony-standard](https://github.com/symfony/symfony-standard/tree/master/src) doesn't have one anymore.

andrerom:
Amending then. *(in regards to web/var/.keep)*

andrerom:
> I don't mind web/var/.keep.

Side: No issues for cluster in regards to mounting in nfs folder and so on right?

vidarl:
A mount point can contain files, yes. Such files will not be visible after the mount though ( but will be visible again after a umount ).
At least, this how it works on Linux.

andrerom:
as is review ping @bdunogier @vidarl @crevillo

crevillo:
+1.

andrerom:
Thanks @crevillo

PR #36, opened 2015-08-30
EZP-24745: Make SolrBundle ready for release

andrerom:
https://jira.ez.no/browse/EZP-24745

review ping @pspanja @bdunogier @lolautruche 

Todo:
- [x] update installer ezpublish.yml config
 - [x] https://github.com/ezsystems/ezpublish-kernel/pull/1389
 - [x] https://github.com/ezsystems/DemoBundle/pull/177
- Bugs encountered:
 - [x] Console command default siteaccess https://github.com/ezsystems/ezpublish-kernel/pull/1395
 - [x] Instance of almighty content type with invalid data on it and without location 
- [x] Docker setup https://github.com/ezsystems/ezpublish-docker/pull/40
- [x] DOC: INSTALL.md and https://doc.ez.no/display/EZP/Search+Engines

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,12 +17,21 @@
 
 ## Install
 
-0. **Create a clean database**
+0. **Create an empty database**, and optionally setup Solr
 
     The following step will ask you for credentials/details for which database to use, so make sure to create one first.
-    *Note: Right now installer only supports MySQL, Postgres support should be re added in one of the upcomming releases.*
+
+    *Optional: At this point you can also setup Solr to be used by eZ Platform*
+
+    *Note: Right now installer only supports MySQL, Postgres support should be (re)added in one of the upcoming releases.*
 
 1. **Get eZ Platform**:
+
+    There are two ways to install eZ Platform described below, what is common is that you should make sure
+    relevant settings are generated into `ezpublish/config/parameters.yml` as a result of this step.
+
+    `parameters.yml` contains settings for your database, mail system, and optionally [Solr](http://lucene.apache.org/solr/)
+    if `search_engine` is configured as `solr`, as opposed to default `legacy` *(a limited database powered search engine)*.
 
     A. **Archive** (tar/zip) *from http://share.ez.no/downloads/downloads*
 
--- a/ezpublish/EzPublishKernel.php
+++ b/ezpublish/EzPublishKernel.php
@@ -49,6 +49,7 @@ class EzPublishKernel extends Kernel
             new EzSystems\PlatformInstallerBundle\EzSystemsPlatformInstallerBundle(),
             new EzSystems\PrivacyCookieBundle\EzSystemsPrivacyCookieBundle(),
             new EzSystems\RepositoryFormsBundle\EzSystemsRepositoryFormsBundle(),
+            new eZ\Bundle\EzPublishSolrSearchEngineBundle\EzPublishSolrSearchEngineBundle(),
         );
 
         switch ($this->getEnvironment()) {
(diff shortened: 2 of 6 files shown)


ezpublish/EzPublishKernel.php
```diff
@@ -49,6 +49,7 @@ public function registerBundles()
             new EzSystems\PlatformInstallerBundle\EzSystemsPlatformInstallerBundle(),
             new EzSystems\PrivacyCookieBundle\EzSystemsPrivacyCookieBundle(),
             new EzSystems\RepositoryFormsBundle\EzSystemsRepositoryFormsBundle(),
+            new eZ\Bundle\EzPublishSolrSearchEngineBundle\EzPublishSolrSearchEngineBundle()
```
andrerom: Name should probably be `EzSystems\PlatformSolrSearchEngineBundle\EzSystemsPlatformSolrSearchEngineBundle(),`, or?

bdunogier:
The approach looks okay.

ezpublish/config/config.yml
```diff
@@ -14,6 +15,36 @@ doctrine:
...
+            core: collection1
+    connections:
+        default:
+            entry_endpoints:
+                - endpoint0
+            mapping:
+                default: endpoint0
+
+# Stash is used for persistence cache in eZ Repository
+# Can have several caches used by different eZ Reposiotries in ezpublish.yml, must be one per database
```
lolautruche: `Repositories`

andrerom: ops, then I have some commits to update :P

pspanja:
Looks good so far.

andrerom:
@bdunogier How do we deal with the binary file exception issue in master? Currently anyone trying solr is forced to comment out https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/IO/TolerantIOService.php#L77

bdunogier:
Do we know why this is happening with Solr, and not with other parts of the
system ?

andrerom:
BD: no idea, first step is to fix demo data, second is to consider having script for upgrades before stable release.. :/

andrerom:
"Simple", the given content does not have a location. @clash82 Can we remove this\* instance of Almighty from the demo data?

\* *content id 126, it does not have any locations so it's not causing issues in UI, but it causes issues when trying to index database with solr, see link to DemoBundle PR above*

clash82:
> Can we remove this* instance of Almighty from the demo data

I think this can be removed. It shouldn't been there. I'll check it and create an update.

clash82:
Extra feature ;) removed in https://github.com/ezsystems/DemoBundle/pull/179

andrerom:
Ready for review now, ping @bdunogier @glye @lolautruche @michalpipa

INSTALL.md
```diff
@@ -17,13 +17,22 @@
 
 ## Install
 
-0. **Create a clean database**
+0. **Create a empty database**, and optionally setup Solr
```
glye: Not sure if relevant for this PR, but same as for the other INSTALL.md I commented on this one also has a mismatch between header number in the source and when rendered, and a "skip to step 3" part which should say "skip to step 4".

glye:
+1 AFAICT, apart from INSTALL.md note.

INSTALL.md
andrerom: I fixed that in the other PR, can't fix it in both :)

bdunogier:
+1 besides maybe the small text suggestions.

andrerom:
Fixed the suggested text addition, thanks!

PR #38, opened 2015-09-06
BDD: Test Console siteaccess

bdunogier:
> Story: [EZP-24774](https://jira.ez.no/browse/EZP-24774)

Adds a `core` profile with a `console` suite to `behat.yml.dist`.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -34,6 +34,12 @@ platformui:
                   - EzSystems\PlatformUIBundle\Features\Context\PlatformUI:
                       uri: /ez
 
+core:
+    suites:
+        console:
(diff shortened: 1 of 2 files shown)


andrerom:
+, but then we should make more clear we have convention to map profiles to bundles, cause core has many meanings unfortunately, rest is also part of core namespace.

bdunogier:
"core" is supposed to match the CoreBundle, but it's indeed not the greatest name ever.

Maybe we can fix it later, when @miguelcleverti moves forward on the structuration of behat tests throughout the bundles ?

andrerom:
ok +1

PR #39, opened 2015-09-08
Add platformUI profile and suite to behat.yml

miguelcleverti:
PlatformUI Behat profile and suites needed to run the BDD tests

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -21,6 +21,19 @@ default:
     # default profile: no suites
     suites: ~
 
+platformui:
+      suites:
+          standard:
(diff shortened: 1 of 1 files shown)


miguelcleverti:
@andrerom @bdunogier @joaoinacio @pcardiga

andrerom:
+1

CopyMoveDelete is bad name for a scenario, but I currently don't have a suggestion for anything better.

joaoinacio:
@andrerom: indeed, how about `ContentOperations` ?

In any case note that it's probably better if scenario name matches the feature folder name, so I'd change both (extra PR/commit... )

otherwise +1

PR #40, opened 2015-09-10
Added log settings in parameters.yml.dist

vidarl:
This PR add log settings in parameter.yml.dist

IMO, configuring of logging is a infrastructure/deployment parameter and should be made available in parameters.yml

### DIFF
--- a/ezpublish/config/parameters.yml.dist
+++ b/ezpublish/config/parameters.yml.dist
@@ -24,3 +24,9 @@ parameters:
 
     # Solr root endpoint, relevant if `solr` is set as search_engine
     solr_dsn: http://localhost:8983/solr
+
+    # Logging settings
+    # Log type is one of "stream", "error_log" or other types supported by monolog
+    log_type: stream
+    # Log path, where to store the log files. php://stderr may be used in order to log to standard error
+    log_path: "%kernel.logs_dir%/%kernel.environment%.log"
--- a/ezpublish/config/config_dev.yml
+++ b/ezpublish/config/config_dev.yml
@@ -14,8 +14,8 @@ web_profiler:
 monolog:
     handlers:
         main:
-            type:   stream
-            path:   "%kernel.logs_dir%/%kernel.environment%.log"
+            type:   %log_type%
+            path:   %log_path%
             level:  debug
         console:
             type:   console
--- a/ezpublish/config/config_prod.yml
+++ b/ezpublish/config/config_prod.yml
@@ -19,8 +19,8 @@ monolog:
             action_level: critical
             handler:      nested
         nested:
-            type:  stream
-            path:  "%kernel.logs_dir%/%kernel.environment%.log"
+            type:  %log_type%
+            path:  %log_path%
             level: debug
         console:
             type:  console


andrerom:
Do you need all of them? I agree this needs to be configurable for docker, but be aware living breathing humans will have to click true everything we add here also, at least until we find a better solution where we can differentiate between sinple/advance options.

andrerom:
Also, because of point above make sure the names here are self describing, as there won't be any hints to humans other then the parameter name. So maybe rather:
```diff
+    # Logging settings
+    # Log type is one of "stream" or "xxx"
+    log_type: stream
+    # Log level, select how verbose logs should be in production, in dev they will always be set to debug
+    log_level: debug
+    # Log path, where to store the log files
+    log_path: "%kernel.logs_dir%/%kernel.environment%.log"
```


SoI think the log level should only be set for production, and also be aware loggers in monolog are nested, so this is not covering everything. @lolautruche what's your take?

ezpublish/config/parameters.yml.dist
```diff
@@ -24,3 +24,8 @@ parameters:
 
     # Solr root endpoint, relevant if `solr` is set as search_engine
     solr_dsn: http://localhost:8983/solr
+
+    # Logging settings
+    monolog_type: stream
+    monolog_level: debug
+    monolog_path: "%kernel.logs_dir%/%kernel.environment%.log"
```
lolautruche: I'd use `log_*` instead of `monolog_*`

andrerom: what he said :)

lolautruche:
Well, the problem here is that we can't cover everything that way in the standard distribution. One will most likely want different logging configuration in dev and in prod (and potentially other environments). We can't predict that.

For Docker, I'd suggest to use a different *distribution* of eZ Platform (as a fork of this one maybe), adding whatever you need (and following updates we do here of course)

andrerom:
> For Docker, I'd suggest to use a different distribution of eZ Platform (as a fork of this one maybe), adding whatever you need (and following updates we do here of course)

not needed, we just need to make sure everything is configurable :) :P

vidarl:
I don't need all the settings, no. Strictly speaking I only need "type":
```monolog_type: error_log```

This will cause phpfpm to send error message to nginx ( over the wire which is a bit bad though )

An alternative is to keep "typ"e as "stream", but instead log to stderr
```monolog_path: php://stderr```

This would be considered best practice as php errors then will actually be logged to stderr in the phpfpm container itself... However, in order for that to work we also need to set nginx setting ```catch_workers_output = yes``` which introduces a quite significant performance penalty per hit

So, I can change from "monolog_...." to just "log_...".
I should only expose "log_type" then ?

**Comment by andrerom** [comment]
> I should only expose "log_type" then ?

I'm ok with keeping them all as well, but both works for me. @lolautruche any pref?

**Comment by vidarl** [comment]
>   # Log level, select how verbose logs should be in production, in dev they will always be set to debug
In my current PR, I also set log level in config_dev.yml... You want the PR to only affect config_prod.yml then ?
That would be OK for mee too...

**Comment by andrerom** [comment]
The level yes, dev should be debug, and not dictated by same as prod.

For dev setup with docker containers we consider eventually look into generating 2 vhosts, one for prod and one for dev. Making it easy to switch to debug when having any issues, but avoid the overhead of dev when using backend UI for instance.

**Comment by vidarl** [comment]
ok. PR updated
 - Added inline description of new settings
 - Fixed so that new settings only apply to prod environment

Only question remaning then if we for instance should leave out log_level from parameters.yml to keep the number of settings in the file to a bare minimum

**Comment by andrerom** [comment]
> Only question remaning then if we for instance should leave out log_level from parameters.yml to keep the number of settings in the file to a bare minimum

yes:
- skip log_level, this is what prod vs dev kind of provides
- reintroduce type and path on config_dev.yml, *what I was saying earlier is that log_level should not be set there*
- rebase branch so it is ready for merge

**Comment by vidarl** [comment]
ok, done.

There has been quite some back and forth... Should I also rewrite the history of this PR ( merge all 6 commits to 1 ) ?

**Comment by andrerom** [comment]
yes :)

**Comment by vidarl** [comment]
done

**Comment by bdunogier** [comment]
I'm not sure I agree with adding those to `parameters.yml`.

We could:
- add the settings
- add default values to `default_settings.yml`
- make docker use its own parameters.yml that overrides those values to what it needs (could be done by appending one line to composer.json, that defines a new location for the parameters.yml.dist)

That would work, right @lolautruche ?

**Comment by andrerom** [comment]
that would work, we could move any setting that is not critical for instal there, to reduce amount of settings users are prompted for.

**Comment by lolautruche** [comment]
+1 with @bdunogier's proposal

**Comment by bdunogier** [comment]
@vidarl about the one liner to modify composer.json, we do something similar in `setup_from_external_repo.sh`:

```
sed -i '$d' composer.json
echo ',    "extra": {"incenteev-parameters":{"dist-file": "/path/to/your/custom/parameters.yml.dist}}}' >> composer.json
```

It should replace the last line from `composer.json` (closing curly bracket), and insert the config for a custom `.dist` file.

vidarl:
To which default_settings.yml should settings be added to actually?

Also, just tried to add it to some ( actually all ) existing default_settings.yml.

If doing for instance
    ```php ezpublish/console --env=prod cache:clear```
I'll get error message : ```You have requested a non-existent parameter "log_type".```
It seems like Extension::load() ain't been executed by console when clearing the cache

vidarl:
That problem with load() makes bdunogier proposal void ?

If I understood him correctly, he meant to use ```type:  %log_type%``` in ```ezpublish/config/config_prod.yml``` and to define ```log_type``` in default_settings.yml, not in standard parameters.yml.dist

andrerom:
Maybe I misunderstood what @bdunogier meant, but I thought  what he implied was adding `ezpublish/config/default_settings.yml` with inline comments that these can be changed by setting them in parameters.yml. and change ``ezpublish/config/config.yml` to import `default_settings.yml` file before `parameters.yml`.

lolautruche:
> Maybe I misunderstood what @bdunogier meant, but I thought what he implied was adding ezpublish/config/default_settings.yml with inline comments that these can be changed by setting them in parameters.yml. and change ezpublish/config/config.yml to import default_settings.yml file before parameters.yml.

Yes, it's exactly what this is about. No reason it wouldn't work. Of course if you add this in a bundle, you'll have the issue you mention.

andrerom:
replaced by #47

PR #41, opened 2015-09-11
Added behat-junit-formatter dependency to composer.json

vidarl:
This PR makes it possible to export results of behat tests in jUnit format, which is applicable when running tests from for instance jenkins.

Unfortunately, the jarnaiz/behat-junit-formatter do not deal with scenarios which contains undefined sentences ( which currently happen to exists our behat tests ). Thus, the custom package  repository...

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -18,6 +18,10 @@ default:
                 env: behat
                 debug: false
 
+        jarnaiz\JUnitFormatter\JUnitFormatterExtension:
+            filename: report.xml
+            outputDir: %paths.base%/build/tests
+
     # default profile: no suites
     suites: ~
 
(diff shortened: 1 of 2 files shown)


andrerom:
+1

bdunogier:
+1

andrerom:
Just had to revert this, as it was failing on travis.

So will have to be re done, in a way so it does not fail.

vidarl:
I then suggest to change my composer.json so it is written on one line, like this
```
    "repositories": [
        { "type": "vcs", "url": "git@github.com:vidarl/behat-junit-formatter.git" }
    ],
```

Then change setup_from_external_repo.sh like this
```
-echo ',    "repositories": [{"type":"git","url":"'$REPO_DIR'"}]}' >> composer.json
+sed -i '/.*git\@github.com:vidarl\/behat-junit-formatter.git.*/i {"type":"git","url":"'$REPO_DIR'"},' composer.json
```

@andrerom  : You agree? Should I fix composer.json as described ?

PR #42, opened 2015-09-18
Use SYMFONY_ENV for consistency

xrow:
The name of the environment variable isn't consistent. Renaming all occurences to SYMFONY_ENV.


andrerom:
consistency with what? This will have to be changed several places, including apache doc.

xrow:
Please look in ezpublish/console. They should be the same from my point of view. I guess you also need to update documentation.

andrerom:
Good point, ok we should look into this then before/during beta, could make sense to do this together with renaming of `ezpublish` folder and so some other intended restructuring.

andrerom:
Handled in #70

xrow:
Thanks for the support.

PR #43, opened 2015-09-18
Added behat-junit-formatter dependency to composer.json

vidarl:
This PR makes it possible to export results of behat tests in jUnit format, which is applicable when running tests from for instance jenkins.

This PR is quite similar to https://github.com/ezsystems/ezplatform/pull/41, except that we no longer need a custom package repository has https://github.com/j-arnaiz/behat-junit-formatter/pull/10 has  been merged.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -18,6 +18,10 @@ default:
                 env: behat
                 debug: false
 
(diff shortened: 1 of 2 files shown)


andrerom:
+1

PR #44, opened 2015-09-18
EZP-24796: update INSTALL doc with additional info for using chmod

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24796

Add some additional information for using `chmod` when setting up eZ Publish.
Even though not ideal, the added notes provide additional info and general "guidelines" for this method.

ping @andrerom @bdunogier @lolautruche

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -84,7 +84,7 @@
 
        C. **Using chown on systems that don't support ACL**
 
-       Some systems don't support ACL at all. You will either need to set your web server's user as the owner of the required directories.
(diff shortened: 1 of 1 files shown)


pedroresende:
+1

joaoinacio:
I guess ezrobot can be safely ignored here...

andrerom:
> I guess ezrobot can be safely ignored here...

well, it seems to be because you applied the change on old commit ;)


could you give your inout on #35? Cause I'm kind of trying to simplify permissions handing, but this is kind of further complicating it :)

lolautruche:
+1

andrerom:
+1

PR #47, opened 2015-09-29
Log settings in parameters

andrerom:
Closes #40


andrerom:
review ping @bdunogier @lolautruche

lolautruche:
+1

PR #48, opened 2015-09-29
Helper branch for EZP-24869

bdunogier:
See https://github.com/ezsystems/ezpublish-kernel/pull/1425

### DIFF
--- a/doc/apache2/vhost.template
+++ b/doc/apache2/vhost.template
@@ -52,7 +52,7 @@
     # Whether to use debugging.
     # Possible values: 0 or 1
     # Defaults to 0 if omitted, unless ENVIRONMENT is set to: "dev"; supported on 5.2 and higher
-    #SetEnv USE_DEBUGGING 0
+    SetEnv USE_DEBUGGING %USE-DEBUGGING%
 
     # Whether to use Symfony's HTTP Caching.
     # Disable it if you are using an external reverse proxy (e.g. Varnish)
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
   - PROFILE="demo" TEST="clean"
   - PROFILE="rest" TEST="fullJson"
   - PROFILE="rest" TEST="fullXml"
-
+  - PROFILE="core"
 
 # test only master (+ Pull requests)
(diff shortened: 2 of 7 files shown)


doc/apache2/vhost.template
```diff
@@ -52,7 +52,7 @@
     # Whether to use debugging.
     # Possible values: 0 or 1
     # Defaults to 0 if omitted, unless ENVIRONMENT is set to: "dev"; supported on 5.2 and higher
-    #SetEnv USE_DEBUGGING 0
+    SetEnv USE_DEBUGGING 1
```
andrerom: should be omitted from merge

bdunogier: Well, no...

Without debugging, we won't get any feedback from errors. It is required by the exceptions parsing added to ezsystems/BehatBundle#34

Why should we absolutely disable it ?

bdunogier: Added a `--use-debugging` option to the `generatevhost.php` script & `vhost.template`. Defaults to 0, set to 1 by `prepare_apache2.sh`.

Does that work for you @andrerom ?

andrerom: better like that
But personally I think prod is what we should test with behat, faster and closer to how it is meant to be run in prod, but I also see the aspect of exceptions detection, so this is ok for now so we can easier spot regressions.

andrerom:
+1

doc/apache2/vhost.template
bdunogier: > But personally I think prod is what we should test with behat, faster and closer to how it is meant to be run in prod

We can do prod + debugging, no problem, they're different settings. Debugging enables stacktraces and a couple other features, but doesn't enforce dev mode at all.

Right now, the `behat` env we use is kind of dev, but mostly with prod settings: https://github.com/ezsystems/ezplatform/blob/master/ezpublish/config/config_behat.yml (inherits config_dev, but forces most settings to their prod value).

andrerom: yes, improvement potentials. Started with the test stuff as basis from symfony-standard, but applies badly to a functional testing use case.

PR #49, opened 2015-09-29
[Behat] Add vhost alias for web, selenium & db

andrerom:
To align with naming in containers to avoid complex, unnecessary, failure-prone config file patching there. 

*Note: `web` implies entry point selenium should run against.*

When passing:
- add host aliases on all repos that uses behat testing via travis
 - [ ] kernel
 - [ ] platform ui (PR)
 - [ ] demo
 - [ ] behat bundle 
- alternative to list of changes above would be to add hosts in selenium init scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ env:
   - PROFILE="rest" TEST="fullJson"
   - PROFILE="rest" TEST="fullXml"
 
+# host aliases for behat tests to be able to align config and scripts for future docker changes
+addons:
+  hosts: [web, selenium, db]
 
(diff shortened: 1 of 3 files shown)


andrerom:
cc @vidarl

joaoinacio:
my 2 cents:
Unless I am missing something, the problem I see here is that - besides travis - everyone else will have a default behat.yml that does not work out of the box (every env has a `localhost`, but not `web` host).
There is also no mention of the need to change it on documentation...

andrerom:
@joaoinacio You are right, but also be aware that QA also uses host aliases for this.

PR #50, opened 2015-10-05
jenkins_integration_test_please_ignore

vidarl:
Please ignore this PR. It is used for experimental jenkins integration


andrerom:
should set a different status then just ´ezrobot´, see https://github.com/ezsystems/ezcs/pull/15

PR #54, opened 2015-10-12
Behat imports + PlatformBehatBundle routing

bdunogier:
This PR tries to move behat profile & suite for core to a `behat.yml` file inside `vendor/ezsystems/ezpublish-kernel/eZ/Bundles/EzPublishCoreBundle`, with an import in ezplatform's `behat.yml`.

It would allow PR authors to add scenarii without having to open a PR. Furthermore, it, I think, makes more sense since the Behat feature + context config is in the same repository than the actual tested feature.

The PR also adds a `routing_behat.yml` file, so that we can load the routes defined in The `PlatformBehatBundle` inside the kernel.

### TODO
- [ ] Rebase, removing the behat.yml imports commit


bdunogier:
Ping @miguelcleverti @andrerom.

bdunogier:
The failure is a known one from the refactoring.

miguelcleverti:
+1

andrerom:
+1

bdunogier:
The part about behat.yml and imports has been merged in another PR.

andrerom:
close?

bdunogier:
yep

PR #56, opened 2015-10-13
EZP-24952: [VCL] Clarify User hash documentation

andrerom:


joaoinacio:
+1
IMHO this resolves https://jira.ez.no/browse/EZP-24952

doc/varnish/vcl/varnish3.vcl
```diff
@@ -138,8 +138,14 @@ sub ez_user_hash {
...
 
     if (req.restarts == 0 && (req.request == "GET" || req.request == "HEAD")) {
-        // Anonymous user => Set a hardcoded anonymous hash
+        // Get User (Context) hash, for varying cache by what user has access to.
+        // https://doc.ez.no/display/EZP/Context+aware+HTTP+cache
+
+        // Anonymous user => Set a hardcoded anonymous hash to avoid backend lookup for hash
         if (req.http.Cookie !~ "eZSESSID" && !req.http.authorization) {
+            // You may update this hash with the actual one for anonymous user
+            // to archive a bit better cache hit ratio across anonymous users.
```
lolautruche: s/archive/achieve ? Still not sure about the word though. Maybe just _to get a better cache hit ratio_?

doc/varnish/vcl/varnish4.vcl
```diff
@@ -135,8 +135,14 @@ sub ez_user_hash {
...
 
     if (req.restarts == 0 && (req.method == "GET" || req.method == "HEAD")) {
-        // Anonymous user => Set a hardcoded anonymous hash
+        // Get User (Context) hash, for varying cache by what user has access to.
+        // https://doc.ez.no/display/EZP/Context+aware+HTTP+cache
+
+        // Anonymous user => Set a hardcoded anonymous hash to avoid backend lookup for hash
         if (req.http.Cookie !~ "eZSESSID" && !req.http.authorization) {
+            // You may update this hash with the actual one for anonymous user
+            // to archive a bit better cache hit ratio across anonymous users.
```
lolautruche: Same as above

lolautruche:
So, as far as I understand this, suggestion would ensure that we don't have 2 identical variations of cache of anonymous users (with and without session)?

If so, +1

andrerom:
> So, as far as I understand this, suggestion would ensure that we don't have 2 identical variations of cache of anonymous users (with and without session)?

yes

doc/varnish/vcl/varnish4.vcl
```diff
@@ -135,8 +135,14 @@ sub ez_user_hash {
...
     if (req.restarts == 0 && (req.method == "GET" || req.method == "HEAD")) {
-        // Anonymous user => Set a hardcoded anonymous hash
+        // Get User (Context) hash, for varying cache by what user has access to.
+        // https://doc.ez.no/display/EZP/Context+aware+HTTP+cache
+
+        // Anonymous user w/o session => Use hardcoded anonymous hash to avoid backend lookup for hash
         if (req.http.Cookie !~ "eZSESSID" && !req.http.authorization) {
+            // You may update this hash with the actual one for anonymous user
+            // to get a better cache hit ratio across anonymous users.
+            // Note: You should then update it every time anonymous user rights change.
```
joaoinacio: Nitpick: 
`every time anonymous user rights change.` - note that modifying the anonymous role (adding/removing policies) has no effect in the hash AFAIK, only changing role(s) and assignment(s) for the user/group does.

andrerom: @lolautruche do you remember this part to verify?
 is it role id's + role assignment limitations? Or does it also loop over policies?

If @joaoinacio is right, then maybe:
`// Note: You should then update it every time Anonymous User inherited Role Assignments change.`

andrerom: no feedback, feel free to propose change afterwards, but sentence as-is is fine, it simply does not specify this.

PR #60, opened 2015-10-15
[TMP] Testing contianers for fun and profit

andrerom:
This will probably not work, mysql on container infra is a lot slower.

There is also a new trusty env to consider testing, but it does not have mysql yet, but seems to be added soon.


andrerom:
Not working, sudo needed for generating the needed config.

PR #61, opened 2015-10-21
Implement EZP-24925: Extract XmlText field type out of kernel into separate repository

pspanja:
This PR resolves https://jira.ez.no/browse/EZP-24925

This adds `ezsystems/ezplatform-xmltext-fieldtype` to `composer.json` and registers its bundle with kernel.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
         "ezsystems/ezpublish-kernel": "~6.0@dev",
         "ezsystems/repository-forms": "~1.0.0@dev",
         "ezsystems/ezplatform-solr-search-engine": "~1.0.0@dev",
+        "ezsystems/ezplatform-xmltext-fieldtype": "~1.0.0@dev",
         "ezsystems/platform-ui-bundle": "~1.0.0@dev",
         "ezsystems/platform-ui-assets-bundle": "~1.0.0@dev",
         "ezsystems/demobundle": "~6.0@dev",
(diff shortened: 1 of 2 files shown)


pspanja:
Review ping @andrerom @lolautruche @bdunogier

andrerom:
Q: Do we need to bundle by default for Platform? Is it still used somewhere after demo data is changed?

pspanja:
Not used, but for the time being I would say we should bundle it, I know that @lolautruche and @rpmsauron had issues when I removed XmlText from kernel as they used DB with XmlText data.
For the release and with migration scripts I don't think we need to include it.

andrerom:
ok, clear

lolautruche:
> For the release and with migration scripts I don't think we need to include it.

I don't agree. Think about people who won't use PlatformUI right away. It's more than likely that people will continue to use legacy admin, and there is no legacy datatype covering RichText.
So I really think we should bundle it for the release, while we deprecate XMLText fieldType. We can remove it later (e.g. after 1st LTS).

+1 for this change to be in :-)

andrerom:
> I don't agree. Think about people who won't use PlatformUI right away. It's more than likely that people will continue to use legacy admin, and there is no legacy datatype covering RichText.
So I really think we should bundle it for the release, while we deprecate XMLText fieldType. We can remove it later (e.g. after 1st LTS).


I would say aim to remove it form ezplatform meta repo, and add it to ezpublish.community repo, everything else is aligned so that the latter is the one to pick for platform + publish setup.

PR #63, opened 2015-11-03
[Doc] Add basic htaccess config for simple setup

andrerom:
Pulled in a few tricks from https://raw.githubusercontent.com/symfony/symfony-standard/2.8/web/.htaccess

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -114,10 +114,23 @@
        $ sudo chmod g+s {ezpublish/{cache,logs,config,sessions},web}
        ```
 
-3. **Configure a VirtualHost**:
+3. **Configure a Webserver**:
 
-    A virtual host setup is the recommended, most secure setup of eZ Publish.
-    General virtual host setup template for Apache and Nginx can be found in [doc/ folder](doc/).
+    eZ Platform can theoretically run with any web server cable of running PHP, but for most secure operation
+    we provide ready made configuration templates for the following servers:
+
+    A. **Apache**
+
+    Recommended setup for prod setup is using a virtual host setup, see [doc/apache2 folder](doc/apache2/) for details.
+
+    Tip: For just testing eZ Platform you may copy provided .htaccess file into web folder:
+    ```bash
+    $ cp doc/apache2/.htaccess web/
+    ```
+
+    B. **Nginx**
+
+    Recommended setup for prod setup is using a virtual host setup, see [doc/nginx folder](doc/nginx/) for details.
 
 
 4. **Run installation command**:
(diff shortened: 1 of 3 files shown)


emodric:
Why not make it more inline with vhost config, with environment variables and rewrite rules that check for dev/prod environment and whatnot?

andrerom:
Cause I was thinking of just placing it in web folder as a default fallback, hence it would not be something you edit normally it would just be there. But as that might cause lots of issues for those that forget `AllowOverride None` in vhost setup.

INSTALL.md
```diff
@@ -114,10 +114,23 @@
...
+3. **Configure a Webserver**:
 
-    A virtual host setup is the recommended, most secure setup of eZ Publish.
-    General virtual host setup template for Apache and Nginx can be found in [doc/ folder](doc/).
+    eZ Platform can theoretically run with any web server cable of running PHP, but for most secure operation
+    we provide ready made configuration templates for the following servers:
+
+    A. **Apache**
+
+    Recommended setup for prod setup is using a virtual host setup, see [doc/apache2 folder](doc/apache2/) for details.
```
bdunogier: That's a lot of "setup". What about "A virtual host based setup is recommended for production. See..."

INSTALL.md
```diff
@@ -114,10 +114,23 @@
...
+
+    Tip: For just testing eZ Platform you may copy provided .htaccess file into web folder:
+    ```bash
+    $ cp doc/apache2/.htaccess web/
+    ```
+
+    B. **Nginx**
+
+    Recommended setup for prod setup is using a virtual host setup, see [doc/nginx folder](doc/nginx/) for details.
```
bdunogier: Same as above.

bdunogier:
If somebody has AllowOverride disabled, will the .htaccess file will be ignored ? Or will it throw an error ?

emodric:
> Cause I was thinking of just placing it in web folder as a default fallback, hence it would not be something you edit normally it would just be there. But as that might cause lots of issues for those that forget AllowOverride None in vhost setup, I didn't. pov?

Hm... I wouldn't want to keep it in `web` folder for sure. Doc is enough. IIRC, it was like that for legacy too, vhost variant of apache setup and .htaccess variant of apache setup.

andrerom:
> If somebody has AllowOverride disabled, will the .htaccess file will be ignored ?

Afaik `AllowOverride None` means Apache won't look for it at all, so no lookup, and no conflict.

emodric:
Most of the time .htaccess (if used of course) will need to be edited, because you need to open up additional stuff to the world (google domain verifications and so on). Often .htaccess files differ in dev and prod environment, so having a basic .htaccess file which will not be changed is not desired.

IMO, an .htaccess file that completely matches & mirrors vhost setup should be provided in the docs and that's it (the way it is now in this PR).

andrerom:
Closing, will instead move this inline in readme for apache config as a simple example for prod, to avoid having to cover everything.

PR #65, opened 2015-11-12
Make ezplatform a standard distribution

bdunogier:
> Part of http://jira.ez.no/browse/EZP-25046
> Requires https://github.com/ezsystems/ezpublish-kernel/pull/1496

With these changes, starting a project should be as simple as:
- create-project
- set database parameters
- run `ezplatform/install clean` (could integrate database creation if there isn't one)

The default `AppBundle`, taken from the Symfony standard installer changes, should make it easy to get started, as well as the default configuration (no need for prepends or include, unless you want/need to).

### DemoBundle removal
Removes the demobundle, as well as its dependencies from platform's requirements.
A new distribution of platform, `ezsystems/ezplatform-demo`, will be created for it.

### Config cleanup / revamp
Cleans up the configuration, by moving the minimal ezpublish config to `config.yml`.

### Default `AppBundle`
Adds a default AppBundle in src, enabled by default, similar to the one in Symfony standard.

### BDD optimisations
Since the demobundle does not need to be tested from this repository, `.travis.yml` is simplified.
In addition, the env variables used to configure behat are replaced with a single one that contains the command line arguments.
Scenarios tagged as `@broken` are now skipped by default.

### TODO
- [x] Describe
- [x] Remove other demo references (install etc)
- [x] Think about other default config files (`view.yml`, ` siteaccess.yml`... ?), included from `config.yml`
- [x] Cherry pick default view from `symfony-standard` into in `ezpublish/Resources/views`
- [x] Change text in default AppBundleController ([comment](https://github.com/ezsystems/ezplatform/pull/65#discussion_r44871626))
- [x] Update INSTALL.md / README.md


andrerom:
travis/behat is failing, I guess it should be changed in same pr

bdunogier:
@andrerom yep, i'll check that.

bdunogier:
Removed more stuff, let's see what it does.

src/AppBundle/Tests/Controller/DefaultControllerTest.php
```diff
@@ -0,0 +1,18 @@
...
+class DefaultControllerTest extends WebTestCase
+{
+    public function testIndex()
+    {
+        $client = static::createClient();
+
+        $crawler = $client->request('GET', '/');
+
+        $this->assertEquals(200, $client->getResponse()->getStatusCode());
+        $this->assertContains('Welcome to Symfony', $crawler->filter('#container h1')->text());
```
andrerom: I guess we can say eZ Platform unless idea is to stay as close to sf-standard as possible above all of course.

bdunogier: I haven't even opened this controller. Feel free to commit the change. I'll add it as a TODO.

bdunogier:
And tests are passing :dancers:

emodric:
Is there really need for AppBundle? I'd like this to be as clean as possible :)

emodric:
Related to this todo: "Think about other default config files (view.yml, siteaccess.yml... ?), included from config.yml"

You might wanna think about how we should handle siteaccess matching on different machines. Often, siteaccess matching is done differently on dev, stage and prod machines. When you factor in local development instances (in vagrant for example) the problem becomes really obvious and tough to handle.

Using `config_dev.yml` and `config_prod.yml` for this purpose is not a good solution, since central dev/local vagrant enviroments use the same `dev` Symfony env, while staging and prod mostly use same Symfony `prod` environment.

The best we could come up with is either parametrizing (in `parameters.yml`) various domains that site uses (but this becomes problematic if sites use many domains), or parametrizing (again in `parameters.yml`) the entire siteaccess match array, but this is extremely ugly looking when doing composer install :D

ezpublish/config/config.yml
```diff
@@ -101,3 +101,22 @@ swiftmailer:
     username:  "%mailer_user%"
     password:  "%mailer_password%"
     spool:     { type: memory }
+
+# Base configuration for eZ Platform
```
emodric: Hm... I don't really like this here. In most cases, this will be the first thing you will remove after creating a new site (at least the siteaccess and languages configuration part).

The way it is now makes more sense to me. The file template is kept in platform installer and copied in `ezpublish/config/ezpublish.yml` when installing the site.

Besides, `ezpublish.yml` file looked good as a central eZ Publish configuration file, in addition to `config.yml` which configures the framework.

bdunogier:
> The best we could come up with is either parametrizing (in parameters.yml) various domains that site uses (but this becomes problematic if sites use many domains), or parametrizing (again in parameters.yml) the entire siteaccess match array, but this is extremely ugly looking when doing composer install :D

I agree, that would be best. And that's what you'd gotta do, actually. Because, and you've explored the topic pretty well, it would require that we set constraints that won't fit everybody.

I think that the setup proposed in this PR goes in the right direction: simple. More complicated / siteaccess based cases will have to be configured "manually". But since you can very easily maintain your own fork of ezplatform, it should all be fine, right ? :-)

> The file template is kept in platform installer and copied in ezpublish/config/ezpublish.yml when installing the site.

What's the problem with changing `config.yml` ? It's yours once the project starts :-)

I'm not against having an `ezpublish.yml`. But overall, if you look at config.yml, it's the app's config file. And the fact is that ezplatform is part of the app, it's not something above it.

I'd still like to keep the default file in the application. Nothing prevents an installer from overwriting this file. But you don't _have_ to do so to get started.

The approach above (own fork to get started from) also applies here.

bdunogier:
> Is there really need for AppBundle? I'd like this to be as clean as possible :)

Unfortunately, it is also about what others want ;-)
Also, why wouldn't you use such a bundle ? After all, you _always_ start a project with a Bundle, and this one is _really_ simple.

Overall, we are trying to stick to the standards that are being established by Symfony. There has been a _huge_ effort on developer experience over there, and while some choices might feel odd, they actually make a lot of sense.

emodric:
> Also, why wouldn't you use such a bundle ? After all, you always start a project with a Bundle, and this one is really simple.

I prefer having my company/my client nomenclature in the source files :)

No biggie, I will delete it from our standard distribution ;)

bdunogier:
> I prefer having my company/my client nomenclature in the source files :)

We indeed can't do much about the namespace, agreed :-)

emodric:
> I think that the setup proposed in this PR goes in the right direction: simple. More complicated / siteaccess based cases will have to be configured "manually". But since you can very easily maintain your own fork of ezplatform, it should all be fine, right ? :-)

Sure, but I was thinking of you maybe providing some best practices, not necesarilly an implementation :)

> What's the problem with changing config.yml ? It's yours once the project starts :-)

Any processing of `config.yml` (for example with project generator I showed at summer camp) removes all comments (and we don't want that ;)). It's convenient to have eZ Publish config separate.

> I'm not against having an ezpublish.yml. But overall, if you look at config.yml, it's the app's config file. And the fact is that ezplatform is part of the app, it's not something above it.

Sure, but there's eZ Publish and there's all other stuff :D

> I'd still like to keep the default file in the application. Nothing prevents an installer from overwriting this file. But you don't have to do so to get started.

I would too, just separate from `config.yml`

> The approach above (own fork to get started from) also applies here.

Sure :)

bdunogier:
Config moved to `ezplatform.yml`, imported from `config.yml`.

emodric:
> Config moved to ezplatform.yml, imported from config.yml.

:+1:

ezpublish/EzPublishKernel.php
```diff
@@ -82,18 +80,12 @@ public function registerBundles()
...
 
-        // if config file or base config file is missing, then install is not done yet
-        if (!is_file($configFile) || !is_file(__DIR__ . '/config/ezpublish.yml')) {
-            $configFile = __DIR__ . '/config/ezpublish_setup.yml';
-        }
+        $loader->load($rootDir . '/config/config_' . $environment . '.yml');
 
-        if (!is_readable($configFile)) {
-            throw new RuntimeException("Configuration file '$configFile' is not readable.");
+        if (file_exists($ezpublishConfigFile = $rootDir . '/config/ezpublish_' . $environment . '.yml')) {
```
andrerom: nitpick, but be aware that this implies people may skip having the file and have a uncached fs call for every request.

ezpublish/config/config.yml
```diff
@@ -2,6 +2,7 @@ imports:
     - { resource: default_parameters.yml }
     - { resource: parameters.yml }
     - { resource: security.yml }
+    - { resource: ezplatform.yml }
```
andrerom: to avoid the extra fs call, what about just doing import of corresponding ezpublis_<env>.yml file from each and every config_<env>.yml file. This way siteaccess config added in `ezpublish/config/config_behat.yml` could rather be in ezpublish_behat.yml. But then again, this creates convention to have env settings also in ezpublish files which is probably not the direction you are taking here, so maybe not the way to go.

emodric: > But then again, this creates convention to have env settings also in ezpublish files which is probably not the direction you are taking here, so maybe not the way to go.

Other config files have environment variants, I don't see a reason why `ezplatform.yml` shouldn't.

bdunogier: Not all, if you look carefully: `parameters.yml` does not, `security.yml` don't have an env variant, for instance.

ezpublish/EzPublishKernel.php
bdunogier: The idea was to stay backward compatible with existing `ezpublish.yml` files. But you're right about the extra FS call...

bdunogier:
Created one `ezplatform.yml` per environment, imported from the matching `config.yml`.
The behat config is moved from `config_behat.yml` to `ezplatform_behat.yml`.

ezpublish/EzPublishKernel.php
emodric: Is file_exists needed at all?

I mean, if $loader->load() call fails, we have bigger issues anyhow. :)

emodric: Ah, I now see the note above about BC :)

bdunogier: I have removed this line. Update doc should be sufficient, it's just about renaming a file anyway.

emodric:
@bdunogier Do you intend to remove cookie and comments bundle?

ezpublish/config/config.yml
```diff
@@ -85,7 +85,7 @@ twig:
 assetic:
     debug:          "%kernel.debug%"
     use_controller: false
-    bundles:        [ eZDemoBundle ]
+    bundles:        []
```
emodric: Just wondering. Now that `bundles` array is empty, is it going to overwrite any bundles that add them selves here via prepends method?

ezpublish/config/default_settings.yml
```diff
@@ -0,0 +1 @@
+parameters: {  }
```
emodric: What's the usecase for this file?

bdunogier: Good question, as a matter of fact.

It is used in the AppBuilder thing I've told you about. The idea is that default values are placed in that file. But in any case, it doesn't belong in this PR.

ezpublish/config/config.yml
bdunogier: That is a good point. I don't know :-)

I'd say that it won't behave differently, since we already had an array value there in any case. What did you do when you changed this in your project, after removing the demobundle ? Leave the array empty ? Replace with another one ?

bdunogier:
> Do you intend to remove cookie and comments bundle?

And they're gone.

ezpublish/config/config.yml
blankse: I leave the array empty and add every Bundle via prepend method:
```
$container->prependExtensionConfig('assetic', ['bundles' => ['AppBundle']]);
```
Works good for me :)

emodric: I do this :D

```
$container->prependExtensionConfig(
    'assetic',
    array(
        'bundles' => array_keys(
            $container->getParameter( 'kernel.bundles' )
        )
    )
);
```

bdunogier: Maybe we should just drop the array then ? I mean, it's not very useful, and the AppBundle doesn't need it.

bdunogier:
All todos have been taken care of.
I need +1s from the team now. @yannickroger @andrerom @lolautruche ? @emodric and @blankse you can still +1. It won't get it accepted, but I'd consider a -1 if something is missing :-)

src/AppBundle/Tests/Controller/DefaultControllerTest.php
bdunogier: I'm actually gonna remove it. While it makes sense to provide a default controller for people to get started from in a Symfony context, it isn't really required for eZ Platform. The root content will be displayed thanks to the default templates, and the first change to the code will probably be to add a new content_view.

andrerom:
+1, minor, but make sure to take care about #66 as part or post merge this one.

emodric:
Looks good to me :+1:

composer.json
```diff
@@ -26,16 +26,12 @@
...
+        "ezsystems/ezpublish-kernel": "dev-default_view_templates as 6.0.x-dev",
         "ezsystems/repository-forms": "~1.0.0@dev",
         "ezsystems/ezplatform-solr-search-engine": "~1.0.0@dev",
         "ezsystems/ezplatform-xmltext-fieldtype": "~1.0.0@dev",
         "ezsystems/platform-ui-bundle": "~1.0.0@dev",
         "ezsystems/platform-ui-assets-bundle": "~1.0.0@dev",
-        "ezsystems/demobundle": "~6.0@dev",
-        "ezsystems/demobundle-data": "~1.0@dev",
-        "ezsystems/comments-bundle": "~6.0@dev",
-        "ezsystems/privacy-cookie-bundle": "~1.0@dev",
```
lolautruche: What about adding privacy-cookie-bundle and comments-bundle to suggest section?

lolautruche:
+1

bdunogier:
Cleaned up, merging.

PR #66, opened 2015-11-16
Disabling Stash Tracking in example config

janit:
Stash tracking (http://www.stashphp.com/Symfony.html#tracking) in development mode fills up the disk when working with large numbers of content. It is also not very crucial during for developers creating sites with eZ Platform, so I suggest disabling it by default.

### DIFF
--- a/ezpublish/config/ezpublish.yml.example
+++ b/ezpublish/config/ezpublish.yml.example
@@ -12,6 +12,7 @@ doctrine:
 
 # Stash is used for persistence cache
 stash:
+    tracking: false # enables query logging
     caches:
         default:
             drivers:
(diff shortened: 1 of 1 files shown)


andrerom:
+1, but would be better with `tracking_values`, however not part of a release yet: https://github.com/tedious/TedivmStashBundle/issues/80

yannickroger:
ping @lolautruche @bdunogier

lolautruche:
+1

bdunogier:
Why not put the comment above the line as we usually do ?

andrerom:
well, this needs to be changed now, files does not exists anymore.

andrerom:
1 second :clock10: , there is a new minor stash release comming

PR #68, opened 2015-11-19
Disable Stash cache value tracking by default

andrerom:
replaces / closes #66 

Should speed up things in dev env a bit, in addition to avoiding disk the issue with large disk use by profiler files. Very optimistic, but might be it gets rid of the strange issues some people have with profiler not loading, but that would be just to much of a luck :)


andrerom:
@bdunogier aware of the behat failures?

PR #69, opened 2015-11-25
Update ezplatform.yml to re add missing %search_engine% param

andrerom:


bdunogier:
Okay with the parameter. But I wouldn't add inline doc in this file, as we should limit changes there to the strict minimum.

bdunogier:
By the way, `storage: ~` is the default, not sure why it would have to be added to this file.

PR #70, opened 2015-11-29
Renamed 'ezpublish' to 'app'

bdunogier:
> Implements http://jira.ez.no/browse/EZP-25110

Renames the `ezpublish` folder to `app`, to be closer from symfony standards and documentation.

Also renames:
- `EzPublishKernel` to `AppKernel`
- `web/index.php` to `web/app.php`

### TODO
- [x] Update `INSTALL.md`
- [x] Update nginx and apache2 config
- [x] Update Kernel commands: https://github.com/ezsystems/ezpublish-kernel/pull/1520


masev:
:thumbsup:

web/app.php
```diff
@@ -50,7 +50,7 @@
     if (($httpCacheClass = getenv('HTTP_CACHE_CLASS')) !== false) {
         $kernel = new $httpCacheClass($kernel);
     } else {
-        require_once __DIR__ . '/../ezpublish/EzPublishCache.php';
+        require_once __DIR__ . '/../app/AppCache.php';
         $kernel = new EzPublishCache($kernel);
```
andrerom: Rename?

INSTALL.md
```diff
@@ -61,7 +61,7 @@
 
 2. *Only for *NIX users* **Setup folder rights**<a name="install-2-folder-rights"></a>:
 
-       One common issue is that the `ezpublish/cache`, `ezpublish/logs` and `ezpublish/config` directories **must be writable both by the web server and the command line user**.
+       One common issue is that the `app/cache`, `app/logs` and `app/config` directories **must be writable both by the web server and the command line user**.
        If your web server user is different from your command line user, you can run the following commands just once in your project to ensure that permissions will be set up properly.
```
michalpipa: Do we still need `app/config` to be writable by web server?

bdunogier: By the web server, no, not that I know.

web/app.php
```diff
@@ -50,8 +50,8 @@
     if (($httpCacheClass = getenv('HTTP_CACHE_CLASS')) !== false) {
         $kernel = new $httpCacheClass($kernel);
     } else {
-        require_once __DIR__ . '/../ezpublish/EzPublishCache.php';
-        $kernel = new EzPublishCache($kernel);
+        require_once __DIR__ . '/../app/AppCache.php';
```
blankse: This file and class aren't renamed yet

bdunogier: Done by  @andrerom in 889da90.

emodric:
`app/sessions` folder is not used by default, is it? Why have it in install instructions?

andrerom:
> app/sessions folder is not used by default, is it? Why have it in install instructions and repo?

good question, what is the default now actually?

emodric:
> good question, what is the default now actually? #lazyWeb

On my system its `/var/lib/php5/sessions`

andrerom:
Passes, ready for review @bdunogier @lolautruche @blankse @masev @emodric @michalpipa

INSTALL.md
andrerom: fixed

.gitignore
```diff
@@ -4,14 +4,14 @@
...
+/app/cache/
+/app/logs/
 /ezpublish_legacy
-/ezpublish/config/parameters.yml
-/ezpublish/config/ezpublish_prod.yml
-/ezpublish/config/ezpublish_dev.yml
-/ezpublish/sessions
-/ezpublish/bootstrap.php.cache
+/app/config/parameters.yml
+/app/config/ezplatform_prod.yml
```
bdunogier: Do we want to keep those here ? We do (version them)[https://github.com/ezsystems/ezplatform/tree/master/ezpublish/config].

emodric: Should be removed, yes.

emodric:
@andrerom No thoughs about `app/sessions` folder?

.gitignore
andrerom: Which files are supposed to be edited by user now? That is my problem here. Is it ok if we generate these during install maybe so they are safly editable by downstream? (Where did we recently discuss this btw?)

emodric: IMO, only `app/config/parameters.yml` is generated on install and it should be ignored and outside of repo. `app/config/ezplatform_prod.yml` & `app/config/ezplatform_dev.yml` should be in the repo since they will be versioned almost (if not) always.

andrerom:
> @andrerom No thoughs about app/sessions folder?

I didn't really want to dig into that as part of this, but here goes: it was introduced in #133 ( https://jira.ez.no/browse/EZP-22747 ), but I see I removed it, potentially by mistake in 0cd0fc3cfac574ce9cb429772522fdc6ba835304. But ok, since we aim to really sync up with Symfony here, it is now gone.

.gitignore
andrerom: fixed

emodric:
> But ok, since we aim to really sync up with Symfony here (within what we have here now), it is now gone.

+1 :)

bdunogier:
> but I see I removed config for it, potentially by mistake in 0cd0fc3

Serendipity strikes again.

+1

tharkun:
+1

michalpipa:
+1

bdunogier:
That is plenty of +1, i'll merge this today.

PR #73, opened 2015-12-02
Removed xmltext fieldtype

bdunogier:
> Part of https://jira.ez.no/browse/EZP-24925

Not part of ezplatform by default, but can be installed separately from http://packagist.org/packages/ezsystems/ezplatform-xmltext-fieldtype.


andrerom:
+1  (context: as in this will be documented in upgrade doc as an option, while we won't ship it by default).

bdunogier:
Thank you @andrerom, added to the description.

michalpipa:
+1

PR #74, opened 2015-12-16
Set package type to "project"

jeromegamez:
eZ Platform is a project/application, not "just" a library… Well, that's a massive PR for once :smile:


bdunogier:
Looks great, let's just see what Travis thinks (nothing in particular, I assume).

andrerom:
+1

PR #78, opened 2016-01-06
[Security][Travis] Update approach for github token to same as symfony uses

andrerom:
The [symfony approach](https://github.com/symfony/symfony/commit/6215437bdd6d6fc67ee167aea5e88324117be668) is more straight forward, and avoids setting a bad example for others *(risking they create tokens with a scope without being protected by githubs commit token scanners, unintentionally exposing rights to their repos)*.

This issue was reported by @ChALkeR as a possible security risk for us and especially downstream projects. *Note: The user connected to the token in question afaik never had membership rights on any of our repositories, however it had a scope enabled so if user by mistake would have had access to anything, then so would the token. The approach here is thus safer, as Gitub will automatically revoke any tokens they detect with scopes that pose a direct security risk.*

### DIFF
--- a/bin/.travis/composer-auth.json
+++ b/bin/.travis/composer-auth.json
@@ -1,0 +1,8 @@
+{
+    "github-oauth": {
+        "github.com": "PLEASE DO NOT USE THIS TOKEN IN YOUR OWN PROJECTS/FORKS",
+        "github.com": "This token is reserved for testing with ezsystems repositories",
+        "github.com": "NB: You create your own token without(!) any scope to use same approach",
+        "github.com": "d0285ed5c8644f30547572ead2ed897431c1fc09"
+    }
+}
--- a/bin/.travis/install_composer_github_key.sh
+++ b/bin/.travis/install_composer_github_key.sh
@@ -1,9 +1,0 @@
-#!/bin/sh
-
-# This is unfortunately only way to avoid Github API limits with Travis on an open source project.
-# Please don't use our token, neither token nor it's user gives access to anything other then authentication, so better if you create your own.
-# PS: The simple obfuscation here is only to avoid Github from detecting this on commits.
-export EZ_GITHUB_TOKEN_A=`echo "574503f4468c5de5820d" | rev`
-export EZ_GITHUB_TOKEN_B=`echo "90cf1c134798de2dae27" | rev`
-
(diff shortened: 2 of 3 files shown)


bdunogier:
+1

andrerom:
Merged in bf80b0baa87e3bcdf7f0aa967e6ed2ce8ae8ffc1.

andrerom:
## Note
Code using this approach used by the community, and those used as tutorials should be fixed as well. I'll handle the other repos on ezsystems, but can't change the forks

Affected forks:
@lolautruche https://github.com/lolautruche/ezconference2015/blob/d15c64935770f273ca5424fca40cddc898227ec1/bin/.travis/install_composer_github_key.sh
@crevillo https://github.com/crevillo/ezplatform-legacy/blob/3cb93a76820dcb1e861c677f0ce2e3da2e52b32d/bin/.travis/install_composer_github_key.sh
https://github.com/search?p=1&q=EZ_GITHUB_TOKEN_A&type=Code
@bdunogier https://github.com/bdunogier/platform-workshop/blob/e767eb3a943c9101058dd704d94eb32b9d35734e/bin/.travis/install_composer_github_key.sh
@flovntp https://github.com/flovntp/BikeTutorialWebsite/blob/234e26a437ca089067e4bdd2632fc7af0f663b3f/bin/.travis/install_composer_github_key.sh

There is also use of this approach in eZ SummerCode 2015 code:
@emodric https://github.com/netgen/summercamp-2015/blob/01ad5beffe1a41def76a46fddf4bbede4769158d/run.sh

crevillo:
thanks @andrerom , i'll update it tomorrow

emodric:
@andrerom Yep, that token is already revoked. Thanks!

crevillo:
done on my side @andrerom https://github.com/crevillo/ezplatform-legacy/commit/bf80b0baa87e3bcdf7f0aa967e6ed2ce8ae8ffc1

andrerom:
> @andrerom Yep, that token is already revoked. Thanks!

@emodric it is not so much about the token, but the bad example it sets, when people use the approach they risk exposing tokens they shouldn't.

emodric:
@andrerom I know, I get that. I've updated the code to include the token which has no scopes and which does not circumvent github's checks.

andrerom:
thanks :)

PR #79, opened 2016-01-07
WIP : Updated web config so settings can be injected via environment variables

vidarl:
This PR makes it easier to deal with provided vhost config autocratically  as values now can be injected using tools like envsubst instead having to do search&replace using regexp and sed etc...


vidarl:
Todo:
 - Update readme ( nginx.rst )
 - Update apache vhost accordingly

andrerom:
The problem with envsubst is that we can' define defaults, neither in template using ` ${variable-default}`, or in a script like I do in the [branch](https://github.com/ezsystems/ezplatform/tree/tmp_bash_vhost) where I'm trying to do the same for apache.

bdunogier:
I personally don't really see the problem with using sed. It's what most container images do, and it's not that complicated.

yannickroger:
Yeah, it's not like we have a 100 keys to be replaced.

vidarl:
The problem with sed is that you explicitly have to mention all replacements. So if you want to replace PORT, BASEDIR + 10 other variables, you have write them all on the command line.

However, in AR's ```tmp_bash_vhost```  branch he anyway creates a script for doing the replacements. So then I am actually fine usinging sed ( or bash substitution which AR attempts to do )
But I initially misunderstood AR as I though he  he did not want to have a separate script to handle the substitution.

andrerom:
> But I initially misunderstood AR as I though he he did not want to have a separate script to handle the substitution.

Actually you did not, ideally I wanted to avoid this, but not at the cost of making the template unreadable, and not at the cost of usability *(I consider having to define, and understand all env's, every time, not very usable)*.

vidarl:
I'll close this PR and create a new one for nginx based on AR's tmp_bash_vhost branch

vidarl:
FYI : new PR : https://github.com/ezsystems/ezplatform/pull/83/

PR #80, opened 2016-01-12
Improve EZP-25369: Remove configuration from installers

andrerom:
https://jira.ez.no/browse/EZP-25369

### DIFF
--- a/app/config/ezplatform.yml
+++ b/app/config/ezplatform.yml
@@ -19,8 +19,8 @@ ezpublish:
     # System settings, grouped by siteaccess and/or siteaccess group
     system:
         site_group:
-            # These reflect the current installers, preferably do your prefered installation before you change these.
-            # For changing var_dir it is recommended to install clean, then change it before you start adding binary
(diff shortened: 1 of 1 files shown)


app/config/ezplatform.yml
```diff
@@ -19,4 +19,8 @@ ezpublish:
     # System settings, grouped by siteaccess and/or siteaccess group
     system:
         site_group:
+            # These reflect the current installers, preferably do your prefered installation before you change these.
+            # For changing var_dir it is recommended to install clean, then change it before you start adding binary
+            # content, otherwise you'll need to manually modify your database data to reflect this to avoid excepations.
```
bdunogier: typo on "exceptions"

bdunogier:
I'm not entirely sure it helps to customize the vardir. As said in the comment, if you want to setup multisite, you _need_ to change it.

Keeping the default config, with just `var`, doesn't make it more complicated to change it.

Or should just make `var/site` the default value in `default_settings.yml` ?

andrerom:
Closing and moving to `default_settings.yml` (kernel)

PR #81, opened 2016-01-14
EZP-25383: Change web server conf to be able to generate from bash

andrerom:
https://jira.ez.no/browse/EZP-25383

Todo:
- [x] Apache conf
- [x] Script `./bin/vhost.sh`
- [x] Nginx *(change structure for containers and adopt config variables for the script)*
- [x] Further doc update(s)

#### Not covered in script: Clustering!
For clustering as we intend to move to Symfony 3 structure where `var` is in root *(not public)*, and for instance symlink public binaries to web folder, this means we can theoretically adopt Symfony rewrite rules including `RewriteCond %{REQUEST_FILENAME} !-f` meaning it will serve file from web folder if it is there, and if not serve over app.php.

So once that is in place we can avoid having any complicated logic around clustering in web server config, and in the script added here.

/cc @vidarl @bdunogier


bin/vhost.sh
```diff
@@ -0,0 +1,189 @@
...
+    "localhost"
+    '*'
+    "80"
+    ""
+    "49152"
+    "60"
+    "48m"
+    "60s"
+    "localhost *.localhost"
+    ""
```
vidarl: One element too many... $option_values includes one element more than $template_vars

bin/vhost.sh
```diff
@@ -0,0 +1,189 @@
...
+    %HTTP_CACHE_CLASS%
+    %HOST_NAME%
+    %IP_ADDRESS%
+    %PORT%
+    %REVERSE_PROXIES%
+    %BODY_SIZE_LIMIT%
+    %TIMEOUT%
+    %BODY_SIZE_LIMIT_M%
+    %TIMEOUT_S%
+    %HOST_LIST%
```
vidarl: Is this element needed? In nginx we can just do ```server_name  %HOST% %HOST_ALIAS%;```

bin/vhost.sh
```diff
@@ -0,0 +1,189 @@
...
+# Script to generate virtual host config based on template containing variables among the once define below.
+# For help text, execute: ./bin/vhost.sh -h
+
+# Available template variables
+declare -a template_vars=(
+    %BASEDIR%
+    %CLASSLOADER_FILE%
+    %DEBUG%
+    %ENV%
+    %HOST_ALIAS%
```
vidarl: Nitpicking : I would prefer this order : ( ..., ENV, **IP_ADDRESS, PORT,HOST_NAME,HOST_ALIAS, HTTP_CACHE, HTTP_CACHE_CLASS**, REVERSE_PROXIES, ... )

doc/apache2/Readme.md
```diff
@@ -1,38 +1,143 @@
...
+    </IfModule>
+
+    # As we require ´mod_rewrite´  this is on purpose not placed in a <IfModule mod_rewrite.c> block
+    RewriteEngine On
+
+    # Environment.
+    # Possible values: "prod" and "dev" out-of-the-box, other values possible with proper configuration
+    # Defaults to "prod" if omitted. If Apache complains about this line and you can't install `mod_setenvif` then
+    # comment out "%{ENV:SYMFONY_ENV}" line below, and comment this out or set via: SetEnv SYMFONY_ENV "prod"
+    SetEnvIf Request_URI ".*" SYMFONY_ENV=prod
```
vidarl: why the SetEnv**if** .... ".*", and not simply a SetEnv ?
same SetEnvIf is in vhost.template

bin/vhost.sh
vidarl: I now see why $option_values has one more element.. Cause you inject --template-file parameter into option_values[16] .... But don't think I like this hardcoding

vidarl:
FYI : I have created https://github.com/ezsystems/ezplatform/pull/83/ based in this branch.

bin/vhost.sh
andrerom: Correct, your suggestion to use another variable is cleaner.

bin/vhost.sh
andrerom: Mybe not, it was nitpicking really to avoid generating `server_name  localhost ;` *(space after)*

doc/apache2/Readme.md
andrerom: Use of `SetEnvIf` enabled `RewriteCond %{ENV:SYMFONY_ENV} !^(dev)`.
If you know of any way to use variables set with `SetEnv` like this, then that would be the best of course to not rely on `mod_setenvif` at all.

bin/vhost.sh
vidarl: Ok. I'll keep the HOST_LIST then

doc/apache2/Readme.md
vidarl: Will not ```SetEnvIf Request_URI ".*" ....``` **always** match ( due to ".*" ).... ???

bin/vhost.sh
```diff
@@ -0,0 +1,189 @@
+#!/bin/bash
+# Script to generate virtual host config based on template containing variables among the once define below.
+# For help text, execute: ./bin/vhost.sh -h
+
+# Available template variables
+declare -a template_vars=(
+    %BASEDIR%
+    %CLASSLOADER_FILE%
+    %DEBUG%
+    %ENV%
```
vidarl: What about calling all the SYMFONY_* variables like "%SYMFONY_ENV%, "%SYMFONY_CLASSLOADER_FILE%" etc ?

doc/apache2/Readme.md
andrerom: yes but that is the intention, env is for all urls.

As said above we did not find a way to use the env variable in rewrite conditions unless we used SetEnvIf.

bin/vhost.sh
andrerom: feel free.

bin/vhost.sh
andrerom: feel free, but guess I wanted to keep variables and script arguments somewhat in sync.

vidarl:
FYI : bin/vhost.sh lacks execute permission !

andrerom:
> FYI : bin/vhost.sh lacks execute permission !

Sure? I get `-rwxr-xr-x` here locally.

vidarl:
Hmm... No exec perm when doing a ```composer --no-interaction create-project --prefer-dist --no-progress --no-interaction ezsystems/ezplatform /var/www dev-bash_vhost_script```

andrerom:
This has never been a issue before, could you check against other files?

vidarl:
Same behaviour ( missing exec perm) for all .sh files in bin/.travis/ too

If I change to ```--prefer-source```, I get exec perm ( for both vhost.sh and all .sh scripts in bin/.travis )

andrerom:
Seems you need `unzip` in the docker image: https://github.com/composer/composer/issues/4471

andrerom:
### Ready for review

For better view of the updated readme files for Apache and nginx see here:
- https://github.com/ezsystems/ezplatform/blob/bash_vhost_script/doc/apache2/Readme.md
- https://github.com/ezsystems/ezplatform/blob/bash_vhost_script/doc/nginx/Readme.md

/cc @vidarl @yannickroger @bdunogier

doc/nginx/Readme.md
```diff
@@ -0,0 +1,83 @@
...
+1. Place virtualhost config *(example below)* in suitable nginx config folder, typically:
+   - Debian/Ubuntu: `/etc/nginx/sites-enabled/<yoursite>.conf`
+   - RHEL/CentOS/Amazon-Linux: `/etc/nginx/conf.d/<yoursite>.conf`
+2. Adjust the basics to your setup:
+   - [listen](http://nginx.org/en/docs/http/ngx_http_core_module.html#listen): IP and port number to listen to.
+   - [server_name](http://nginx.org/en/docs/http/ngx_http_core_module.html#server_name): Host list, example `ez.no www.ez.no`.
+   - [root](http://nginx.org/en/docs/http/ngx_http_core_module.html#root): Point this to "web" directory of eZ installation.
+   - [fastcgi_pass](http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_pass): Socket or tcp address of `php-fpm`.
+2. Copy `ez_params.d` directory to folder you placed virtualhost config in above, examples:
+   - Debian/Ubuntu: `sudo cp -R doc/nginx/ez_params.d /etc/nginx/sites-enabled/`
```
yannickroger: you mean `sudo cp -R doc/nginx/ez_params.d /etc/nginx/` ?

doc/nginx/Readme.md
```diff
@@ -0,0 +1,83 @@
...
+        # Additional Assetic rules for environments different from dev,
+        # remember to run php app/console assetic:dump --env=prod
+        # and make sure to comment these out in "dev" environment.
+        include ez_params.d/ez_prod_rewrite_params;
+
+        # Access to repository images in single server setup 
+        include ez_params.d/ez_rewrite_params;
+
+        # upload max size
+        client_max_body_size 40m;
```
vidarl: In bin/vhost.sh, the default is 48M, not 40m... Guess we should be consistent

doc/nginx/Readme.md
```diff
@@ -0,0 +1,83 @@
...
+        location / {
+            location ~ ^/app\.php(/|$) {
+                include ez_params.d/ez_fastcgi_params;
+
+                # FPM socket
+                # Possible values : unix:/var/run/php5-fpm.sock or 127.0.0.1:9000
+                fastcgi_pass 127.0.0.1:9000;
+
+                # FPM fastcgi_read_timeout
+                fastcgi_read_timeout 90s;
```
vidarl: In bin/vhost.sh, the default is 60s, not 90s... Guess we should be consistent

vidarl:
+1 besides the comments from me and Yannick

doc/nginx/Readme.md
andrerom: aha, I assumed "include" was relative to current file, but it's relative to nginx folder then right? because that is where the main config file that included the vhost config is.

vidarl: > aha, I assumed "include" was relative to current file, but it's relative to nginx folder then right? 

Right, it is relative to nginx folder

doc/nginx/Readme.md
andrerom: ok, can bump to 90 then, thing is 90s is a somewhat safe default for dev, but not recommendation for prod.

andrerom:
updated @yannickroger, but there will be a few more fixes coming from @krzyraj soon on the doc.

andrerom:
Updated. Looks ok now @yannickroger / @bdunogier ?

web/app.php
```diff
@@ -10,13 +10,13 @@
 }
 
 // Depending on the SYMFONY_DEBUG environment variable, tells whether Symfony should be loaded with debugging.
-// If not set it is activated if in "dev" environment.
-if (($useDebugging = getenv('SYMFONY_DEBUG')) === false) {
+// If not set, or "" it is activated if in "dev" environment.
+if (($useDebugging = getenv('SYMFONY_DEBUG')) === false || $useDebugging === '') {
```
emodric: Why not test with `empty()`?

web/app.php
```diff
@@ -38,16 +38,17 @@
 }
 
 // Depending on the SYMFONY_HTTP_CACHE environment variable, tells whether the internal HTTP Cache mechanism is to be used.
-// If not set it is activated if not in "dev" environment.
-if (($useHttpCache = getenv('SYMFONY_HTTP_CACHE')) === false) {
+// If not set, or "" it is activated if not in "dev" environment.
+if (($useHttpCache = getenv('SYMFONY_HTTP_CACHE')) === false || $useHttpCache === '') {
```
emodric: Same as `$useDebugging`, why not use `empty()`?

web/app.php
andrerom: can, but don't make much difference functional wise afaik.

emodric: Well, you can use a zero as value for environment variable too then.

andrerom: which imho should mean false, and not auto pick which is what `''` means.

andrerom: I did first use the word "auto" for this, however it became more clean to use `''` for the whole chain from config generator and so on.

web/app.php
andrerom: because `0` should mean false right?

web/app.php
emodric: Ah, I didn't catch that `''` means automatic. Sorry for confusion.

yannickroger:
+1

bdunogier:
+1

PR #83, opened 2016-01-18
WIP: Nginx bash vhost script

vidarl:
Continuation of AR's https://github.com/ezsystems/ezplatform/pull/81

This PR adds nginx part of the changes + some fixes to AR's PR


doc/nginx/vhost.template
```diff
@@ -1,11 +1,8 @@
+
 server {
-    # Port to listen to, example:
-    # listen       80;
     listen       %PORT%;
-    # Host and alias to respond to, example:
-    # server_name  ez.no www.ez.no;
     # Further documentation: http://nginx.org/en/docs/http/server_names.html
-    server_name  %HOST% %HOST_ALIAS%;
+    server_name  %HOST_NAME% %HOST_ALIAS%;
```
andrerom: You can change this to:
```
server_name  %HOST_LIST%;
```
Added this variable for nginx to have a value that is proper format.

bin/vhost.sh
```diff
@@ -89,6 +96,30 @@ Arguments:
 EOF
 }
 
+# This function checks if there variables like BASEDIR, CLASSLOADER_FILE etc exists ( checks all variables defined in template_vars )
+# If environment variable exists, it's value is used as default when parsing template
```
andrerom: This part is breaking one aspect of the script, variables %BODY_SIZE_LIMIT_M%, %TIMEOUT_S% and %HOST_LIST% are internal aliases with somewhat other format to be able to inject into nginx, and not exposed as arguments. Idea is that you should be able to run the command more or less the same no matter which server you want to configure. 

With change here this changes, it means you directly use the variables in format for nginx instead of the format documented for the script.

doc/nginx/vhost.template
andrerom: Further discussion here: https://github.com/ezsystems/ezplatform/pull/81#discussion_r50015239

bin/vhost.sh
vidarl: Right. Fixed...

andrerom:
+1

PR #86, opened 2016-01-22
Fixed internal server / ezrouter.php

bdunogier:
> Fixes [EZP-25406](https://jira.ez.no/browse/EZP-25406)

Applies the ezpublish -> app rename to ezrouter.

The internal server runs fine again: `php app/console server:run -r bin/ezrouter.php 0.0.0.0:8080`


andrerom:
+1  *(I see we can remove clause for REST v1, but we can do that on merge)*

bdunogier:
Good catch about REST v1, added.

bdunogier:
Works here, I'll merge it (need it on another PR).

PR #88, opened 2016-01-25
Updated README.md (nginx)

krzyraj:


doc/nginx/Readme.md
```diff
@@ -7,12 +7,12 @@ For recommended versions of [NGINX](http://nginx.org/), see [online requirements
 Prerequisites
 -------------
 - Some general knowledge of how to install and configure NGINX
-- A working PHP FPM setup which NGINX can be configured to use as fastcgi server *(either using tcp or unix socket)*.
+- A working [PHP FPM](https://wiki.apache.org/httpd/PHP-FPM) setup which NGINX can be configured to use as fastcgi server *(either using tcp or unix socket)*.
```
andrerom: this is apache doc, should either be official php doc\*, and/or official nginx doc\*\*.

\* http://php.net/manual/en/install.fpm.php
\*\* https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/

andrerom:
merged to #81

MindTooth:
Is %HOST_LIST% removed, yet referenced from the NGINX template?  Don't respect when I add `--host-name`.

andrerom:
It's there, are you maybe using either old templates or old script?
see https://github.com/ezsystems/ezplatform/blob/master/bin/vhost.sh#L216

MindTooth:
I'm using the ones from `composer create-project ezsystems/ezplatform`.

I'll try the ones straight from source.  Though I think it doesn't matter. I only get `No input file specified.` on FreeBSD.

andrerom:
> I'll try the ones straight from source. Though I think it doesn't matter. I only get No input file specified. on FreeBSD.

how are you using it?
_Side: be aware we haven't tested it on BSD yet, so maybe you need to improve the script a bit, PR's welcome :)_

MindTooth:
Generated config with the tool.
Manually changed server_name since the tools were broke (I passed --host-name=eztest.dev, but still puts *.localhost).
Copied the ez_params.d folder inside NGINX etc folder.
Config tests successfully.
Can't access bare URL, but with /app.php, I get a page without styling.
I've been testing at work in VirtualBox.

andrerom:
Any particular iso or vagrant box used for this if we want to reproduce?

MindTooth:
ftp://ftp.freebsd.org/pub/FreeBSD/releases/amd64/amd64/ISO-IMAGES/10.2/FreeBSD-10.2-RELEASE-amd64-bootonly.iso

Installed all using pkg.  NGINX 1.8, PHP 5.6 and MariaDB 10.

PR #89, opened 2016-01-29
Update INSTALL.md to reference web/var folder for setting up permissions instead of web

emodric:
Setting permissions to the whole `web` folder might not be a good idea, as it means making `app.php` writeable by the web user, which can be a security issue.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -62,7 +62,7 @@
 2. **Setup folder rights**<a name="install-2-folder-rights"></a>:
 
        Like most things, [Symfony documentation](http://symfony.com/doc/current/book/installation.html#checking-symfony-application-configuration-and-setup)
-       applies here, in this case meaning `app/cache`, `app/logs` and `web` need to be writable by cli and web server user.
+       applies here, in this case meaning `app/cache`, `app/logs` and `web/var` need to be writable by cli and web server user.
        Furthermore, future files and directories created by these two users will need to inherit those access rights. *For
        security reasons, in production there is no need for web server to have access to write to other directories.*
 
@@ -78,8 +78,8 @@
        ```bash
        $ rm -rf app/cache/* app/logs/*
        $ HTTPDUSER=`ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1`
-       $ sudo chmod +a "$HTTPDUSER allow delete,write,append,file_inherit,directory_inherit" app/cache app/logs  web
-       $ sudo chmod +a "`whoami` allow delete,write,append,file_inherit,directory_inherit" app/cache app/logs web
+       $ sudo chmod +a "$HTTPDUSER allow delete,write,append,file_inherit,directory_inherit" app/cache app/logs  web/var
+       $ sudo chmod +a "`whoami` allow delete,write,append,file_inherit,directory_inherit" app/cache app/logs web/var
        ```
 
        B. **Using ACL on a *Linux/BSD* system that does not support chmod +a**
@@ -91,8 +91,8 @@
 
        ```bash
         $ HTTPDUSER=`ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1`
-        $ sudo setfacl -R -m u:"$HTTPDUSER":rwX -m u:`whoami`:rwX app/cache app/logs web
-        $ sudo setfacl -dR -m u:"$HTTPDUSER":rwX -m u:`whoami`:rwX app/cache app/logs web
+        $ sudo setfacl -R -m u:"$HTTPDUSER":rwX -m u:`whoami`:rwX app/cache app/logs web/var
+        $ sudo setfacl -dR -m u:"$HTTPDUSER":rwX -m u:`whoami`:rwX app/cache app/logs web/var
        ```
 
        C. **Using chown on *Linux/BSD/OS X* systems that don't support ACL**
@@ -104,9 +104,9 @@
 
        ```bash
        $ HTTPDUSER=`ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1`
-       $ sudo chown -R "$HTTPDUSER":"$HTTPDUSER" app/cache app/logs web
-       $ sudo find {app/{cache,logs},web} -type d | xargs sudo chmod -R 775
-       $ sudo find {app/{cache,logs},web} -type f | xargs sudo chmod -R 664
+       $ sudo chown -R "$HTTPDUSER":"$HTTPDUSER" app/cache app/logs web/var
+       $ sudo find {app/{cache,logs},web/var} -type d | xargs sudo chmod -R 775
+       $ sudo find {app/{cache,logs},web/var} -type f | xargs sudo chmod -R 664
        ```
 
        D. **Using chmod on a *Linux/BSD/OS X* system where you can't change owner**
@@ -115,8 +115,8 @@
        Note that this method really isn't recommended as it allows any user to do anything.
 
        ```bash
-       $ sudo find {app/{cache,logs},web} -type d | xargs sudo chmod -R 777
-       $ sudo find {app/{cache,logs},web} -type f | xargs sudo chmod -R 666
+       $ sudo find {app/{cache,logs},web/var} -type d | xargs sudo chmod -R 777
+       $ sudo find {app/{cache,logs},web/var} -type f | xargs sudo chmod -R 666
        ```
 
        When using chmod, note that newly created files (such as cache) owned by the web server's user may have different/restrictive permissions.
@@ -124,7 +124,7 @@
 
        It may also possible to add the group ownership inheritance flag so new files inherit the current group, and use `775`/`664` in the command lines above instead of world-writable:
        ```bash
-       $ sudo chmod g+s {app/{cache,logs},web}
+       $ sudo chmod g+s {app/{cache,logs},web/var}
        ```
 
        E. **Setup folder rights on Windows**
@@ -133,7 +133,7 @@
        write access to the following directories:
        - app/cache
        - app/logs
-       - web
+       - web/var
 
 
 3. **Configure a VirtualHost**<a name="install-3-vhost"></a>:


crevillo:
I agree. +1.

andrerom:
This will fail as folder is missing out of the box, as mentioned in #35 imho we should deal with this once we switch to sf 3.0 structure. As if we add the folder, it means we can't work with 5.x and 1.x in the same checkout.

crevillo:
Then we should documment it somehow... still don't see a good idea to have web webserver writable, as @emodric says.

emodric:
So move it outside of `web` right now?

crevillo:
@emodric you mean move app.php outsite web?

emodric:
@crevillo No, move `var` outside `web` folder, like in Symfony 3. That is, make it partially ready for Sf3, and when the time comes, just move `cache` and `logs` to `var` folder too.

crevillo:
ah... that's something still cannot say :)

andrerom:
@emodric might be an option, but will have impacts on kernel, something to take with @bdunogier on slack a day maybe.

emodric:
@andrerom @bdunogier Should be relatively easy:

As far as I can see, only this needs to be changed:

`ezsettings.default.io.root_dir: "%webroot_dir%/$var_dir$/$storage_dir$"`

to something like:

`ezsettings.default.io.root_dir: "%kernel.root_dir%/../$var_dir$/$storage_dir$"`

andrerom:
true, and given symfony has no notion of storage directory, we can potentially move it before we move over cache and logs for the full sf3 structure.

emodric:
Moreover, `var/site` will probably be superfluous in that case. Storage dir could live directly in `var` (`var/storage`).

andrerom:
sure, but then you need to update demo data ;)

emodric:
I know :) But it's not too much work I think.

bdunogier:
My, that's a big burden on dynamic settings. But yes, it should indeed be fine to change the vardir.

andrerom:
So then one question remains: do we rely on symlink to web folder? and what do we symlink and where?

emodric:
How about this? (`web/storage` is symlinked to `var/storage`).

```
root
  |
  |
  |---var
  |    |
  |    |--cache
  |    |
  |    |--logs
  |    |
  |    |--storage
  |
  |---web
       |
       |--storage
```

andrerom:
Would work, just need to update config, demo data and add the missing upgrade script to update path prefix for the changes *(needed anyway for when people change the settings, even better would probably be to not have the prefix in the db at all, cc @bdunogier )*

emodric:
> even better would probably be to not have the prefix in the db at all

I could go with that :) To only have `images/.../some_image.jpg` stored in the DB.

bdunogier:
> (needed anyway for when people change the settings, even better would probably be to not have the prefix in the db at all, cc @bdunogier )

With legacy out of the way, it should not be too complicated. 
The fact that images in legacy were stored with the full path actually made things much more complicated. Anyone up for it ? :-)

emodric:
> With legacy out of the way, it should not be too complicated.

I'd still like to maintain compatibility with legacy in a way not to do something that can't be configured when using ezplatform + legacy combination.

bdunogier:
> I'd still like to maintain compatibility with legacy in a way not to do something that can't be configured when using ezplatform + legacy combination.

I understand. But it also means changing eZImageAliasHandler and eZImageFile to support this.

bdunogier:
> I'd still like to maintain compatibility with legacy in a way not to do something that can't be configured when using ezplatform + legacy combination.

I understand. But it also means changing eZImageAliasHandler and eZImageFile to support this.

emodric:
Wll, I'd rather go the other way around: implementing two "handlers" or whatever in the new stack, something like image alias path generators you implemented.

bdunogier:
Hmm, but if you change how data is stored in the database, you need to change legacy itself, not the new stack.

And yes, it would be fairly easy to change that in the new stack alone.

emodric:
> Hmm, but if you change how data is stored in the database, you need to change legacy itself, not the new stack.

Well, it depends, I think. Obviously, new way would not be compatible with legacy, but that's why I said it should still be configurable where storage folder is located and what exactly is being written to the database.

We're not changing the schema, we're changing data, so it should be doable with just configuration on Symfony side and symlinks (at least I hope so).

eZ Systems of course doesn't need to support or detail compatibility with legacy. I would be perfectly fine creating subfolders inside `web` and symlinking `var/storage` wherever I need to, as long as I can configure what is being written to the database :)

But this is something that clearly needs some more thought and testing, and I'll try to do that soon :)

bdunogier:
> Moreover, var/site will probably be superfluous in that case. Storage dir could live directly in var (var/storage).

I was also kind of favor of going for this, but we discussed it with @andrerom, and this does not cope well with multi-site.

emodric:
> I was also kind of favor of going for this, but we discussed it with @andrerom, and this does not cope well with multi-site.

Why? Multisite (meaning different DB) will have to have separate var folders anyhow because of the caches. Also, separate var folders mean separate web folders too, so no symlink clashes. That's the only logical setup we could come up with.

bdunogier:
As I said, I was rather in favor of it myself.

Why would you _have_ to use multiple web folders ? If you have subdirs in var, you don't need that, afaict.

emodric:
Multisite in most cases brings different bundles being activated on different sites which means generated assets will be different for each site so it's always a good idea to keep them separate. Sometimes you will need to change the front controller (`app.php`) with stuff specific to one site and so on.

emodric:
If you use a single var folder for multisite, then the structure would have to be:

var/site/cache
var/site/logs
var/site/storage

var/site2/cache
var/site2/logs
var/site2/storage

which is not Sf 3 standard (var/cache var/logs).

andrerom:
@emodric On Domain naming here I would rather say repository, kernel, cache and logs should be separated, while sites are within a repository and actually need to share cache and logs.

emodric:
Well yes, that's exactly my point. Sites which use the same DB (repository) need to have their caches and logs shared. But sites that have the separate DBs, need to have separate not only caches and logs, but storage folder as well, and in most cases web folder too.

You can't have var/cache, var/logs, var/site/storage and var/site2/storage. It needs to be either:

```
var/cache
var/logs
var/storage
var2/cache
var2/logs
var2/storage
```

or 

```
var/site/cache
var/site/logs
var/site/storage
var/site2/cache
var/site2/logs
var/site2/storage
```

andrerom:
If your talking site/site2 => repository/repository2 in the paths then we somewhat agree. Aka I don't think this should be per "site".

bdunogier:
Should we just say that we don't support multi-sites + multi-repo within one install ?
'cause really, it's all manual :-)

Then in that case, yes, we only need one storage folder.

emodric:
> Should we just say that we don't support multi-sites + multi-repo within one install ?
> 'cause really, it's all manual :-)

> Then in that case, yes, we only need one storage folder.

I'm not so sure about it :) One of the main selling points of eZ has always been multisite.

bdunogier:
> eZ has always been multisit

It still is (you can serve multiple domains, with different layouts, from the same repository).

Given that as you demonstrated, a LOT of the application needs to be _copied_ to make multi-site + multi-repo work, I don't think we _support_ it. It is _possible_, but afaict, not something we explicitly support by means of config & features.

emodric:
> It still is (you can serve multiple domains, with different layouts, from the same repository).

I know, I meant that it was supported officially in legacy.

> Given that as you demonstrated, a LOT of the application
needs to be copied to make multi-site + multi-repo work, I don't think we support it. It is possible, but afaict, not something we explicitly support by means of config & features.

Still, it would be nice to provide a recommended setup (although not supported)

bdunogier:
> I know, I meant that it was supported officially in legacy.

It was, yes. But in my opinion, it is not a good reason to support it.
Legacy was architectured during the previous millenium (I'm almost not exagerating at all). The needs and ways of doing things were different. We loved huge apps back then.

In the meanwhile, we got virtualization, micro-services, composer... 

> Still, it would be nice to provide a recommended setup (although not supported)

Isn't a lot of it in this discussion ? :-)

andrerom:
What I meant above is that a given set of binary storage *have to* follow database, that was my whole point. Multi site and multi repo is fine, they are supported, and I did not intended to kick start a discussion on if it should not be.

bdunogier:
Did we reach a conclusion on this particular PR ?

andrerom:
-1: CLI user needs access to write to whole web folder, www-data does however not afaik.

emodric:
@andrerom Remind me again why CLI user needs access to whole `web` folder?

If this is as you say, than this PR still makes sense, right?

emodric:
I mean, the whole point was preventing web user having access to the whole `web` folder, and @andrerom, you said yourself that web user doesn't need it.

andrerom:
> @andrerom Remind me again why CLI user needs access to whole web folder?

assetic:dump, ezpublish:legacy:assets_install, but cli user should probable own the whole thing in the first place as he/she might want to do composer update as well. point is, we have some cleanup to do there, patch as is can easily casue users to not have rights to run asset commends.

emodric:
> assetic:dump, ezpublish:legacy:assets_install, but cli user should probable own the whole thing in the first place as he/she might want to do composer update as well.

Right, ofcourse, but I thought that "user should probable own the whole thing" is implied :)

> point is, we have some cleanup to do there, patch as is can easily cause users to not have rights to run asset commends.

Okay, so let's make it work then. Specify that CLI user should own it from the start and add the `setfacl` call just for web user. The way it is now is also too open.

andrerom:
Be aware some here want to remove the setfacl option, and we skipped it for docker use. 

@vidarl / @dpobel input on suggestions above?

emodric:
> Be aware some here want to remove the setfacl option, and we skipped it for docker use.

Why? In my tests and as far as I know, `setfacl` is the only one that worked really well, without modifying PHP files to include calls to `umask`.

andrerom:
I don't have that answer, hence the ping.

vidarl:
I also love `setfacl`, as chmod introduces a lot of other issues. ( you mentioned one, the fact that flysystem has hardcoded umask does not help either )

Despite that, I am still the one who removed setfacl in our docker code. This is because ACLs are not supported by most ( if any ) union filesystems. So storing ACLs in docker images simply do not work.
I don't remember the details now ( I think I did that rewrite 2 months ago and have worked on something completely different ever since .. ) but I guess we can still use facl with docker if we :
 - have a hard requirement that all folders where we use ACLs are defined as volumes ( volumes are placed directly on host's fs, not in union filesystems ). This should anyway be considered best practice
 - User need to run a script which sets corrects ACLs on those volumes.

With my rewrite, user did not have to run any scripts as permissions was "correct" ( as correct as they can be with chmod ) out-of-the-box inside the image.

This post was for your information. I guess a discussion whatever we use setfacl or not in docker do not belong to this PR....

emodric:
@vidarl Thanks for detailed explanation! But that still doesn't mean we have to remove references to `setfacl` for regular usage outside of docker, right?

When running outside of Docker, we should communicate that all files should be owned/writeable by CLI user/group from the start, and only `app/cache`, `app/logs` and `web/var` should be `setfacl`-ed to web users. That way, there would be no issues with permissions, and everything would be as secure as possible. What's your POV on that?

vidarl:
@emodric : I totally agree with you

PR #90, opened 2016-02-10
Added EzBehatExtension declaration to behat.yml.dist

miguelcleverti:
This is needed for the loading of the needed services in the behat Contexts. PR https://github.com/ezsystems/BehatBundle/pull/47 is dependent of this.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -21,6 +21,8 @@ default:
             filename: report.xml
             outputDir: %paths.base%/app/logs/junit
 
+        EzSystems\BehatBundle\ServiceContainer\EzBehatExtension: ~
+
     # default profile: no suites
     suites: ~
 
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "behat/mink-goutte-driver": "*",
         "behat/mink-selenium2-driver": "*",
         "jarnaiz/behat-junit-formatter": "^1.3",
-        "ezsystems/behatbundle": "~6.1.0@dev"
+        "ezsystems/behatbundle": "^6.2@dev"
     },
     "suggest": {
         "ezsystems/legacy-bridge": "Provides the full legacy backoffice and legacy features"


miguelcleverti:
@joaoinacio @bdunogier @andrerom

bdunogier:
Why are the builds failing ?

miguelcleverti:
@bdunogier most likely travis is fetching a tag of behat that has not the EzBehatExtension yet!

bdunogier:
Well, I'd rather see the tests pass before we merge :-)

It fetches behatbundle v6.1.0, while it needs the current master. Your base is outdated, it seems, you should pull the latest master and rebase ([list of changes on master since your branch's base](https://github.com/miguelcleverti/ezplatform/compare/addEzBehatExtension...ezsystems:master))

miguelcleverti:
@bdunogier add my remotes in git with wrong names. Pulled and rebased!

miguelcleverti:
but still the same error!

andrerom:
well, as was hinted above travis pulls in the wrong version of behat bundle. side effect of rapid releases with no script yet to make sure everything gets right on first try. So you'll need to change composer.json.

May I suggest you propose to change all ez packages from using `~` to `^`?

meaning:
```diff
-    "ezsystems/behatbundle": "~6.1.0@dev"
+    "ezsystems/behatbundle": "^6.1.0@dev"
```


This will men it will always pick latests 6.x dev package instead being looked to a minor branch, I think this is a better match for master, but it will have to be up to @bdunogier to give the verdict on that.

bdunogier:
We were using ^x.y before the 15.12.1 release, but realized that it would automatically upgrade existing enterprise users to the FTR, and it was changed to ~x.y.z to be on the safe side.

But we can change it, at least for behatbundle.

In any case, the requirements _needs_ to be changed here, since we are using a feature (from ezsystems/behatbundle) that only exists from 6.2.0.

Miguel, can you add a commit to the PR that changes the behatbundle requirement from `~6.1.0` to `^6.2`, with the message "[dependencies] set behatbundle to ^6.2" ?

bdunogier:
Ping ? Needs to be rebased and revived.

miguelcleverti:
Maybe we should close this one for now because if the the BehatExtension is moved into the kernel has proposed in the POC PR(ezsystems/ezpublish-kernel/pull/1586) then this will be outdated

andrerom:
well you can also remove change to composer.json and rebase, and theoretically this should then pass given it is part of last behatbundle tag afaik.

miguelcleverti:
Outdated and no longer applies, closing.

PR #91, opened 2016-02-15
Document use of php's built-in web server using standard symfony routes

andrerom:
Changes:
- remove custom router file as was requred when we had legacy around
- add app_dev.php as used by built in routes in symfony
- add mention in install doc *(re arrange so install is done before starting server)*

Benefit of these changes is that running server in dev env will now enable debug
(and web debug-toolbar), and we reuse the router file in symfony.

Todo:
- [x] Should not be merged before this is fixed: https://jira.ez.no/browse/EZP-25403
- ~~Ideally we should also run some of the behat/functional tests using this server.~~ *(later, will need to figure out how to run with behat env w/o adding another app file, or fix what is discussed below to avoid these extra files all together)*


emodric:
Should `app_dev.php` be readily available like this in web folder? I see a lot of future eZ Platform websites running in dev mode on production servers :)

andrerom:
> EDIT: Can't you use SYMFONY_ENV=dev php app/console server:run to achieve the same effect?

I guess so, but if we want to avoid adding this file I would rather suggest we find a way to modify server:run for this, either by extending it, or contributing to it.

emodric:
> I would rather suggest we find a way to modify server:run for this, either by extending it, or contributing to it (but features are typically not backported, so would be pita)

You can, ofcourse, but in general, I have no objection running the built in server with `SYMFONY_ENV=(prod|dev) php app/console server:run`. Moreover, since dev is the default, you should be able to run the server without the environment variable in most usecases. Environment variable would be needed only in prod mode.

So if it's documented well enough, it should be okay :)

andrerom:
> Moreover, since dev is the default, you should be able to run the server without the environment variable in most usecases. Environment variable would be needed only in prod mode.

not really, with this pr you can just set env using the ``--env=<env>`` param.
If we don't add app_dev.php, which server::run expects when ``--env=dev`` *(as you know: default)* we would have to do:
```bash
# Somewhat how people would have to run for dev
SYMFONY_ENV=dev && php app/console server:run [--env=$SYMFONY_ENV]


# How people would have to run for prod, with the warning from server:run that this is not recommended for prod
php app/console server:run --env=prod
```

emodric:
> not really, with this pr you can just set env using the --env=<env> param.

Hm, right, I get it.

I do like the new change about preventing access from non localhost clients :)

andrerom:
> I do like the new change about preventing access from non localhost clients :)

yes, I forgot about that one, it has to be there of course :)

PR #92, opened 2016-02-17
Fixed so that bin/vhost.sh correctly disables prod rewrite rules in dev mode in nginx vhost

vidarl:
In doc/nginx/vhost.template we have

    ## and make sure to comment these out in DEV environment.
    ## You need the following rewrite rule enabled in PROD environment, but disabled in DEV environment.
    include ez_params.d/ez_prod_rewrite_params;

This causes the the prod rewrite rules to still be included when generating the vhost using ```bin/vhost.sh (...) --sf-env=dev```

This PR fixes this problem

### DIFF
--- a/bin/vhost.sh
+++ b/bin/vhost.sh
@@ -28,6 +28,7 @@ template_vars=(${option_vars[*]})
 template_vars+=("%BODY_SIZE_LIMIT_M%")
 template_vars+=("%TIMEOUT_S%")
 template_vars+=("%HOST_LIST%")
+template_vars+=("%INCLUDE_PROD_REWRITE_RULES%")
 
 # Default options
 declare -a template_values=(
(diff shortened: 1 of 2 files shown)


andrerom:
-1 this makes the template less readable for those checking template as a doc resource.
Ideally we should be able to have conditions in the rules _(somewhat like apache)_, but as a last resort the script could do a replace of the string in output even if that is clearly hackish.

vidarl:
I agree that the template is less readable. However, I think it is still very much readable, and fully understandable, both the template and actuall conf file generated by vhost.sh

Conditions in the rules like apache is not possible in vanilla nginx AFAIK. Might be possible with some modules though.

IMO, the current proposal is way better than doing replace of a string on the template. That would kinda contradict the whole implementation of current vhost.sh

andrerom:
> IMO, the current proposal is way better than doing replace of a string on the template. That would kinda contradict the whole implementation of current vhost.sh

I know, but I'll prefer some mess in vhost.sh over lower readability of the template imo. @bdunogier / @yannickroger wdyt?

bdunogier:
I don't really have a strong opinion here, but can we move this forward somehow ?

vidarl:
Sorry, This PR can be closed

Fixed by https://github.com/ezsystems/ezplatform/pull/103 and https://github.com/ezsystems/ezplatform/pull/97

PR #93, opened 2016-02-17
Fix EZP-25479: better support for nginx in front of varnish

gggeek:


gggeek:
If accepted, please port this to ezpublish-kernel as well :-)

doc/varnish/vcl/varnish4.vcl
```diff
@@ -34,10 +40,18 @@ sub vcl_recv {
     // Add a unique header containing the client address (only for master request)
     // Please note that /_fragment URI can change in Symfony configuration
     if (!req.url ~ "^/_fragment") {
-        if (req.http.x-forwarded-for) {
-            set req.http.X-Forwarded-For = req.http.X-Forwarded-For + ", " + client.ip;
+        // only accept the x-forwarded-for header if the remote-proxy is trusted
+        // also we add our ip to the list of forwarders, as it is logged by apache by default
+        if (req.http.x-forwarded-for && client.ip ~ proxies) {
+            // funnily enough, it seems that this bit is executed twice, so we do a bit of dirty coding
```
yannickroger: Any idea why this happens ?

gggeek: the 'adding happens twice' ? nopes :-(
but I tested it, and the behaviour was consistent - without the workaround either no ip added or added twice

andrerom: ESI at play?

andrerom:
> If accepted, please port this to ezpublish-kernel as well :-)

you meant ezpublish-community ?

doc/varnish/vcl/varnish3.vcl
```diff
@@ -32,10 +38,18 @@ sub vcl_recv {
...
+            // funnily enough, it seems that this bit is executed twice, so we do a bit of dirty coding
+            if (req.http.X-Added-Forwarded-For) {
+                unset req.http.X-Added-Forwarded-For;
+            } else {
+                set req.http.X-Forwarded-For = req.http.X-Forwarded-For + ", " + server.ip;
+                set req.http.X-Added-Forwarded-For = "1";
+            }
         } else {
-            set req.http.X-Forwarded-For = client.ip;
+            set req.http.X-Forwarded-For = "" + client.ip + ", " + server.ip;
```
andrerom: why this part?

dspe: When you have varnish in front of eZ, if you don't do that you will receive only Varnish IP. That could be a shame if you track IP address (apache access log for example)

doc/varnish/vcl/varnish4.vcl
gggeek: not sure - we do have ESI, but this was in the request for the main page

doc/varnish/vcl/varnish3.vcl
gggeek: The original code put only the client ip in the x-forwarder-for header.

I have noticed that, unlike this, Nginx by default (in its stock reverse-proxy config) adds itself as well to the ips in x-forwarded-for.
And, in the Apache logs, 'common' format, the 1st field will contain both ips.
Which is a bit weird at 1st, but makes it actually easy to see which requests has come through the proxy and which ones not.
With this change, the apache log will contain the whole list of ips: client, nginx, varnish.

Eg:
`192.168.127.1, 172.18.0.7, 172.18.0.3 - - [17/Feb/2016:10:55:08 +0000] "GET /setup/info/php HTTP/1.1" 200 ...`

andrerom: Ok, so we should make sure it is in sync with what Symfony expects on this header then, but at least clear for me.

doc/varnish/vcl/varnish4.vcl
andrerom: > Userhash at play?

maybe, some way to log the url for debug?

gggeek: will try

gggeek: After reading the vcl a bit more, I think that this might happen because of the way the user-hash request is sent. If this is the case, we might improve the code by testing for "req.restarts" instead...
Further digging ongoing

doc/varnish/vcl/varnish3.vcl
gggeek: will do

gggeek:
Cleaned up, ready for further review and merge.
Let me know if you'd like me to rebase.

PR #94, opened 2016-03-04
EZP-25635: Added rewrite rule for assets folder

andrerom:
Adopt convention from ezplatform-demo.

Issue: https://jira.ez.no/browse/EZP-25635

### DIFF
--- a/doc/apache2/vhost.template
+++ b/doc/apache2/vhost.template
@@ -93,7 +93,7 @@
         # More info here : http://en.wikipedia.org/wiki/P3p
         RewriteRule ^/w3c/p3p\.xml - [L]
 
-        # The following rule is needed to correctly display assets from bundles, and project assets
+        # The following rule is needed to correctly display bundle and project assets
         RewriteRule ^/bundles/ - [L]
         RewriteRule ^/assets/ - [L]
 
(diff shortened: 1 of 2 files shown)


doc/apache2/vhost.template
```diff
@@ -93,8 +93,9 @@
         # More info here : http://en.wikipedia.org/wiki/P3p
         RewriteRule ^/w3c/p3p\.xml - [L]
 
-        # The following rule is needed to correctly display assets from eZ Platform / Symfony bundles
+        # The following rule is needed to correctly display assets from bundles, and project assets
```
bdunogier: "display bundles and project assets" ?

bdunogier:
+1

doc/apache2/vhost.template
gggeek: then "bundle and project assets" - the singular makes it slightly easier to understand who goes where :-)

andrerom:
Updated, will create issue and rebase for that, looks ok @dpobel ?

*Side: would potentially be nice to configure asset:install and assetic:dump to use this folder as well so we can get rid of permission rules to write to whole web dir.*

bdunogier:
> Side: As future possible follow up

JIRA ? :-)
I'm not sure I understand the link to permissions, btw.

andrerom:
> I'm not sure I understand the link to permissions, btw.

We currently need to give write permission to whole web folder, it would be nice to restrict that, but part of the sf 3.0 topic for later I guess.

dpobel:
+1

bdunogier:
+1

PR #95, opened 2016-03-07
Router behat + reorganized .travis shell scripts

dpobel:
Needed by https://github.com/ezsystems/PlatformUIBundle/pull/524

### DIFF
--- a/bin/.travis/configure_mysql.sh
+++ b/bin/.travis/configure_mysql.sh
@@ -1,4 +1,4 @@
 #! /bin/bash
 
 echo "> Create database and grant premissions to user 'ezp'"
-mysql -uroot -e "CREATE DATABASE IF NOT EXISTS behattestdb; GRANT ALL ON behattestdb.* TO ezp@localhost IDENTIFIED BY 'ezp';"
+mysql -uroot -e "CREATE DATABASE IF NOT EXISTS behattestdb CHARACTER SET utf8; GRANT ALL ON behattestdb.* TO ezp@localhost IDENTIFIED BY 'ezp';"
--- a/bin/.travis/configure_mysql.sh
+++ b/bin/.travis/configure_mysql.sh
@@ -1,0 +1,4 @@
+#! /bin/bash
+
+echo "> Create database and grant premissions to user 'ezp'"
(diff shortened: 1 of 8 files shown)


dpobel:
ping @bdunogier @andrerom

bin/.travis/prepare_system.sh
```diff
@@ -10,8 +10,7 @@
 
 # Disable xdebug to speed things up
 if [[ "$SYMFONY_DEBUG" == "" && "$TRAVIS_PHP_VERSION" != "" && "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then
```
andrerom: move this into disable_xdebug as well? otherwise you'll have to duplicate it if there is ever a env where xdebug is missing to avoid error => failure.

bin/.travis/configure_mysql.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/bash
+
+echo "> Create database and grant premissions to user 'ezp'"
+mysql -uroot -e "CREATE DATABASE IF NOT EXISTS behattestdb; GRANT ALL ON behattestdb.* TO ezp@localhost IDENTIFIED BY 'ezp';"
```
andrerom: side: should specify utf8

andrerom:
+1 besides some nitpicks.

bdunogier:
Looks good. Have you tried triggering travis on ezpublish-kernel to see how it behaves ?

andrerom:
also would be nice to change to use container here as well as part of this *(so scripts are always tested on containers here as well)*.

dpobel:
@andrerom I would prefer doing that in a second step. If we except the chmod change (I could not resist removing a chmod 777 ;)), I did not change anything what is executed, I only created *sub-*script. But I agree we should aim at reconfiguring TravisCI to use container everywhere.

@bdunogier good suggestion, done in https://github.com/ezsystems/ezpublish-kernel/pull/1606

andrerom:
ok

dpobel:
@andrerom Created a task for that: https://jira.ez.no/browse/EZP-25544

@bdunogier TravisCI seems happy on ezpublish-kernel with these changes, ok for you ?

bdunogier:
> @bdunogier TravisCI seems happy on ezpublish-kernel with these changes, ok for you ?

Yep, perfect. +1

PR #96, opened 2016-03-12
[Travis] Switch to Trusty infra

andrerom:
for:
- possible faster execution
- more stable _(less mysql/postgres/solr gone issues)_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
 
 # run tests on php 5.5 only and let unit / integration tests deal with php differences
 php:
-  - 5.5
+  - 7.0
 
 # list of paths/bundle to execute
 env:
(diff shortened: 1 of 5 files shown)


bdunogier:
The Travis error comes from an issue with the create session HTTP request. Can't say why (yet), but the object that is parsed from the HTTP response (`$this->getResponseObject()` in the REST authentication subcontext is obviously not a `Session`, hence the error.

We could at the very least check the type of object we get as a Response, and provide better feedback if it is not a session.

andrerom:
Closing in favour of #97, will 


@bdunogier What is gitcloney? _(see checks)_

bdunogier:
gitcloney might be an aging actor that sells overpriced coffee.

[gitcolony.com](gitcolony.com) on the other hand is a review-request thing that sits on top of github. Stumbled upon it while browsing the git integration directory, looked promising. Gave it a try, but disabled it for now.

andrerom:
Duno, it ran at some point on this PR.

bdunogier:
It was active for maybe half an hour.

andrerom:
Repos/forks on own account? :)
*I don't mind us trying services, was just not aware this one was tested*

PR #97, opened 2016-03-28
EZP-25876: As Maintainer I want Behat tests running on docker containers so I can run the same setup locally

andrerom:
Issue: https://jira.ez.no/browse/EZP-25876
Status: Ready for Review

Todo:
- [x] Further reduce steps done on startup in dev mode to avoid fpm not yet running once user enters url in browser *(skip deleting cache content now that we use shm for dev, strip out ez user, make sure permissions are set by image in prod and using uid for dev, check how that works on linux)*
- [x] ~~Investigate reported cache folder permission issue with the setup with Docker Toolbox on Mac~~ *Not reproduced* 
 - [x] ~~TODO: Check if this is also a issue with clean install in Docker for Mac beta.~~ *Afaik not.*

Followup:
- enable deployment of public `ezplatform-demo` docker image
- add more setups here _(incl cluster with memcached/redis and varnish, and possibility to inject config for that cleanly, like is the case for db and solr)_
 - especially cluster setup with redis and php7 would be relevant for certification use, given memcached is not stable for php7 yet.


bdunogier:
Didn't we want to use the trusty images instead ?

andrerom:
It is using trusty image, and when done you can easily run it locally as well for whenever you want to debug stuff.

bdunogier:
I can only trust your word about it :-)

Could we create a JIRA issue about it ?

andrerom:
> Could we create a JIRA issue about it ?

yes, but it's WIP, meaning outcome unknown in this case, just exploring at this point.

app/config/env.php
```diff
@@ -0,0 +1,24 @@
+<?php
+// On Symfony container compilation*, reads parameters from env variables if defined and overrides the yml parameters.
+// * For typical use cases like Docker, make sure to recompile Symfony container on run to refresh settings.
+
+
+if ($value = getenv('SYMFONY__DATABASE_DRIVER')) {
```
andrerom: Should maybe use a different convention than [How to Set external Parameters in the Service Container](http://symfony.com/doc/current/cookbook/configuration/external_parameters.html).

Reason why this cant be used as is is because defined parameters takes precedence over env parameters, which makes the feature somewhat useless:
https://github.com/symfony/symfony/issues/7555

And as setting this on run time as suggested in issue will be buggy, something ala `env.php` is more safe, however then with maybe different variable name convention.

bdunogier:
+1 as long as Travis passses in a reliable way.

bdunogier:
I've checked the behat error, and I think that somehow, Goutte (the browser used by behat) catches the 403 exception, and our custom exception catching code doesn't match.

What could be responsible for that difference ? Behat passes locally on that test.

bdunogier:
About the error, @andrerom, could you double check that the server runs with the `behat` env, and not `dev` ? According to feedback I got from @crevillo, this test fails in dev env for him. It would mean that Sf behaves a bit differently when an exception occurs, and would indicate that we are running in dev here. It would explain that failure.

crevillo:
i've made a dirty test to check that. here it goes. just modify viewAction from our ViewController in the eZ Publish Core Bundle. just add `throw new UnauthorizedException('foo', 'bar', []);` (with the proper use statment before the `return $view` part. 
Now, visit http://your-dev-env, and then visit http://your-behat-env. In my case, for dev i'm being redirected to ezplatform login page (as expected), In behat env, i get the 403 error page from Apache.

crevillo:
Some more hacking and now i found where is the problem. Afaik, Symfony only intercepts exceptions when `$useDebugging` in app.php is set to true. but actually that's seems only be true when in dev. 
https://github.com/ezsystems/ezplatform/blob/master/web/app.php#L14-L16

So i guess that setting `SYMFONY_DEBUG` env to true could make the test pass. And if not, you can always modify that to something like `$useDebugging = in_array($environment, array('dev', 'behat'));`. This is exactly what i did and now the test pass for me. 

But thinking on it, from the test written, do we really need to check if a symfony exception template will be shown? At the end, when going to production, syfmony won't intercept exceptions at all, so, do we really need that test? or in other way, could we just modify that test to check if we are getting a 403 response?

crevillo:
passing, isn't it? to me the way to solve this is just don't test that thing this way, thinking that you're not supposed to show that kind of templates in production.

bdunogier:
What this was testing was a bit special: it verifies that _when you're logged in_, you get a permission denied error if you don't have permissions. It's a distinct use-case.

There is another scenario right above this one that checks that you do get the login form if you get an unauthorized exception.

Aren't error pages meant to be customized in real life ?

About the DEBUG thing: I'd enable it for behat. It allows us to get informations about what actually failed. We could have specific tests for production specific things.

crevillo:
@bdunogier is that really? what i think it i's trying to text is if the exception is displayed and for assertubg that it tries do some xpath...
```
Then an eZ\Publish\Core\Base\Exceptions\UnauthorizedException is displayed                            # eZ\Bundle\EzPublishCoreBundle\Features\Context\ExceptionContext::anAccessDeniedExceptionIsThrown()
      Element matching css "abbr[title='eZ\Publish\Core\Base\Exceptions\UnauthorizedException']" not found.
```

bdunogier:
Yeah, I was the one who wrote that :-)

We could remove the exception text check, and verify that we get a 403 only.

bin/.docker/docker-compose.yml
```diff
@@ -0,0 +1,82 @@
...
+  ports:
+   - "8080:80"
+  environment:
+   - SYMFONY_ENV
+   - BASEDIR=/var/www
+   - MAX_BODY_SIZE=20
+   - FASTCGI_PASS=php_fpm:9000
+   - TIMEOUT=190
+   - DOCKER0NET
+  command: /bin/bash -c "cd /var/www && cp -a doc/nginx/ez_params.d /etc/nginx && bin/vhost.sh --template-file=doc/nginx/vhost.template --sf-env=behat --sf-debug=1 > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'"
```
andrerom: `--sf-env=behat --sf-debug=1` it temporary:
1. at the very least tests should pass regardless of debug like @crevillo says
2. however if we pass in SYMFONY_ENV and SYMFONY_DEBUG above, this should be taken into acount so some of the tests can test using debug, in essence the whole vhost.sh approach might need some rethinking/adjustments.

bdunogier: > 1. at the very least tests should pass regardless of debug like @crevillo says
Sure thing. Remember that at some point we were complaining that behat failures were very hard to understand and identify. That's why this kind of test was added.

Could `behat` to the lists of environments that require debugging, in `app.php` ? The fact _is_ that behat tests require debugging, so we might as well make it official.

Or we can tweak the exception check to only look for exception text if it should be here (either check if debugging is enabled, or check for Sf HTML output we would get on an exception, and skip if it isn't there.

andrerom: > We should in any case tweak the sentence. The exception isn't what we are really testing, as described above.

I think we should do that.
And in parallel I'll look for ways to make sure containers pick up env and debug settings w/o having to hard code it in the docker-compose file.

andrerom:
Found a way to avoid hardcoding the env and debug in docker-composer.yml.

Review ping @bdunogier @vidarl, see ecffde5: This would imply `vhost.sh` is not needed anymore, unless we want to continue have a script that normilizes params across nginx, apache and so on. If so it will need adaptation to first and foremost takes params from env and just take arguments if they are set, however w/o like here there is less to learn basically.

andrerom:
@vidarl Second approach: https://github.com/ezsystems/ezplatform/compare/60caa32...fd2d9621e6812256db55ed783d1402d078fdb5f0**

vidarl:
Looks good !
+1

vidarl:
On second thoughts.... 
Some improvements + bugfixes here:
https://github.com/ezsystems/ezplatform/pull/103

What about adding some tests for vhost.sh? Changing it easily causes regressions....

andrerom:
Ok, I think we have something worth reviewing now, and as it is vastly different then what this PR started with I have done some spring cleaning *(commits can be squashed on merge)*.

Some things we might want to consider:
- place the docker-compose files in a separate repo and use git submodules or something, they will grow


As for doc changes, see inline, but also https://github.com/ezsystems/ezplatform/blob/f757a3ad6eb583383f8054127053d2a9b3c647f4/README.md#install

review ping @bdunogier @DavidLiedle

.dockerignore
```diff
@@ -0,0 +1,58 @@
...
+
+# PHPUnit
+/app/phpunit.xml
+/phpunit.xml
+
+# Build data
+/build/
+
+# Composer PHAR
+/composer.phar
```
galileo: Why do we need it here? Isn't composer installed globally on container?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# Do composer install, remove cache, fix owner, and remove auth.json if REMOVE_AUTH=1
+RUN composer install --optimize-autoloader --no-progress --no-interaction --prefer-dist \
+ && rm -Rf app/logs/* app/cache/*/* .git/ web/var \
+ && mkdir web/var \
+ && chown ez:ez -R /var/www \
+ && [ "$REMOVE_AUTH" = "1" ] && rm -f auth.json
+
+# Declare volumes so it an can be shared with other containers
+# Also since run.sh will use setfacl, and that does not work on aufs (but volumes does not use that)
+VOLUME /var/www /var/www/web/var
```
galileo: The `/var/www/web/var` is underneath the `/var/www` volume, do you need to add the `/var/www/web/var` volume explicitly? If yes, what is the reason?

andrerom: as it might be a different mount point *(nfs, as needed for [Cluster](https://doc.ez.no/display/TECHDOC/Clustering) setup)*.

.dockerignore
andrerom: Indeed, which is why we don't need composer.phar which might be downloaded by user inside the container.

andrerom:
Update on docker stuff for DEV use:
- docker for mac 1.11.1-beta13.1 somewhat fixes the performance issues by caching file system reads
- writes are still slow so used tip to put Symfony cache and logs on `/dev/shm`
- Also dumping composer autloads has a noticable positive impact in this setup 
- load time is down to 2 seconds _(1 second if opcache is tuned to not check if file has changed)_, once misc caches has been filled
- But Platform UI is to slow, so looking into loading assets via nginx despite being in dev, and maybe also find a way to run platform UI + REST calls in prod while frontend is dev which covers 90% of the use case for now.

andrerom:
@bdunogier besides still slow on shared folder setup, all other remarks _(on chat and here)_ have been taken into account, so ready for review.

andrerom:
Splitted out work on trying to get this to work for cross platform for dev in separate PR: #117

doc/docker-compose/Readme.md
```diff
@@ -0,0 +1,134 @@
...
+
+*For Windows you'll also need to [install bash](https://msdn.microsoft.com/en-us/commandline/wsl/about), or adapt instructions below for Windows command line where needed.*
+
+
+## Production / Demo use
+
+From root of a local download/clone of this distribution, then execute the following:
+
+*First step will do a composer install as part of the build process, for this to work smoothly create auth.json file in project root with your
+readonly GitHub token and if you use eZ Enterprise software also [tokens for updates.ez.no](https://doc.ez.no/display/TECHDOC/Using+Composer).*
```
bdunogier: Aren't we phasing TECHDOC out in favour of DEVELOPER ?

doc/docker-compose/Readme.md
```diff
@@ -0,0 +1,134 @@
...
+docker-compose exec --user www-data app /bin/sh -c "php /scripts/wait_for_db.php; php app/console ezplatform:install clean"
+```
+
+
+At this point, you should be able to browse the site on `localhost:8080` and the backend UI on `localhost:8080/ez`.
+
+
+## Behat use
+
+From root of a local download/clone of this distribution, then execute the following:
```
bdunogier: The "then" seems out of place.

doc/docker-compose/Readme.md
```diff
@@ -0,0 +1,134 @@
...
+
+*First step will do a composer install as part of the build process, for this to work smoothly create auth.json file in project root with your
```
bdunogier: Given that this part is repeated three times in total, maybe it could have its own paragraph in the document, with an anchor ?

### Configuring composer

For composer to run smoothly, we recommend that you create an `auth.json` file in your project root with your github readonly token:

```
example here
```

If you use eZ Enterprise software, add your [tokens for updates.ez.no](https://doc.ez.no/display/DEVELOPER/Using+Composer) as well.

doc/docker-compose/Readme.md
andrerom: I'll fix, this was written before that.

doc/docker-compose/Readme.md
andrerom: added, with examples

glye:
+0.1, looking forward to use it though.

doc/docker-compose/Readme.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Installing Docker
+
+Before jumping into steps below for either production / demo or development use, make sure you have recent versions of
+[Docker & Docker-Compose](https://www.docker.com/) installed on your machine.
+
+*For Windows you'll also need to [install bash](https://msdn.microsoft.com/en-us/commandline/wsl/about), or adapt instructions below for Windows command line where needed.*
+
+
+## Production / Demo use
```
bdunogier: Do we want to skip doc for prod and dev use, given the now restricted scope of this particular PR ?

bdunogier:
Added a remark about the doc (maybe we can shrink it since the scope was shrunk to behat), but that aside, +1.

andrerom:
Thanks, added info about this being beta and linked to online page further meant to describe that now.

PR #98, opened 2016-03-28
Fix EZP-25628: Reduce the amount of headers sent by Varnish

gggeek:


doc/varnish/vcl/varnish3.vcl
```diff
@@ -213,12 +213,22 @@ sub vcl_deliver {
     // Sanity check to prevent ever exposing the hash to a client.
     unset resp.http.x-user-hash;
 
+    // Remove cache-control headers on content pages, as we do not want downstream proxies/caches to cache these
+    if (req.url !~ "\.(css|js|gif|jpe?g|bmp|png|tiff?|ico|img|tga|wmf|svg|swf|ico|mp3|mp4|m4a|ogg|mov|avi|wmv|zip|gz|pdf|ttf|eot|wof)$") {
+        unset resp.http.cache-control;
```
emodric: Why wouldn't we? Why disable browser the possibility to cache images for example with `max-age` config?

gggeek: The reasoning is:
- I have varnish enabled
- I set a TTL on content of 1 year *BECAUSE* I can send ban/purge commands to varnish to refresh content
- user X visits site, views a page displaying content A
- user Y does NOT visit the site
- editor publishes new version
- varnish gets the ban command and clears its cache
- browser of user X does not get notified
- user X does not see new version of A
- user Y visits the the site, sees new version of A

This leads to annoying problems:
- hard to troubleshoot because different users will see different contents
- even if you replace '1 year' with '6 hours', the problem remains the same: end users will see stale content
- the only 'fix' that I know is to set the max-age to 5 minutes, but then what is the point of varnish?

The root of the problem is that we should be able to send different cache-control headers:
- one to be obeyed by varnish
- one to be obeyed by downstream proxies and user-agents

emodric: That all may be true, but the fact still remains that you remove headers that developer explicitly set in the response object and this is not something that should be done by default.

gggeek: If you prefer, we can add this line as a comment, so that devs will see it and decide whether they want to enable it or not.

But I am still not in agreement about the 'default'. To me the default way of using Varnish with eZ is the one I described above, and my proposal fits it better. Devs who want to add caching headers with destination end users browsers as opposed to destination varnish are probably a minority, and should understand fully this 'double-caching-plus-ban' mechanism, so the onus is on them.

emodric: > If you prefer, we can add this line as a comment, so that devs will see it and decide whether they want to enable it or not.

That would be better IMO

andrerom:
Closing this as fixed by #216

PR #99, opened 2016-03-28
Add to Apache conf expiry headers for assets

gggeek:
This PR also replaces usage of mod_expires with mod_headers. The benefit is that it makes it easier to take this vhost file and turn it into an .htaccess file

### DIFF
--- a/doc/apache2/vhost.template
+++ b/doc/apache2/vhost.template
@@ -69,6 +69,38 @@
     <IfModule mod_rewrite.c>
         RewriteEngine On
 
+            # Everything in this block is optional to improve performance by forcing
+            # clients to cache image and design files; do change the max-age time
+            # to suit project needs.
+            # Note: the slightly convoluted syntax adopted has the advantage of
+            # not requiring usage of the LocationMatch directive.
+            # This makes it ideal for .htaccess based configurations, beside vhost ones.
+            <IfModule mod_headers.c>
+                RewriteCond %{REQUEST_URI} ^/var/[^/]+/storage/images/.*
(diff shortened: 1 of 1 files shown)


bdunogier:
Not sure what to think. Ping @andrerom & @vidarl maybe ?

vidarl:
I haven't tested gggeek's code but it looks sane.

However, it does add (unneeded ? ) complexity to the vhost config. And putting this in a .htaccess is not something you anyway would do in production. This is only applicable for testing/dev IMO, unless you have a odd edge case..

Therefore I would vote against this change

gggeek:
@vidarl about .htaccess and production: you'd be surprised at how many sysadmins I recently found who actually encourage usage of it, esp. in the case of 'externally managed systems', which seems to be quite common for big companies and for cloud deployments. Setting up an .htaccess means that the sysadmin keeps control over the 'main' server config, and the devs do not need to poke him every 30 secs for small changes needed for the application.

But I can redo the same PR with the rest of the changes applied, and just a commented-out example introduced for the rewrite-based expiration rules, if you prefer

andrerom:
This makes it really hard to read the expiry rules, besides this would need to be applied to the template and all examples would need to be regenerated to reflect it.

So -1 on approach I think, new PR with expiry rules on assets folder would be nice though.

gggeek:
@andrerom sorry, what templates?

vidarl:
@gggeek : André is referring to https://github.com/ezsystems/ezplatform/blob/master/bin/vhost.sh and templates like https://github.com/ezsystems/ezplatform/blob/master/doc/nginx/vhost.template and https://github.com/ezsystems/ezplatform/blob/master/doc/apache2/vhost.template

PR #100, opened 2016-03-31
[travis] removed -v option from behat command line

bdunogier:
Causes an error on builds, with no specs being found.
See https://travis-ci.org/ezsystems/ezplatform/builds/116877568.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_script:
 
 # execute behat as the script command
 script:
-  - php bin/behat -vc behat.yml.dist $ARGS --tags='~@broken'
+  - php bin/behat -c behat.yml.dist $ARGS --tags='~@broken'
 
 # disable mail notifications
 notifications:


andrerom:
+1 ish, but don't know why

bdunogier:
Same here...

Could be that verbosity triggers some extra behaviour that fails, and that failure has this side effect. I've seen an issue on the behat tracker where this error would pop up because of an inconsistent parameters combo.

glye:
+1, as -v is for version number, not verbosity according to this: http://docs.behat.org/en/v2.5/guides/6.cli.html
Still doesn't really explain the error, but anyway we have no need to request the version number here.

andrerom:
Wrong version ;)

Here is help text for v3.0:
```
  -v, --verbose                          Increase verbosity of exceptions.
  -h, --help                             Display this help message.
      --config-reference                 Display the configuration reference.
  -V, --version                          Display this behat version.
```

Is this blocking you then please just merge this, I haven't seen this recently so thought it was gone again.

PR #102, opened 2016-04-05
[Behat] Remove Behat failure on deprecation warnings

miguelcleverti:
This PR configures Behat so it does not fail when a E_ALL and ~E_USER_DEPRECATED php error occur.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -1,5 +1,7 @@
 # This file contains the default configuration for behat testing using EzSystems/BehatBundle
 default:
+    calls:
+        error_reporting: 14335 # E_ALL & ~E_USER_DEPRECATED
     extensions:
         Behat\MinkExtension:
             base_url: 'http://localhost'


bdunogier:
+1. A custom error handler that lists deprecated features that are used would be convenient, but this is a good step forward.

Ping @andrerom and... @dpobel ? :-)

micszo:
@miguelcleverti do you thing we could also suppress the notices? :) It makes the output hard to read and log size will be affected.

screen shot 2016-04-06 at 09 37 02

I found something about ignoring notices with 32759 but still experimenting w/o success (http://stackoverflow.com/questions/31050716/how-to-run-behat-tests-when-there-are-errors-of-level-e-user-deprecated)

bdunogier:
See this PR @micszo : https://github.com/ezsystems/ezpublish-kernel/pull/1633. It does suppress the warning it seems. What annoys me is that it is supposed to be the _same_ call... see https://github.com/Behat/Behat/blob/master/features/error_reporting.feature (although that scenario doesn't say if the deprecation notice is completely hidden or if it just passes).

bdunogier:
According to http://stackoverflow.com/questions/31050716/how-to-run-behat-tests-when-there-are-errors-of-level-e-user-deprecated/31059663#31059663, both are required...

micszo:
Thanks @bdunogier , I wasn't aware of the PR in kernel. :) (also had the impression that it's possible to pack it all into that integer)

bdunogier:
@micszo I agree... according to at least two different sources, it is not sufficient.

If it was part of the bootstrap file used by behat, maybe it would work. But we don't have a dedicated one for Behat do we ?

andrerom:
+1, even if I also would like this to be optional somehow, for instance by using behat profiles that inherits settings from default profile.

PR #103, opened 2016-04-11
[vhost.sh] Fixed conditional commented lines

vidarl:
Fixes for vhost.sh

This is based on AR's travis-docker branch and I use the same concepts.

Changes:
- Removes "if[var...]" in output ( lines remains ordinary comments without our "if" statement
- Fixed "#if[VAR!=some-other-value] " code so it now works with multiple occurences

### DIFF
--- a/bin/vhost.sh
+++ b/bin/vhost.sh
@@ -236,24 +236,41 @@ while [  "${template_vars[$COUNTER]}" != "" ]; do
     # Replace %VAR% with the actual value
     tmp=${template//${current_var}/${current_value}}
 
+    # Remove "%" from VAR for further use
+    current_var=${current_var//%/}
+
     # If variable has a value then do further replacment logic
     if [ "$current_value" != "" ] ; then
-        # Remove "%" from VAR for further use
-        current_var=${current_var//%/}
 
-        # Remove "#if[VAR] " comments to conditionally uncomment lines
+        # Change "#if[VAR] " comments conditionally to uncommented lines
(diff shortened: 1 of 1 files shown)


bin/vhost.sh
```diff
@@ -236,24 +236,42 @@ while [  "${template_vars[$COUNTER]}" != "" ]; do
...
+    # Remove "%" from VAR for further use
+    current_var=${current_var//%/}
+
     # If variable has a value then do further replacment logic
     if [ "$current_value" != "" ] ; then
-        # Remove "%" from VAR for further use
-        current_var=${current_var//%/}
 
-        # Remove "#if[VAR] " comments to conditionally uncomment lines
+        # Change "#if[VAR] " comments conditionally to no-comment lines
```
andrerom: never heard the term "no-comment" about uncommenting a line :)

andrerom:
+1 besides maybe some improvements to the inline comments.

bin/vhost.sh
vidarl: I didn't feel "uncomment lines" sounded that good either, but "no-comment lines" was maybe not an improvement.

You want it back to "uncomment lines", or is "uncommented lines" better as in :
`+        # Change "#if[VAR] " comments conditionally to uncommented lines`

andrerom: Your new proposal looks good.

vidarl:
> +1 besides maybe some improvements to the inline comments.

Fixed

PR #104, opened 2016-04-12
Fix EZP-25653: Use eZSupportTools in PlatformUI

glye:
> Implement https://jira.ez.no/browse/EZP-25653
> Status: Ready for review
> PlatformUI PR: https://github.com/ezsystems/PlatformUIBundle/pull/545

Adds eZSupportTools to app kernel, main routing, and requirements. Require dev branch of platform-ui-bundle.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,8 @@
         "hautelook/templated-uri-bundle": "~1.0 | ~2.0",
         "doctrine/doctrine-bundle": "~1.3",
         "symfony/expression-language": "~2.4",
-        "sensio/framework-extra-bundle": "~3.0"
+        "sensio/framework-extra-bundle": "~3.0",
+        "ezsystems/ez-support-tools": "~0.1.0@dev"
     },
     "require-dev": {
         "behat/behat": "~3.0",
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -45,6 +45,7 @@ class AppKernel extends Kernel
             new EzSystems\PlatformInstallerBundle\EzSystemsPlatformInstallerBundle(),
             new EzSystems\RepositoryFormsBundle\EzSystemsRepositoryFormsBundle(),
             new EzSystems\EzPlatformSolrSearchEngineBundle\EzSystemsEzPlatformSolrSearchEngineBundle(),
+            new EzSystems\EzSupportToolsBundle\EzSystemsEzSupportToolsBundle(),
             new AppBundle\AppBundle(),
         );
 
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -26,6 +26,3 @@ _ezpublishPlatformUIRoutes:
 
 _ezplatformRepositoryFormsRoutes:
     resource: "@EzSystemsRepositoryFormsBundle/Resources/config/routing.yml"
-
-_ezSupportToolsRoutes:
-    resource: "@EzSystemsEzSupportToolsBundle/Resources/config/routing.yml"
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -26,3 +26,6 @@ _ezpublishPlatformUIRoutes:
 
 _ezplatformRepositoryFormsRoutes:
     resource: "@EzSystemsRepositoryFormsBundle/Resources/config/routing.yml"
+
+_ezSupportToolsRoutes:
+    resource: "@EzSystemsEzSupportToolsBundle/Resources/config/routing.yml"


composer.json
```diff
@@ -30,13 +30,14 @@
         "ezsystems/ezpublish-kernel": "~6.2@dev",
         "ezsystems/repository-forms": "~1.1@dev",
         "ezsystems/ezplatform-solr-search-engine": "~1.0@dev",
-        "ezsystems/platform-ui-bundle": "~1.2@dev",
+        "ezsystems/platform-ui-bundle": "dev-ezp25653_use_ezsupporttools_in_platformui as v1.3.0-dev",
```
glye: > in your ezplatform branch, set the requirements for support tools and platformui using dev-branchname as vX.Y.Z-dev

@bdunogier Thanks - done that, stuck on ´The requested package (...) exists as (...) but those are rejected by your constraint´.
But given what @andrerom [said](https://github.com/ezsystems/PlatformUIBundle/pull/542#discussion_r59348253) about the freeze, there's no need to spend your vacation helping me out here :)

andrerom: Unless its ready for alpha tag now and beta next week, cause if its not ready for that there are a few support issues with your name on it GL :)

glye: @andrerom That's up to you and @bdunogier... it's done, apart from getting composer to cooperate. Presumably simple if you know how :)

bdunogier:
Fixed dependencies (`v0.1.0-dev` instead of `~0.1.0@dev`), as well as platform-ui-bundle's branch name.

It installs now, but the sysinfo page doesn't work yet, it seems. For one, the loading order of bundles probably overwrites what PUI defines.

glye:
Thanks, install indeed works now, and the page (still) works for me.

emodric:
I'm not able to test right now, but are the routes imported here available to anonymous users? Because they should not be. Anonymous should not be able to see info about the system, ofcourse.

If they are, the issue is practically the same as with ez repo forms bundle and it's content edit form: https://jira.ez.no/browse/EZP-25447

bdunogier:
The epic for support tools includes a story for permission checking.

bdunogier:
Edit: tested permissions, I'm not able to access the route without being logged in as admin (same checks than in the previous iteration).

emodric:
Well, they can't be accessed only beacuse view type is not set to `full`, but `pjax_tab`.

Set the view type to full (for example if someone wants to display these pages outside Platform UI), and suddenly all these links are available to anonymous. Example:

http://ezplatform.local/ez-support-tools/view/database
http://ezplatform.local/ez-support-tools/view/symfony_kernel
http://ezplatform.local/ez-support-tools/view/composer

It could be a tricky situation for someone who would expect that these URLs with sensitive content are available only to admins by default (which they are, but purely coincidentally).

bdunogier:
The route+controller from support-tools is another story indeed (and another repo, as a matter of fact). The controller from PlatformUI _is_ protected.

To be honest, I was thinking of ditching the route. We only need the controller, for the time being.

emodric:
> and another repo, as a matter of fact

Sure, but it's imported here :)

> The controller from PlatformUI is protected.

I know, I wasn't referring to Platform UI controller at all :) My original question: but are the *routes imported here* available to anonymous users? :)

glye:
> are the routes imported here available to anonymous users?

So it seems, thanks for noticing. In my brief test they are not available outside of platformui context because the templates are in platformui, so I get an exception, but I will not go so far as to call that access control :)

bdunogier:
@glye: i think we should a) remove the system_info_view route b) remove the import from ezplatform. We do not need nor use them for the time being.

Security should, as described in the story, be taken care of in the SystemInfo service layer, tha does not existe yet.

Without the routes, system info are only exposed by the platformui controller, and access control is covered there.

glye:
Removing the unused system_info_view route:
https://github.com/ezsystems/ez-support-tools/pull/16
https://github.com/ezsystems/ezplatform/pull/106

glye:
The travis/behat error should be fixed by https://github.com/ezsystems/ezplatform/pull/100

composer.json
```diff
@@ -30,13 +30,14 @@
         "ezsystems/ezpublish-kernel": "~6.2@dev",
         "ezsystems/repository-forms": "~1.1@dev",
         "ezsystems/ezplatform-solr-search-engine": "~1.0@dev",
-        "ezsystems/platform-ui-bundle": "~1.2@dev",
+        "ezsystems/platform-ui-bundle": "dev-dev-ezp25653_use_ezsupporttools_in_platformui as v1.3.0-dev",
```
andrerom: *side: before we merge we should of course remove this, implying we merge in Platform UI first, and then here.*

andrerom:
+1 *(but with cleanup and merge of ezsystems/PlatformUIBundle#545 first, then what is said above before merge)*

PR #105, opened 2016-04-13
Fix EZP-25638: Editing content types when default siteaccess only uses nor-NO as language

glye:
> Doc fix for https://jira.ez.no/browse/EZP-25638
> Status: Ready for review

In the immortal words of Boromir: "One does not simply remove eng-GB from the site languages list."

### DIFF
--- a/app/config/ezplatform.yml
+++ b/app/config/ezplatform.yml
@@ -23,4 +23,7 @@ ezpublish:
             # it is recommended to install clean, then change setting before you start adding binary content, otherwise you'll
             # need to manually modify your database data to reflect this to avoid exceptions.
             var_dir: var/site
+            # System languages. Note that by default, content, content types, and other data are in eng-GB locale,
+            # so removing eng-GB from this list may lead to errors or content not being shown, unless you change
+            # all eng-GB data to other locales first.
             languages: [eng-GB]


andrerom:
+1 however should we be doing something more here as well? _(better error reporting in UI, clear way for how to manage to remove a locale, ..)_

yannickroger:
+1

glye:
> better error reporting in UI

Yes, I can add a story for showing error when no data found in specified language

> clear way for how to manage to remove a locale

I wouldn't doc that unless we add proper functionality for it, since right now it means doing lots of manual, error-prone sql queries.

> add support in installer for selecting language to install in

Yes, I'll add a story for that too.

andrerom:
>> better error reporting in UI
> Yes, I can add a story for showing error when no data found in specified language

or UI should detect this. could you ping me on skype with info on where this fails?


>> clear way for how to manage to remove a locale
> I wouldn't doc that unless we add proper functionality for it, since right now it means doing lots of manual, error-prone sql queries.

I was thinking of either a command or something to remove a locale, not manual sql queries.
The language UI should also refuse to delete a language if in use as well if it doesn't already.

>> add support in installer for selecting language to install in
> Yes, I'll add a story for that too.

Note that there probably is one already for improving installer.

glye:
> The language UI should also refuse to delete a language if in use as well if it doesn't already

It does, this seems ok: `Argument 'language' is invalid: Deleting language logic error, some content still references that language and therefore it can't be deleted`

andrerom:
ok, well merging this one and make sure the other enhancement on content type language is handled as part of this issue before sending to QA _(for instance change it to a sub task)_.

PR #109, opened 2016-04-26
[Behat] Add EzBehatExtension in behat.yml.dist

miguelcleverti:
Add to behat.yml.dist the [EzBehatExtension](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/PlatformBehatBundle/ServiceContainer/EzBehatExtension.php) which enables the usage of the [ArgumentResolver](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/PlatformBehatBundle/Context/Argument/AnnotationArgumentResolver.php) so that we can inject services in Behats contexts see [Accessing eZplatform Kernel Services](https://github.com/ezsystems/ezpublish-kernel/tree/master/doc/behat#accessing-ez-platform--kernel-services)

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -23,6 +23,8 @@ default:
             filename: report.xml
             outputDir: %paths.base%/app/logs/junit
 
+        EzSystems\PlatformBehatBundle\ServiceContainer\EzBehatExtension: ~
+
     # default profile: no suites
     suites: ~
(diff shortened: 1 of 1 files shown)


bdunogier:
+1

andrerom:
+1

PR #110, opened 2016-05-08
[WIP] Docker Compose stack: Local env to dev in (dev/prod) mode and Behat ready.

Plopix:
Hello here,

This this a PR that should help to use eZ Platform and Docker really easily

My ultimate goal was:
1. git clone
2. compose up
3. **Done!**

=> It should be as simple as that.

After playing I have seen  that pull request about Travis and the behat tests. https://github.com/ezsystems/ezplatform/pull/97
Then this PR could replace the #97... depending on your opinion ;)

And I also saw the docker-php image. I did not really like it because to me we should not do too much scripting in the PHP image
https://github.com/ezsystems/docker-php/tree/updates

Then because it is always good to share, here is my PR ;-)

I am sure there is a lot to optimize and some tricky things to do, but it works and more important I think my approach it simpler and with less code.
But it could have some disadvantages that I did not see.
I tried to reduce at the maximum the scripting.

I have used very simple image and tried to follow the same name convention as the PR #97
You will find a container architecture schema in PUML here, in short : https://github.com/Plopix/ezplatform/tree/docker/doc/docker

- appvol: Data Volume Container that mounts the hosts and run the first install
- db: the database
- engine: instead of php-fpm, I mean for now, it was and it is more then just FPM, we have php-cli, we could have mysql client in the futur etc..
- web: for nginx
- memcache: for Memcached
- varnish
- mailcatcher
- memcachedadmin

2 Vhosts are provided by (see the diagram) DEV and PROD over Varnish

The installation/usage is described and started here: doc/docker/Readme.md

**Todos**:
- [ ] Create an eZ Systems php image to replace plopix/docker-php-ez-engine
- [ ] Manage a user and some permissions because right now root is the owner.
- [ ] Optimize, improve the documentation for the Docker Machine users
- [ ] We should improve the native configuration of Memcache currently it is done with a echo in the ezplatform.yml
- [ ] We should use the Docker WORKDIR concept
- [ ] Any ideas?

**Travis**
I need help with Travis here, I am not sure why it does not work as expected
The installation is working but the tests fail.(same in local)

**Todos +++**:
- [ ] install HAProxy and give the ability to scale web and engine (but will need a Cluster config)

Let me know what you think!
++
Architecture

### DIFF
--- a/bin/.docker/docker-compose.yml
+++ b/bin/.docker/docker-compose.yml
@@ -1,0 +1,110 @@
+version: '2'
+services:
+    # DB Container
+    db:
+        image: mariadb:10.0
+        ports:
+            - "${PORT_PREFIX}306:3306"
+        environment:
+            - "MYSQL_ROOT_PASSWORD=mysecretrootpassword"
+            - "MYSQL_USER=${DATABASE_USER}"
+            - "MYSQL_PASSWORD=${DATABASE_PASSWORD}"
+            - "MYSQL_DATABASE=${DATABASE_NAME}"
+            - "TERM=dumb"
+
+    # Kind of first container that will start the composer install and the db install from the source
+    # it needs php cli, right now we are taking the php.ini of the "engine" container.
+    appvol:
+        # @todo: Should be a clean eZ PHP image
+        image: plopix/docker-php-ez-engine
+        volumes:
+            - "${ROOT_PATH}/bin/.docker/engine/php.ini:/usr/local/etc/php/php.ini:ro"
+            - "${ROOT_PATH}:/${CONTAINER_PROJECT_DIR}:rw"
+            - "${ROOT_PATH}/bin/.docker/appvol/entrypoint.bash:/entrypoint.bash:ro"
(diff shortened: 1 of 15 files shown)


andrerom:
Todo:
- Need to rebase on top of existing branch
- need to adjust approach so it works with our existing stuff.

bin/.docker/docker-compose.yml
```diff
@@ -0,0 +1,111 @@
...
+        environment:
+            - "MYSQL_ROOT_PASSWORD=mysecretrootpassword"
+            - "MYSQL_USER=${DATABASE_USER}"
+            - "MYSQL_PASSWORD=${DATABASE_PASSWORD}"
+            - "MYSQL_DATABASE=${DATABASE_NAME}"
+            - "TERM=dumb"
+
+    # Kind of first container that will start the composer install and the db install from the source
+    # it needs php cli, right now we are taking the php.ini of the "engine" container.
+    appvol:
```
andrerom: why separate appvol if it is full php image anyway

PR #111, opened 2016-05-21
Handle Docker Machine users when using NFS

Plopix:
Provide a way to manage the folder mapping when you are using a Docker Machine:

- on windows
- on mac because Docker for Mac is too slow on shared folder


bin/.docker/docker-compose.dev.yml
```diff
@@ -8,7 +8,7 @@ services:
     #image: ezsystems/php:${PHP_VERSION}-v0-dev
     image: ezsystems/php:${PHP_VERSION}-v0
     volumes:
-     - ../../:/var/www
+     - $COMPOSE_DIR/../../:/var/www
```
andrerom: hmm this is rather easily confusing since $COMPOSE_DIR refers to project root, given `.env` is in root, so this looks like it is two folders below project root basically when seeing the code.

Plopix: I agree it could be confusing but I don't know if we have another  choice.

by COMPOSE_DIR I mean the docker-compose.yml file, and in this case it is not the project root dir but ./bin/.docker.

Maybe we should rename it.

andrerom:
I didn't really have issues with this at all here, so not sure why this would be needed.

What I do have issues with is permissions when running dev, and only fix I found was to run/exec with root and set UID of www-data to 1000 as suggested here to map to docker:staff VM user: https://github.com/boot2docker/boot2docker/issues/587#issuecomment-114868208

Plopix:
I am using Docker Toolbox but with VMWare for the Docker Machine. I am not using any shared folder system(too slow). Instead, my docker machine mounts my Mac OSX system via NFS. In this configuration(that has good performances) when I run a docker-compose from my Mac it does not work as docker-compose can not resolve the relative path on the Mac. Which to me is logical because from the Docker point of view (on the virtual machine), the files are on a mount point.

Here is the error if I use a relative path:
```
ERROR: for install  stat /Users/Plopix/DOCKFILES/PLOPIX/eZ-Platform/ezplatform: protocol error
Traceback (most recent call last):
  File "/usr/local/Cellar/docker-compose/1.7.1/libexec/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.7.1', 'console_scripts', 'docker-compose')()
  File "/usr/local/Cellar/docker-compose/1.7.1/libexec/lib/python2.7/site-packages/compose/cli/main.py", line 63, in main
    log.error(e.msg)
AttributeError: 'ProjectError' object has no attribute 'msg'
```

Indeed on the Docker Machine, the path  _/Users/Plopix/DOCKFILES/PLOPIX/eZ-Platform/ezplatform_ does not exist.
In my case, it is another one(according to my mounts)

Actually, I think that VirtualBox abstracts that completely and adds a mount point that reflects your Mac User directories.

I bet on your Docker Machine, you have a _/Users/Username/something_, in this case it is transparent.

Does that make sense to you ?

Plopix:
And for the permission issue, because I am using NFS, here is my exports:
```

bin/.docker/docker-compose.dev.yml
andrerom: What if it rather is `COMPOSE_APP_DIR`, default being `../../` ?

andrerom:
Cherry picked in 2e31a5f7c0f2fe61f3d234bd3714aa6d5339d443

PR #113, opened 2016-05-30
[Composer] Bump Symfony to 2.8

andrerom:
https://jira.ez.no/browse/EZP-25098

This will trigger a whole bunch of deprecation warnings for things removed in 3.0,
and if possible we will aim to take that gradually as soon as we merge this.

See: https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     },
     "require": {
         "php": "~5.5|~7.0",
-        "symfony/symfony": "~2.7.0",
+        "symfony/symfony": "~2.8.0",
         "twig/extensions": "~1.0",
         "symfony/assetic-bundle": "~2.3",
         "symfony/swiftmailer-bundle": "~2.3",


andrerom:
ping @bdunogier 

Some community ping @lolautruche @emodric  :)

emodric:
Well, I still have consistent issues with https://jira.ez.no/browse/EZP-25098 and I haven't been able to deduce why. I'm looking in general direction of legacy bridge as a culprit but not really sure yet.

bdunogier:
Unless I am mistaken, the issue here is related to early services right ? Does it happen out-of-the-box on 6.x to anyone ?

bdunogier:
@emodric we have @lolautruche's PR with the custom cache warmer. We should be good with 2.8 now as far as we can tell. Can you confirm ?

emodric:
I think so, judging by my latest tests :)

andrerom:
+1

bdunogier:
+1 as well. We have a stabilization window until the next release to find possible issues.

andrerom:
Ok, will merge this and see towards beta1 and rc1 if we find any further issues.

🚢

PR #114, opened 2016-05-30
[DOC] Removed extra '\' in INSTALL.md vhost generation command

miguelcleverti:
vhost generation command should only have one '\' to break the line.


andrerom:
+1

PR #115, opened 2016-06-03
EZP-25837: inaccessible/missing fonts on eZ Platform 2016.04

damianz5:
https://jira.ez.no/browse/EZP-25837

### DIFF
--- a/doc/apache2/Readme.md
+++ b/doc/apache2/Readme.md
@@ -78,7 +78,7 @@ Example config for Apache 2.4 in prefork mode:
         # Additional Assetic rules for environments different from dev,
         # remember to run php app/console assetic:dump --env=prod
         RewriteCond %{ENV:SYMFONY_ENV} !^(dev)
-        RewriteRule ^/(css|js|font)/.*\.(css|js|otf|eot|ttf|svg|woff) - [L]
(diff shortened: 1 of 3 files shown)


damianz5:
ping @andrerom @bdunogier

emodric:
Isn't just `s?` enough ? You don't need a capturing group :)

damianz5:
yes, but `fonts?` could be misleading

emodric:
Yes, could be :)

andrerom:
> yes, but fonts? could be misleading

how so? Folks thinking it would be a literal `?` in the url? well... :)

bdunogier:
I agree with @emodric, and was about to write the same remark. It's a regexp, it is nou meant to be a nice guy.

That aside +1.

damianz5:
right, PR updated

emodric:
@damianz5 Capturing group still left in `doc/nginx/ez_params.d/ez_prod_rewrite_params` :)

andrerom:
Thanks @damianz5 

Also merged to ezplatform-demo 1.3 and master branch: https://github.com/ezsystems/ezplatform-demo/commit/2eb2201129989b14a1f0c3f1a6e7e19d4b0ac802

PR #118, opened 2016-06-14
[WIP] EZP-25869: As a Developer I want to use Docker in development on Mac/Win

andrerom:
Issue: https://jira.ez.no/browse/EZP-25869
Status: **Work In Progress** *(atm not worked on, awaiting Docker Inc move, even with changes here performance is way to slow)*
 
Work on trying to improve performance within containers in dev to see if it can be made usable on Mac/Windows, hopefully this becomes obsolete if Docker Inc finds a solution to solve/hide the extremely slow IO when using Docker on Mac/Windows, see issue for more info.
 
Work on trying to improve performance within containers in dev.
- [x] Put cache in shm
- [x] Enable rewrite rules even if in dev to not load assets via Symfony
 - [ ] add this for Apache
- [x] Places profiling in /tmp/ as shm is limited to 64mb
- [ ] Investgate alternative solutions using NFS from guest, VMWare Fusion, * ?

### DIFF
--- a/doc/nginx/vhost.template
+++ b/doc/nginx/vhost.template
@@ -9,7 +9,7 @@ server {
     # Additional Assetic rules for eZ Publish 5.1 / 2013.4 and higher.
     ## Don't forget to run php app/console assetic:dump --env=prod
     ## and make sure to comment these out in DEV environment.
-    #if[SYMFONY_ENV!=dev] include ez_params.d/ez_prod_rewrite_params;
+    #if[SYMFONY_ASSET_REWRITE] include ez_params.d/ez_prod_rewrite_params;
 
     # Cluster/streamed files rewrite rules. Enable on cluster with DFS as a binary data handler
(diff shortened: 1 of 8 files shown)


doc/nginx/vhost.template
```diff
@@ -9,7 +9,7 @@ server {
     # Additional Assetic rules for eZ Publish 5.1 / 2013.4 and higher.
     ## Don't forget to run php app/console assetic:dump --env=prod
     ## and make sure to comment these out in DEV environment.
-    #if[SYMFONY_ENV!=dev] include ez_params.d/ez_prod_rewrite_params;
+    #if[SYMFONY_ASSET_REWRITE] include ez_params.d/ez_prod_rewrite_params;
```
andrerom: Note to self, this came from BD: if we can have a switch to be able to put backend in prod env while front is dev, that is a better solution for for those only developing on front.

andrerom:
Closing, this does not provide enough improvment to warrant the hacks.

PR #120, opened 2016-06-16
EZP-25974: Reduce parameters user is asked about on install

andrerom:
https://jira.ez.no/browse/EZP-25974

To simplify install, move parameters that aren't so common to setup on first install to default parameters file.

- [x] Handle ~~by doc or~~ by config to not wipe out old config settings on upgrade: https://github.com/Incenteev/ParameterHandler#keeping-outdated-parameters

### DIFF
--- a/app/config/default_parameters.yml
+++ b/app/config/default_parameters.yml
@@ -1,5 +1,19 @@
 # This file contains defaults for optional parameters, which you can override in parameters.yml
 parameters:
+    locale_fallback: en
+
+    # Setting for mail system, used by SwiftMailer
+    mailer_transport:  smtp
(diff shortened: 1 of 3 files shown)


bdunogier:
I'd go for the "keep outdated" option, at least for the time being. Deprecate the outdated parameters, and remove them in a major version.

andrerom:
Ok, review ping @bdunogier @lserwatka

lserwatka:
Looks good to me +1

lserwatka:
@andrerom do you aim this PR for 1.4.0 final?

andrerom:
yes :)

PR #121, opened 2016-06-17
EZP-25943: [Docker] Add redis config and way to inject it

andrerom:
Issue: https://jira.ez.no/browse/EZP-25943

As a step towards cluster setup, adding redis so we can test that.
Memcached can easily be added as well later when it arrives with PHP 7.0 support, assuming we bother *(Symfony 3.x only support Redis, and adding extensions to php image increases it's size)*.


andrerom:
ping @bdunogier

andrerom:
ping @bdunogier

PR #122, opened 2016-06-22
EZP-25915: Avoid RunTimeException on composer update/install

andrerom:
https://jira.ez.no/browse/EZP-25915

ref https://github.com/ezsystems/ezpublish-kernel/pull/1684


andrerom:
ping @bdunogier

crevillo:
+1.

PR #124, opened 2016-06-23
[Docker] Dump database into image for use directly by mysql container

andrerom:
Downside:
- [x] [Doc] Means prod also will need to run docker-compose -f install.yml step, however see below

Benefits:
- No need to manually run `ezplatform:install` anymore
- Possible to push image to docker hub and reuse on hosted services where you can't run manual steps like `ezplatform:install`


andrerom:
Ready for review and testing.

@bdunogier Ideally I'd want this in 1.4, as it is needed to be able to generate demo image, if so I can set that up before we tag 1.4.1, for in turn to finish documentation on how to install demo setup.

Dockerfile
```diff
@@ -11,8 +11,8 @@ ENV SYMFONY_ENV=prod
 # Copy in project files into work dir
 COPY . /var/www
 
-# Remove cache folders to avoid layer issues, ref: https://github.com/symfony/symfony/issues/12533#issuecomment-76406216
-RUN rm -Rf app/logs/* app/cache/* .git/ \
+# Remove cache folders to avoid layer issues, ref: https://github.com/docker/docker/issues/783
```
andrerom: @bdunogier I still havnt seen this on current code, but I'm pretty sure the issue linked to above is the one you experienced when you tried a few weeks ago.

TL;DR; Aufs refuses rm/chmod operations that potentially exposes previous restricted folders in lower layers, symptom is cryptic permission errors.

andrerom:
merged in https://github.com/ezsystems/ezplatform/commit/26eddf83ce1c437ae5456debb03d34e56a433e8f

PR #125, opened 2016-07-07
EZP - 26043 - PHP Garbage Collector symfony configuration logs out user of Platform UI

pedroresende:
#### EZP - 26043 - PHP Garbage Collector symfony configuration logs out user of Platform UI

Due to the fact that Symfony configuration has the gc_probability: 1 by default, instead of using the php.ini one of '0'.
This means that there is a 1% chance that the GC process will start on each request. 
Since Platform UI make hundreds of requests the 1% probability will occur very fast.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -71,6 +71,7 @@ framework:
     trusted_hosts:   ~
     trusted_proxies: ~
     session:
+        gc_probability: 0
         # handler_id set to null will use default session handler from php.ini
         handler_id:  ~
         # Note: eZ Publish also allows session name and session cookie configuration to be per SiteAccess, by


andrerom:
But gc is afaik supposed to only remove in-active sessions, so if there is a issue, it most likly need a different fix then this.

joaoinacio:
The problem could be related to lack of permissions of www user to the php sessions folder (which is common).
When attempting to run GC it fails with an exception, and this apparently causes the session to be lost (?!)

joaoinacio:
-1 on on the change as is.

Reason: debian has it's own cleanup / gc scripts, other OS's don't so GC should be done by PHP.

The problem only occurs in debian due to permissions, whereas in RHEL for example the session folder group is apache.

More importantly, the issue is caused by a notice triggering an HTTP 500 in dev/debug mode, but not in PROD env.

IMHO this issue/config change could probably be documented somehow (installation notes?).

pedroresende:
What is the disadvantage ? Since in php.ini by default the gc_probability is set to 0 ? The problem is that Symfony has the default value set to 1 and this causes problems

joaoinacio:
> What is the disadvantage ? Since in php.ini by default the gc_probability is set to 0 ? The problem is that Symfony has the default value set to 1 and this causes problems

In debian it is 0, but in RHEL for example the php `gc_probability` is set to 1 (because there are no gc scripts)

At most, this setting override should be removed from symfony instead...

PR #128, opened 2016-07-18
EZP-24854 EZP-26087 Update varnish configuration to purge required lo…

benoitvidis:
…cations only

cf https://jira.ez.no/browse/EZP-26087

The current implementation of varnish ban allows only one Location ID to be set in the X-Location-Id Header.

The proposal is to:

1. update default varnish configuration (this PR)
2. revert https://github.com/ezsystems/ezpublish-kernel/commit/12353b48909873ca5d71d6c11f7a896c70f3efda

### DIFF
--- a/doc/varnish/vcl/varnish3.vcl
+++ b/doc/varnish/vcl/varnish3.vcl
@@ -116,7 +116,7 @@ sub ez_purge {
         }
 
         if (req.http.X-Location-Id) {
-            ban( "obj.http.X-Location-Id ~ " + req.http.X-Location-Id);
+            ban( "obj.http.X-Location-Id ~ \b" + req.http.X-Location-Id + "\b");
             if (client.ip ~ debuggers) {
                 set req.http.X-Debug = "Ban done for content connected to LocationId " + req.http.X-Location-Id;
             }
--- a/doc/varnish/vcl/varnish4.vcl
+++ b/doc/varnish/vcl/varnish4.vcl
@@ -113,7 +113,7 @@ sub ez_purge {
         }
 
         if (req.http.X-Location-Id) {
-            ban("obj.http.X-Location-Id ~ " + req.http.X-Location-Id);
+            ban("obj.http.X-Location-Id ~ \b" + req.http.X-Location-Id + "\b");
             if (client.ip ~ debuggers) {
                 set req.http.X-Debug = "Ban done for content connected to LocationId " + req.http.X-Location-Id;
             }


andrerom:
ping @joaoinacio @bdunogier

bdunogier:
So this will work because the comma is considered as a word boundary, making sure every location id is interpreted individually ?

It looks okay to me, but I'd like QA to a) verify that this is indeed failing b) test the patch as well.

andrerom:
> It looks okay to me, but I'd like QA to a) verify that this is indeed failing b) test the patch.

They will also need to revert @joaoinacio's original kernel patch to test this and test the original issue he solved there.

bdunogier:
Good point @andrerom. I'll open a PR with the revert.

gggeek:
To be backported to 5.4.9 as well :-)

joaoinacio:
untested but looks OK afaict, as @bdunogier said should probably go through QA

dspe:
looks good to me aswell :) +1

bdunogier:
Ping @andrerom can you approve unless you disapprove ?

miguelcleverti:
PR approved by QA

gggeek:
Sorry to chime in so late guys, but I am not 100% sure that putting the \b modifier here might not be the best solution - as it kind of introduces a hard coupling between the VCL and the php code in the purge client (https://github.com/ezsystems/ezpublish-kernel/pull/1806/files).

What if we put the \b modifier directly in the client?
- more flexible vcl
- less BC breaks
- same results

andrerom:
_(note, this is not needed if #143 goes in, as it replaces `X-Location-Id` with `xkey`)_

miguelcleverti:
@benoitvidis I posted a comment in [EZP-26087](https://jira.ez.no/browse/EZP-26087) with the tests results of the PR.

I but in short the PR does not seem to solve the issue.

After making a BAN request for multiple X-Location-Id's there is still cache for the controller with multiple X-Location-Id, however for controllers with only one X-Location-Id the BAN request works has intended. 

Further more if the BAN request is made using the syntax:
`X-Location-Id: (978687|123|7987|7237) ` Instead of: `X-Location-Id: ^(978687|123|7987|7237)$`

It seem to have the expected behavior.

yannickroger:
Could any one help our QA to validate this fix? Or else we will have to close this PR as they can't fix the problem using it.

andrerom:
Closing this as we in the meantime have moved to xkey, if someone has solution that won't break multiple location ids, please share on JIRA issue.

PR #134, opened 2016-08-23
Allow sensio/distribution-bundle 3.x, 4.x or 5.x to allow legacy bridge install

emodric:
Followup to : https://github.com/ezsystems/ezplatform/commit/dfe122b1f386e5d9b70dc6b475038ab0c16a2231#commitcomment-18733964

Legacy bridge requires `sensio/distribution-bundle` 3.x or 4.x and will not be able to install with `^5.0` as a version requirement.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "symfony/assetic-bundle": "~2.3",
         "symfony/swiftmailer-bundle": "~2.3",
         "symfony/monolog-bundle": "~2.4",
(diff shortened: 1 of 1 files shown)


bdunogier2:
+1, I don't see a problem with it. @andrerom ?

andrerom:
+1

emodric:
Thanks guys! 👍

PR #135, opened 2016-09-12
[Insight] YAML files should not contain syntax error

SylvainGuittard:
jira issue: https://jira.ez.no/browse/EZP-26286

Tested on SensioLabsInsight.
@andrerom could you please test it with your install process?


yannickroger:
Could we put all this command line stuff into a dedicated sh file ?

andrerom:
> Could we put all this command line stuff into a dedicated sh file ?

maybe in the future, but this will change with how the docker stuff is setup, and in this way no one reuses it and expects it to work a certain way.

andrerom:
> @andrerom could you please test it with your install process?

I don't need to test this, Travis does that.

PR #136, opened 2016-09-12
[Insight] Symfony secret should be changed

SylvainGuittard:
jira issue: https://jira.ez.no/browse/EZP-26287

Form SensioLabsInsight:

> Symfony secret should be changed
> The secret is the default one from the Symfony Standard Edition

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -1,7 +1,7 @@
 # This file is a "template" of what your parameters.yml file should look like
 parameters:
     # A secret key that's used to generate certain security-related tokens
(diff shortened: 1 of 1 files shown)


andrerom:
Hmm, as end user is the one that needs to change it I'm not sure I agree with this one. But of course we could change to any other kind of sentence that says user should change it asap.

SylvainGuittard:
Updated ;)

andrerom:
+1

yannickroger:
+1
(I thought you would propose something with a joke as a token @SylvainGuittard  :smile: )

PR #138, opened 2016-10-03
[Docker] Switch to slightly updated format for php image with entrypoint

andrerom:
Switch to new [v1 image](https://github.com/ezsystems/docker-php/pull/15) of docker-php which has [entrypoint](https://github.com/ezsystems/docker-php/blob/master/php/scripts/docker-entrypoint.sh#L50-L58) which makes it possible to run custom logic in it on startup without having to extend it. It also contains support for adapting adapting behat.yml so we don't need to hardcode that as container command here.

### DIFF
--- a/.env
+++ b/.env
@@ -12,8 +12,8 @@ DATABASE_PASSWORD=SetYourOwnPassword
 DATABASE_NAME=ezp
 
 ## Docker images (name and version)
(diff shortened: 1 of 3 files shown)


andrerom:
Thanks, small step :)

PR #141, opened 2016-10-18
Pin selenium docker image version to 2.53.1

miguelcleverti:
Selenium 3 causes issue with behat because it uses the new marionette/geckodriver(see https://github.com/mozilla/geckodriver) which are not yet fully implemented. For now the version should be pinned to the previous one

### DIFF
--- a/.env
+++ b/.env
@@ -16,7 +16,7 @@ PHP_IMAGE=ezsystems/php:7.0-v1
 PHP_IMAGE_DEV=ezsystems/php:7.0-v1-dev
 NGINX_IMAGE=nginx:stable
 MYSQL_IMAGE=mariadb:10.0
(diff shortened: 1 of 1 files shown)


miguelcleverti:
ping @andrerom @micszo

andrerom:
In retrospective this should have been done against 1.5 branch

lserwatka:
@andrerom we should revert this and open against 1.5 then merge up to master and then to 1.5/master ezstudio and 1.5/master ezstudio-demo. WDT?

andrerom:
to late, I cherry picked it to 1.5 and merged that into master, I suggest you do the same, but in your case you can also merge up changes from master here _(optionally you can do that later also, as it's only a merge commit with effectively no changes given the change was already there)_.

PR #142, opened 2016-10-28
[Docker] Add support for Varnish setup with Docker

andrerom:
Todo:
- [x] Get purging to work
- [ ] Setup on some of the behat tests _(done, but now getting segmentation fault on the VM which might or might not have appeared before env for pointing behat to varnish was changed)_

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,7 +3,8 @@
 
 ### GIT files
 .git
-vendor/*/*/.git
+# Skipped as it makes build very slow and since it is needed by composer when working on dev branches
+#vendor/*/*/.git
 
 ### Symfony template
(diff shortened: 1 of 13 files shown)


andrerom:
```
Warning: XMLWriter::text() expects parameter 1 to be string, array given in vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Output/Generator/Xml.php line 171
```
@bdunogier Seems there is a regression on xml  REST BDD tests now, from what I can see on travis it has happened between 4 and 11 days ago, and is on a test testing Query / Criterion stuff.

andrerom:
closed in favour of #143.

PR #143, opened 2016-11-12
[WIP][Docker] Add support for using Varnish with xkey

andrerom:
Varnish changes, ezplatform-http-cache package _(uses varnsh xkey for improved varnish performance, and cache ability of our REST api++)_.

Todo:
- [ ] Rebase on top of  #188 once that is merged
- [ ] Change to target 2.0 branch where `ezplatform-http-cache` is the default


andrerom:
Back to passing, cleaned up and with last push it works locally as well again, super fast as always. /cc @bdunogier 


How to tests _(same exact steps as .travis.yml so you know where to look next step you want to try)_:
1. check out branch
2. `export TEST_CMD="bin/behat -vv --profile=rest --suite=fullJson --tags=~@broken" COMPOSE_FILE="doc/docker-compose/base-prod.yml:doc/docker-compose/redis.yml:doc/docker-compose/varnish.yml:doc/docker-compose/selenium.yml" BEHAT_WEB_HOST="varnish"`
3. `docker-compose -f doc/docker-compose/install.yml up --abort-on-container-exit`
4. `docker-compose up -d`
5. Depending on your docker setup you should now be able to browse the install on `localhost` or some ip, where port 8080 is nginx, and 8081 is varish 
6. if you don't change anything you can also run the behat tests for rest with `docker-compose exec --user www-data app sh -c "php /scripts/wait_for_db.php; php $TEST_CMD"` _(one will fail because we registered redis here)_

bdunogier:
For readability, I'd suggest we split the ci/docker changes from the VCL ones (more like doc).

I can do it, @andrerom, just let me know.

bdunogier:
From what I've seen, the VCL looks good. Haven't really looked at the docker part :)

andrerom:
> From what I've seen, the VCL looks good. Haven't really looked at the docker part :)

Should be easier now, just removed the duplicated vcl in entrypoint folder :D

doc/varnish/vcl/varnish4.vcl
```diff
@@ -120,10 +141,28 @@ sub ez_purge {
             return (synth(200, "Banned"));
         }
     }
+
+    if (req.method == "PURGE") {
+        if (!client.ip ~ invalidators) {
+            return (synth(405, "Method not allowed"));
+        }
+
+        # If http header "xkey" is set, we assume purge is on key and you have Varnish xkey installed
```
andrerom: Align header name to be same on occurances here, in kernel and on foshttpcache pr, afaik it should be "key" like on line below and in foshttpcache

doc/docker-compose/varnish.yml
```diff
@@ -0,0 +1,41 @@
...
+## DEBUG??
+# In need of debugging all request going to Varnish, use varnishlog, example:
+# docker-compose exec varnish varnishlog -c -i ReqURL,ReqMethod -I ReqHeader:X-Location-Id
+# Or more relevant only BAN's with all info:
+# docker-compose exec varnish varnishlog -g request -q "ReqMethod eq 'BAN'"
+# Or rather PURGE which is used now with xkey:
+# docker-compose exec varnish varnishlog -g request -q "ReqMethod eq 'PURGE'"
+#
+# But before doing that check that http and not local purge client is set:
+# docker-compose exec app app/console --env=dev debug:container ezpublish.http_cache.purge_client
```
bdunogier: Needs to be adapted to the service used by the http-cache package.

composer.json
```diff
@@ -27,7 +27,7 @@
         "sensio/generator-bundle": "~2.3",
         "incenteev/composer-parameter-handler": "~2.0",
         "tedivm/stash-bundle": "~0.4",
-        "ezsystems/ezpublish-kernel": "~6.7@dev",
+        "ezsystems/ezpublish-kernel": "dev-feature-EZP-25696 as 6.7.x-dev",
```
bdunogier: To be replaced with the enablement of the `EzSystemsPlatformHttpCacheBundle` in the `AppKernel`.

doc/docker-compose/varnish.yml
andrerom: What does it need to be?

bdunogier:
Given that the feature has been merged to ezsystems/ezplatform-http-cache, it has been decided that the VCL would go into the new package. I'll push a new version without the VCL changes. The Docker ones only make sense in this repository, don't they ?

I suggest that we leave the current varnish configuration untouched in this repository, and update the documentation to mention the external package and corresponding documentation.

andrerom:
> I suggest that we leave the current varnish configuration untouched in this repository, and update the documentation to mention the external package and corresponding documentation.


@bdunogier For 1.8 yes

But I think I'll leave this PR open with aim to at some point have official Varnish container, and ideally it should use the xkey variant of the VCL, so guess it will have to wait for version where we switch the default.

andrerom:
Closing, solved in:
- 1.x (added xkey moduke to docker image): https://github.com/ezsystems/ezplatform/commit/2c946db7db9351e986e4377e3461eba93ef36504
- 2.x (change to use ezplatform-http-cache and it's updated VCL): https://github.com/ezsystems/ezplatform/commit/a83894c3d0e19d82734dda93c21aeba73698d9cc

PR #144, opened 2016-11-26
EZP-26682: Add Content on the Fly Bundle in clean install

andrerom:
> Issue: https://jira.ez.no/browse/EZP-26682

To be able to more strongly advice people to use clean for install and upgrade* make sure it has same level of features, given the rewrite to integrate the feature is currently on hold.

This applies to ezstudio as well, and once this is merged it can be naturally merged down there.

*_Doc on Studio says people can use demo for upgrade but that causes issue whenever demo add new demo specific features and people merge that in on upgrade. Note: On studio templates and config for landing pages is missing, so this is most likely the root cause for people using demo on upgrade._

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -46,6 +46,7 @@ class AppKernel extends Kernel
             new EzSystems\PlatformInstallerBundle\EzSystemsPlatformInstallerBundle(),
             new EzSystems\RepositoryFormsBundle\EzSystemsRepositoryFormsBundle(),
             new EzSystems\EzPlatformSolrSearchEngineBundle\EzSystemsEzPlatformSolrSearchEngineBundle(),
+            new EzSystems\EzContentOnTheFlyBundle\EzSystemsEzContentOnTheFlyBundle(),
             new Bazinga\Bundle\JsTranslationBundle\BazingaJsTranslationBundle(),
             new AppBundle\AppBundle(),
         );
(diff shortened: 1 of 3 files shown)


andrerom:
review ping @bdunogier @nicolas-bastien /cc @lserwatka

bdunogier:
> Doc says people can use demo for upgrade

This needs to go. Demo can't be used for anything except demo :)

bdunogier:
@dpobel any objection ? :)

bdunogier:
Sent the story to QA for testing.

mere-teresa:
Please @andrerom could you provide me the source of this info. I would like to correct the doc.

> *Doc says people can use demo for upgrade but that causes issue whenever demo add new demo specific features and people merge that in on upgrade.

andrerom:
@mere-teresa https://doc.ez.no/display/DEVELOPER/Updating+eZ+Enterprise  _(so it's the doc for studio, aka the platform flavor our customers use, and it states that you can choice yourself, but demo is not really suitable for upgrades unless we then explain all the possible issues you can get by doing that :) )_

andrerom:
Seems QA approved, I reported only issue I know about _(ballon showing "undefined" when content type description is empty)_: https://jira.ez.no/browse/EZS-1160  /cc @lserwatka

lserwatka:
@andrerom we will look into it later this week. Should be easy to fix.

andrerom:
@lserwatka you can now merge up to studio, I'll take care about ezplatform-demo as the demos will likely give some merge conflicts now until resolved.

PR #146, opened 2016-11-30
[composer] Update ezsystems/content-on-the-fly-prototype to version 0.1.7

lserwatka:
ping @andrerom


andrerom:
+1, however be aware it is just minimum version, don't need to bump it for every patch, using latests versions are assumed anyway.

PR #150, opened 2016-12-20
Fix EZP-26748: Wrong image URI when content name has extended UTF characters

glye:
> Fixes https://jira.ez.no/browse/EZP-26748
> Test in https://github.com/ezsystems/ezpublish-kernel-ee/pull/145
> Taken from https://github.com/ezsystems/ezpublish-platform/pull/20 where the tests don't run
> Status: Ready to merge

Fixes the problem by ensuring UTF-8 is used, though this probably isn't the ideal way to do it. Anyway, let's see how the full tests go.

### DIFF
--- a/app/console
+++ b/app/console
@@ -7,6 +7,9 @@
 
 set_time_limit(0);
 
+// Ensure UTF-8 is used in string operations
+setlocale(LC_CTYPE, 'C.UTF-8');
+
 // Use autoload over boostrap here so we don't need to keep the generated files in git
 require_once __DIR__ . '/autoload.php';
 require_once __DIR__ . '/AppKernel.php';
--- a/web/app.php
+++ b/web/app.php
@@ -3,6 +3,9 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Debug\Debug;
 
+// Ensure UTF-8 is used in string operations
+setlocale(LC_CTYPE, 'C.UTF-8');
+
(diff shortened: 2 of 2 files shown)


glye:
Review ping @andrerom @bdunogier - this is a (perhaps too) simple solution. I have hardcoded it because AFAIK ezp is utf-8 only.

bdunogier:
I'm not a charset expert, but based on what I know it looks okay.

Do we need to add the same thing to our tests bootstrap files, to be consistent ?

glye:
> Do we need to add the same thing to our tests bootstrap files, to be consistent?

@bdunogier Sounds good. Could you give some pointers? (Travis/Behat config?)

bdunogier:
For phpunit, it would probably be in `bootstrap.php`. Behat will be taken care of by the change you've done here, as it uses `app.php`.

glye:
Ok, that would be bootstrap.php in kernel, repoforms, pui and other repos, then?

andrerom:
Bootstrap is one thing, we also more importantly need to add this in `app/console`, right?

glye:
Update: Added same in `app/console`, thanks.
Wondering to what degree console commands are tested here. Locally at least cache clear seems fine with the change. (In most cases it doesn't matter, anyway.)

yannickroger:
Don't forget to send it to QA 😄

glye:
Sent to QA.

glye:
Merged in 1.7: https://github.com/ezsystems/ezplatform/commit/45aa9a67015f33956501a33fe5517fee1b2b58ca

PR #152, opened 2017-01-09
EZS-1281: Adjust setup_from_external_repo.sh

damianz5:
Adjust setup_from_external_repo.sh in ezplatform meta

https://jira.ez.no/browse/EZS-1281

### DIFF
--- a/bin/.travis/trusty/setup_from_external_repo.sh
+++ b/bin/.travis/trusty/setup_from_external_repo.sh
@@ -37,13 +37,14 @@ ls -al .
 ./bin/.travis/trusty/update_docker.sh
 
 echo "> Modify composer.json to point to local checkout"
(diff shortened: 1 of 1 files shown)


lserwatka:
Once this is merge I will pull this to other repos.

micszo:
@andrerom Could you confirm the list of occurrences listed in the Jira ticket? Do you know of any other cases that should be taken into account?

andrerom:
Somewhat, but not sure without checking.

But script seems to be safe, it falls back to how it already behaves so don't see any issues with this now.

micszo:
@andrerom I confirmed with @damianz5 this is ready, please merge. :)

andrerom:
Thanks, merged and merged up to master.
So it can be merged further to studio repos now.

PR #153, opened 2017-01-16
EZP-26834: Synchronize translations in ezplatform-i18n

nicolas-bastien:
Jira : https://jira.ez.no/browse/EZP-26834

# Description

This PR adds a script to ease translation synchronisation to ezplatform-i18n

### DIFF
--- a/bin/synchronize-translations.sh
+++ b/bin/synchronize-translations.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env sh
+# This script should be used to synchronize ezplatform-i18n repository
+# Translations must be up-to-date in all packages
+echo 'Translation synchronization';
+
+echo '# Clean corresponding ezplatform-i18n folder';
+rm ./vendor/ezsystems/ezplatform-i18n/platform-ui-bundle/*.xlf
+rm ./vendor/ezsystems/ezplatform-i18n/ezpublish-kernel/*.xlf
+rm ./vendor/ezsystems/ezplatform-i18n/repository-forms/*.xlf
+
+echo '# Mirror the translation files';
+cp ./vendor/ezsystems/platform-ui-bundle/Resources/translations/*.xlf ./vendor/ezsystems/ezplatform-i18n/platform-ui-bundle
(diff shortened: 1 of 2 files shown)


bin/synchronize-translations.sh
```diff
@@ -0,0 +1,21 @@
...
+# This script should be used to synchronize ezplatform-i18n repository
+# Translations should be extracted in each bundles before this step
+echo 'Translation synchronization';
+
+echo '# Clean corresponding ezplatform-i18n folder';
+rm ./vendor/ezsystems/ezplatform-i18n/platform-ui-bundle/*.xlf
+rm ./vendor/ezsystems/ezplatform-i18n/ezpublish-kernel/*.xlf
+rm ./vendor/ezsystems/ezplatform-i18n/repository-forms/*.xlf
+
+echo '# Miror the translation files';
```
bdunogier: "Mirror"

composer.json
```diff
@@ -48,7 +48,8 @@
         "behat/mink-goutte-driver": "*",
         "behat/mink-selenium2-driver": "*",
         "jarnaiz/behat-junit-formatter": "^1.3",
-        "ezsystems/behatbundle": "^6.3"
+        "ezsystems/behatbundle": "^6.3",
+        "ezsystems/ezplatform-i18n": "^1.0@dev"
```
bdunogier: Could you move the new requirement to another line ? We did say that we would try to avoid generating conflicts for users, and the last line of `require` or `require-dev` is a good place to start from :)

I'd move it either in first position, or around the top.

bin/synchronize-translations.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env sh
+# This script should be used to synchronize ezplatform-i18n repository
+# Translations should be extracted in each bundles before this step
```
bdunogier: I'd replace this with "Translations must be up-to-date in all packages".

Since translations are supposed to be update in pull-requests, they should always be up-to-date ;)

PR #155, opened 2017-01-20
EZP-26869: Documented in-context translation usage

bdunogier:
Title says it all.

### DIFF
--- a/doc/i18n/in_context_translation.md
+++ b/doc/i18n/in_context_translation.md
@@ -1,0 +1,46 @@
+### In-context UI translation
+
+Since eZ Platform 1.7.0, the interface has been fully translatable. Version 1.8.0 introduces
+official support for [crowdin.com](crowdin.com) as a translation management system. In
+addition, it integrates support for [in-context translation](in-context), a feature that
+allows you to translate strings from the interface, _in context_.
+
+![In-context translation of Platform UI](https://cloud.githubusercontent.com/assets/235928/21649816/44fc2ea0-d2a3-11e6-8c0e-1b5493ea47e9.png)
+
+## Toggling in-context translation
+To start translating, you need to set a browser cookie. There are several ways to do this,
+but we will highlight a couple here.
+
+### Using the debugging console
+
+One way is to open the development console and run these lines:
+- enable: `document.cookie='ez_in_context_translation=1;path=/;'; location.reload();`
+- disable: `document.cookie='ez_in_context_translation=;expires=Mon, 05 Jul 2000 00:00:00 GMT;path=/;'; location.reload();`
+
+### Using bookmarks
+You can easily create two bookmarks to toggle in-context on/off.
+
+Right-click your browser's bookmark bar, and create a new one, with the following label and link: 
+- Enable in-context: `javascript:(function() {document.cookie='ez_in_context_translation=1;path=/;'; location.reload();})()`
+- Disable in-context: `javascript:(function() {document.cookie='ez_in_context_translation=;expires=Mon, 05 Jul 2000 00:00:00 GMT;path=/;'; location.reload();})()`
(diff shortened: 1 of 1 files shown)


doc/i18n/in_context_translation.md
```diff
@@ -0,0 +1,46 @@
+### In-context UI translation
+
+Since eZ Platform 1.7.0, the interface has been fully translatable. Version 1.8.0 introduces
+official support for [crowdin.com](crowdin.com) as a translation management system. In
+addition, it integrates support for [in-context translation](in-context), a feature that
+allows you to translate strings from the interface, _in context_.
+
+![In-context translation of Platform UI](https://cloud.githubusercontent.com/assets/235928/21649816/44fc2ea0-d2a3-11e6-8c0e-1b5493ea47e9.png)
```
yannickroger: Should we store this screenshot along with the documentation?

doc/i18n/in_context_translation.md
```diff
@@ -0,0 +1,46 @@
...
+The first time you enable in-context, if you're not logged into Crowdin, it will ask you
+to login or register an account. Once done, it will ask you which language you want to
+translate in, from the list of configured languages in crowdin.
+
+Choose your language, and you can start translating right away. Translatable strings in the
+interface will be outlined in red (untranslated), blue (translated) or green (approved).
+When moving over them, an edit button will show up on the top left corner of the outline.
+Click on it, and edit the string in the window that shows up.
+
+## Troubleshooting
```
yannickroger: You should add another paragraph explaining what is going to happen next. Like when the translations will actually be integrated in the product.

doc/i18n/in_context_translation.md
```diff
@@ -0,0 +1,46 @@
+### In-context UI translation
+
+Since eZ Platform 1.7.0, the interface has been fully translatable. Version 1.8.0 introduces
+official support for [crowdin.com](crowdin.com) as a translation management system. In
+addition, it integrates support for [in-context translation](in-context), a feature that
+allows you to translate strings from the interface, _in context_.
+
+![In-context translation of Platform UI](https://cloud.githubusercontent.com/assets/235928/21649816/44fc2ea0-d2a3-11e6-8c0e-1b5493ea47e9.png)
+
+## Toggling in-context translation
```
yannickroger: How does it work regarding the version used by the user. I mean 1.8.x or master etc.

doc/i18n/in_context_translation.md
bdunogier: Maybe, maybe not. @mere-teresa should confirm that, but storing images is one of the open questions when it comes to doc & github. Based on what I have seen, using images stored via pull-requests / issues is a viable way to do it, and it is simple. We can change it at some point indeed.

doc/i18n/in_context_translation.md
bdunogier: Will work starting from the hash where this was merged to ezplatform. So from master of yesterday and 1.8.0.

doc/i18n/in_context_translation.md
bdunogier: I'm not yet sure about that :)

SylvainGuittard: This documentation page is not related to the translation in the project. It's only "how to use in-context-translation". I would not add the "next steps".

doc/i18n/in_context_translation.md
yannickroger: We should at some point as it kills the value of being able to read the documentation offline.

doc/i18n/in_context_translation.md
yannickroger: So the fact that the doc is there means the version is able to handle that :)

PR #156, opened 2017-01-22
sensio/generator-bundle update

xrow:
Allow newer versions of sensio/generator-bundle.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "symfony/swiftmailer-bundle": "~2.3",
         "symfony/monolog-bundle": "~2.4",
         "sensio/distribution-bundle": "^3.0.36|^4.0.6|^5.0",
-        "sensio/generator-bundle": "~2.3 | ~ 3.0",
+        "sensio/generator-bundle": "*",
         "incenteev/composer-parameter-handler": "~2.0",
         "tedivm/stash-bundle": "~0.4",
         "ezsystems/ezpublish-kernel": "~6.7@dev",
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "symfony/swiftmailer-bundle": "~2.3",
         "symfony/monolog-bundle": "~2.4",
         "sensio/distribution-bundle": "^3.0.36|^4.0.6|^5.0",
-        "sensio/generator-bundle": "~2.3",
+        "sensio/generator-bundle": "~2.3 | ~ 3.0",
         "incenteev/composer-parameter-handler": "~2.0",
         "tedivm/stash-bundle": "~0.4",
         "ezsystems/ezpublish-kernel": "~6.7@dev",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -24,7 +24,7 @@
         "symfony/swiftmailer-bundle": "~2.3",
         "symfony/monolog-bundle": "~2.4",
         "sensio/distribution-bundle": "^3.0.36|^4.0.6|^5.0",
-        "sensio/generator-bundle": "~2.3",
+        "sensio/generator-bundle": "*",
```
andrerom: We would typically rather bump it to "~3.0"

emodric: I don't think it is used directly by eZ Platform, but rather it's a part of Symfony Standard Edition. In any case, I'd rather go for `^2.3|^3.0` than unbound constraint.

xrow: I checked you don`t use it at all, but of course you want to to start a new bundle.

xrow:
Can one rerun CI it looks like it is not my fault. Don`t know how to trigger that.

composer.json
xrow: One last comment. @emodric I wouldn`t define a constraint, if there isn`t a real one.

emodric: @xrow What does that mean? It's a bad practice to use unbound constraints, as stuff can break at any moment with them.

xrow: @emodric 

If you are reffering to this https://getcomposer.org/doc/faqs/why-are-unbound-version-constraints-a-bad-idea.md. Form my point of view is not applicable since there is nothing that can brake, since ezplatform doesn`t use it. We just want the feature. The alternative is not to have he dependency at all.

andrerom:
> Can one rerun CI it looks like it is not my fault. Don`t know how to trigger that.

It will be rerun once PR is amended or new commit is added to fix the inline comment ;)

composer.json
emodric: Well, since this is not a kernel requirement, you can change it yourself in your projects, but by default I don't think it's a good idea.

xrow: @emodric I can`t change it in my projects because I need to use version 3 and eZ wants version 2.

emodric: @xrow I'm not sure I understand you. eZ kernel does not have a dependency on `sensio/generator-bundle`, therefore, you can change it yourself in `composer.json` in your own projects.

xrow: But I have my requirement on ezplatform and not on ezpublish-kernel since I want to include all of eZ.

emodric: Uhm... But that's not how it works :)

You should not require `ezystems/ezplatform` in your project, your project should be a fork of `ezsystems/ezplatform`.

xrow: Maintaining a fork of ezsystems/ezplatform is to complicated for us. Having the dependency is way easier. Please lets concentrate on solving this now.

bdunogier: No problem with the dependency update, but please let's use `^2.3|^3.0`. I'd rather avoid the warning on `composer validate` :)

andrerom:
Thanks @xrow and @emodric :)

PR #166, opened 2017-02-15
Docker misc improvments

andrerom:
Smaller improvements to simplify use and clarify doc a bit:
- Don't specify hard coded default of SYMFONY_DEBUG in .env, make it optional so you don't need to specify it for dev mode
  - _I had missed the fact that it will not be set in container if not set in parent env, I thought based on miss reading doc that it would become empty string causing issues_
- Add possibility to inject SYMFONY_HTTP_CACHE env variables, to control http cache
- Update doc in regards to redis and solr status


andrerom:
travis is green, ping @bdunogier

PR #168, opened 2017-02-26
EZP-27047: Composer update uses extra memory on wide version constraints

andrerom:
> Issue: https://jira.ez.no/browse/EZP-27047

Reduces memory on composer update:
- PHP 5.6: From 1.5GB and 72sec, to 1.3GB and 67sec
- PHP 7.1: From 1.1GB and 32sec, to 950MB and 22sec

##### Related _(but different)_ issue:
- [EZS-1134](https://jira.ez.no/browse/EZP-27047) On composer install, Composer buffers packages in memory and consumes a lot of memory on packages with lots of binary content
- [6168](https://github.com/composer/composer/pull/6168) `Improve memory usage resolving dependencies` _a PR recently merged for next version of Composer_

Todo:
- [x] See if just specifying minor versions gives most of the benefit to avoid wrongly implying we need to keep this up to date on every patch version of third party packages.


composer.json
```diff
@@ -18,40 +18,39 @@
...
+        "symfony/monolog-bundle": "^2.12.1",
+        "sensio/distribution-bundle": "^3.0.36|^4.0.6|^5.0.18",
+        "sensio/generator-bundle": "^2.5.3|^3.1.2",
+        "incenteev/composer-parameter-handler": "^2.1.2",
+        "tedivm/stash-bundle": "~0.6.1",
+        "ezsystems/ezpublish-kernel": "~6.9@dev",
+        "ezsystems/repository-forms": "^1.6.0",
+        "ezsystems/ezplatform-solr-search-engine": "^1.2.1@dev",
+        "ezsystems/platform-ui-bundle": "^1.9.0@dev",
+        "ezsystems/ez-support-tools": "^0.1.2",
```
andrerom: Unsure if it is ok they we allow all 0.x @bdunogier ?

composer.json
```diff
@@ -18,40 +18,39 @@
...
-        "symfony/expression-language": "~2.4",
-        "sensio/framework-extra-bundle": "~3.0",
+        "egulias/listeners-debug-command-bundle": "^1.9.1",
+        "white-october/pagerfanta-bundle": "^1.0.8",
+        "hautelook/templated-uri-bundle": "^2.0.1",
+        "doctrine/doctrine-bundle": "^1.6.7",
+        "sensio/framework-extra-bundle": "^3.0.22",
         "ezsystems/content-on-the-fly-prototype": "~0.1.7",
-        "willdurand/js-translation-bundle": "^2.7@dev"
+        "willdurand/js-translation-bundle": "^2.6.4"
```
andrerom: As far as I could see no extra commits in master over 2.6.4 tag.

andrerom:
review ping @glye @yannickroger

yannickroger:
@glye That's also what I thought when I first had a look at the PR a couple days ago.

andrerom:
yes, once in a while check our requriments and keeping them up to date basically.

composer plugin for phpstorm is to much help here, it shows you inline what version you have installed, so just put that in and open PR to see if travis uncovers conflict with something on a given php version for instance. Or do it as part of version bumps.

But it's not like we need to do this all the time, and I don't think we need to actually require specific patch versions like I did here.

I'll test how it looks like just specifying minor versions _(see now todo)_.

bdunogier:
Why not !

Isn't it what https://www.versioneye.com/ is tracking ?

andrerom:
True, maybe there is a lib doing that we can do on Jenkins?

andrerom:
Or service, I think one of them are free for open source and send PR with changes on updates.

andrerom:
With last commit: ` Memory usage: 201.31MB (peak: 1057.91MB), time: 22.47s`

andrerom:
Turns out just forcing latest symfony version does wonders to the memory usage.
I'm not saying we need to do this all the time, but this fits with bumping it on security updates.

`Memory usage: 205.71MB (peak: 963.18MB)`_ (different machine so I'm omitting the time, for comparison this machine used 1067MB where the other one used 1057MB above)_.

andrerom:
> Could we try to provide this as an opt-in ? Like a composer.json-OPTIMIZED. A custom composer.json can be used by means of environment variable. It should make it easy for developers & bots to switch to the other version for efficiency.

no, then I rather do it as v2 as I originally planned when I mentioned it to you.

andrerom:
Updated numbers with Composer `v1.4.0`:
- master: Memory usage: 208.22MB (peak: 598.21MB), time: 37.63s
- this branch: Memory usage: 208.16MB (peak: 547.07MB), time: 32.66s


So: Update composer guys ;)

PR #170, opened 2017-04-04
EZP-27191: Add ezsystems/ezplatform-multi-file-upload package into the ezplatform meta repository

lserwatka:
https://jira.ez.no/browse/EZP-27191


app/config/routing.yml
```diff
@@ -30,3 +30,7 @@ _ezplatformRepositoryFormsRoutes:
 _contentOnTheFly:
     resource: "@ContentOnTheFlyBundle/Resources/config/routing.yml"
     prefix:   "%ezpublish_rest.path_prefix%"
+
+_multiFileUpload:
+    resource: "@EzSystemsMultiFileUploadBundle/Resources/config/routing.yml"
+    prefix:   "%ezpublish_rest.path_prefix%"
```
andrerom: nitpick: new line

composer.json
```diff
@@ -40,7 +40,8 @@
         "doctrine/doctrine-bundle": "~1.5",
         "sensio/framework-extra-bundle": "~3.0",
         "ezsystems/content-on-the-fly-prototype": "~0.1.7",
-        "willdurand/js-translation-bundle": "^2.6.4"
+        "willdurand/js-translation-bundle": "^2.6.4",
+        "ezsystems/ezplatform-multi-file-upload": "^0.1.0@dev"
```
bdunogier: Could you move that requirement a bit higher ? The goal is to reduce the likeliness of conflicts for our users. The last one that was added is `ezplatform-i18n/ezplatform-i18n-ach_ug`, this one should go right below it.

app/config/routing.yml
lserwatka: I added new line but for some reason Github cannot see it.

bdunogier:
One retrospect question: with these changes alone, the package won't be usable out-of-the-box, will it ? It requires that the [configuration](https://github.com/ezsystems/ezplatform-multi-file-upload#configuration) is added to the config, does it not ?

Should we consider some default configuration ?

PR #171, opened 2017-04-07
EZP-27226: Move MultiFileUploadBundle above the PlatformUIBundle

sunpietro:
https://jira.ez.no/browse/EZP-27226

This PR fixes the order of bundles inside the `AppKernel.php` file.
The `MultiFileUploadBundle` extends `PlatformUIBundle` hence it should be placed above the `PlatformUIBundle` in the `$bundles` array. This way we can avoid any unexpected issues with the `MultiFileUploadBundle`.

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -36,6 +36,7 @@ class AppKernel extends Kernel
             new eZ\Bundle\EzPublishCoreBundle\EzPublishCoreBundle(),
             new eZ\Bundle\EzPublishLegacySearchEngineBundle\EzPublishLegacySearchEngineBundle(),
             new eZ\Bundle\EzPublishIOBundle\EzPublishIOBundle(),
+            new EzSystems\MultiFileUploadBundle\EzSystemsMultiFileUploadBundle(),
             new eZ\Bundle\EzPublishRestBundle\EzPublishRestBundle(),
             new EzSystems\PlatformUIAssetsBundle\EzSystemsPlatformUIAssetsBundle(),
             new EzSystems\PlatformUIBundle\EzSystemsPlatformUIBundle(),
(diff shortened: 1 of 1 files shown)


sunpietro:
ping @lserwatka @bdunogier @kamilmusial @micszo

sunpietro:
ping @andrerom

yannickroger:
You should have this PR against the 1.9 branch

sunpietro:
@yannickroger I've just switched to 1.9 :)

bdunogier:
I'll check what's happening with Travis.

bdunogier:
I have tested this PR locally, after pulling ezplatform 1.9 and installing dependencies.

PlatformUI loads correctly, which leads me to think that we can merge this. But when trying to upload images, I get a javascript error:

```
Uncaught TypeError: callback is not a function
    at XMLHttpRequest.XHR.onreadystatechange (_ezcombo?app-base/app-base-min.js&querystring-stringify-simple/querystring-stringify-simple-min.js&…:1586)
```

Can you confirm that it is not related to that PR @sunpietro ?

sunpietro:
@bdunogier it's not related to this PR

dpobel:
> @bdunogier it's not related to this PR

this deserves some details. composer installs PlatformUIAssetsBundle 3.10 which contains ez-js-rest-client 1.4 while MultiFileUploadBundle requires ez-js-rest-client 1.5 provided by PlatformUIAssetsBundle 3.2.0-beta1. So a dependency is not set correctly somewhere

bdunogier:
> composer installs PlatformUIAssetsBundle 3.10 which contains ez-js-rest-client 1.4 while MultiFileUploadBundle requires ez-js-rest-client 1.5 provided by PlatformUIAssetsBundle 3.2.0-beta1. So a dependency is not set correctly somewhere

In which case the requirement must be set. But the problem here is that assets-bundle is required by platform-ui (not by ezplatform itself). I need to think about the cleanest way to fix this...

sunpietro:
Maybe the approach from `ezplatform-ee` (keeping dependencies up-to-date) would be a solution? We're using it for months and we have not found any issues like that so far.

yannickroger:
Anyway, the travis problem is not introduced by this PR, so I think we should merge it to be part of the RC.

On the release, we force the version of asset bundle in order to work around this problem.

We can find a way to improve that on travis after the RC.

PR #172, opened 2017-04-10
EZP-27222: Add a warning comment enabling basic rest auth

yannickroger:
Link: https://jira.ez.no/browse/EZP-27222

## Description
Lots of users are not aware of the current limitation with REST basic auth preventing PlatformUI from working. I added a comment in order to make it more clear.

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -10,6 +10,9 @@ security:
             pattern: ^/(_(profiler|wdt)|css|images|js)/
             security: false
 
+        ## WARNING
+        ## Until https://jira.ez.no/browse/EZP-22192 is implemented,
+        ## enabling basic auth in REST will prevent PlatformUI from working
 #        ezpublish_rest:
 #            pattern: ^/api/ezp/v2
 #            stateless: true


yannickroger:
Any idea why travis is not happy?

dpobel:
most likely a "random" failure we got from time to time, I've just restarted the build

app/config/security.yml
```diff
@@ -10,6 +10,9 @@ security:
             pattern: ^/(_(profiler|wdt)|css|images|js)/
             security: false
 
+        ## WARNING
+        ## Until https://jira.ez.no/browse/EZP-27222 is fixed,
```
bdunogier: I'd replace "fixed" with "implemented", as it is a limitation and not a bug.

dpobel: wrong issue number, it should be https://jira.ez.no/browse/EZP-22192

app/config/security.yml
```diff
@@ -10,6 +10,9 @@ security:
...
+        ## enabling basic auth in REST will prevent the PlatformUI from working
```
bdunogier: I'd remove "the".

app/config/security.yml
yannickroger: copy/paste fail…

yannickroger:
Feedback taken into account

PR #173, opened 2017-04-18
[WIP] Update to use Docker 17.04 for improved performance

andrerom:
Consists of three changes:
1. Use Docker 17.04 [`:cached`](https://gist.github.com/yallop/d4af9dd1bb33ae61d48adf86692cdf9e) [(PR)](https://github.com/docker/docker/pull/31047) volume mount point feature to improve performance
2. [WIP] Change nginx config to by default use SYMFONY_ENV=prod for backend & REST url's (`/ez|/api`)

Those two changes improves the performance to a level where working with containers is ok'ish, even on a older 2011 Macbook Pro.

_Note: With changes we do in v2 and further improvements on docker itself last change for nginx conf might not be needed anymore (for people wanting to develop on backend they would anyway want it to also run in DEV)_

Todo:
- [ ] JIRA Issue
- [x] Link to blog post with performance numbers for this: https://ezplatform.com/Blog/A-look-at-Docker-for-Mac-17.04-Performance
- IF use of prod on backend is accepted:
  - [ ] Cleanup Nginx config
  - [ ] Adapt Apache config

Possible Followups:
- eZ Platform:
  - Optional: See if we can do something by default when running server:start as well to get backend to run in prod env _(can probably be done with custom symfony server router, but unsure if we want to go back to that)_
- Docker files:
  - Optional: Move docker compose and docker stuff a place where it can be used to generate flat `docker-compose.yml` file at root for simplified use _(no need to always joggle which files to use)_
  - Get rid of volume usage in prod so we can run on Docker Swarm
    - Affects anything that needs config, data or assets built in: _solr, nginx, mariadb, varnish_
  - Optional: See if we can move to alpine images now that it is supported by blackfire _(but still not available for mysql/mariadb image)_

----
### How to run this:

Before you try, make sure you to install Docker 17.04 or higher *(only available in Docker for Mac edge channel atm)!*

For full instructions see the `doc/docker-compose/Readme.md`, short version for dev setup:
```bash
git clone --branch docker_17.04 https://github.com/ezsystems/ezplatform.git
cd ezplatform
export COMPOSE_FILE=doc/docker-compose/base-dev.yml
# If you have used this before make sure to run `docker-compose pull` first at this point!
docker-compose -f doc/docker-compose/install.yml up --abort-on-container-exit
docker-compose up -d --force-recreate
```

### DIFF
--- a/doc/docker-compose/README.md
+++ b/doc/docker-compose/README.md
@@ -10,7 +10,7 @@
 
 ## Overview
 
-This setup requires Docker Compose 1.9 or higher, and Docker 1.12 or higher. Defaults are set in `.env`, and
+This setup requires Docker Compose 1.12 or higher, and Docker 17.04 or higher. Defaults are set in `.env`, and
 files to ignore are set in `.dockerignore`. By default `.env` specifies that production image is built and setup for use.
 _**NB:** For this and other reasons all docker-compose commands **must** be executed from root of your project directory._
 
--- a/doc/docker-compose/base-dev.yml
+++ b/doc/docker-compose/base-dev.yml
@@ -1,12 +1,12 @@
-version: '2.1'
+version: '3.2'
 # Single server setup for dev
 
 services:
   app:
     image: ${PHP_IMAGE}
     volumes:
-     - ${COMPOSE_DIR}/../../:/var/www
-     - ${COMPOSER_HOME}:/root/.composer
+     - ${COMPOSE_DIR}/../../:/var/www:cached
+     - ${COMPOSER_HOME}:/root/.composer:cached
     depends_on:
      - db
     environment:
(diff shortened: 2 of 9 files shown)


timiTao:
any progres with this feature to add this to platform?

andrerom:
@timiTao None, it was a prototype and would need cleanup on which folders use which mount options, and might need Docker for Mac to also gain support for `:delegate`

timiTao:
@andrerom I run this on my local machine with ezPlatform + ee and working fine. I convince one person in Polish office to do that too on mac. Others consider this, with solution like that.

But we of them manually edit file `base-dev.yml` and ignore this changes in PR.

That is why I asked when will be this officialy 👍 

Moving to new format always be some progress.

Funny things is that, it was one of first things in office to improve. After I did my job mostly, somebody told me from other office, that you did this PR - nobody used  Docker For Mac in my office.

andrerom:
Closing in favor of #192, concept for setting backend is not part of that, but can be opened separately for anyone that wants to try. Instead of the rewrite rules approach attempted here is should perhaps be:
- changes to web/index.php to set to prod when url matches /ez and /rest, and with a env variable to opt out
- changes to stash config to share cache between dev and prod when in file system mode

This way also php built in server _(`app/console server:start`)_ can benefit to.

PR #174, opened 2017-04-26
EZEE-1483: Added `twig/twig` to `composer.json`

webhdx:
**JIRA: https://jira.ez.no/browse/EZEE-1483**

Recent change in the dependencies causes Twig 2.x to install instead of 1.x which was normally provided with our releases. We are not forcing any Twig version to install in our packages which results in broken eZ Platform EE demo installation. More explanations can be found in https://github.com/ezsystems/ezplatform-ee/pull/49

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
     "require": {
         "php": "~5.6|~7.0",
         "symfony/symfony": "~2.8.0",
+        "twig/twig": "~1.28",
         "twig/extensions": "~1.0",
         "symfony/assetic-bundle": "~2.3",
         "symfony/swiftmailer-bundle": "~2.3",


webhdx:
Please hold on with merging I was notified it also affects 1.7 so I will update this PR

bdunogier:
I'm not sure I agree: if something requires Twig 1.x, this something should require it. Or am I missing something ?

andrerom:
@webhdx What @bdunogier  is saying is that ideally this constrints should be in packages that doen't support twig 2, like `knplabs/knp-menu-bundle` if that is still the case there.

That said, while slim, what we risk is that composer just installs a older patch version of `knp-menu-bundle` in order to install Twig 2. So anyway we can make sure `knp-menu-bundle` can be adjusted to support Twig 2?  _( /cc @emodric something you are aware of given you push us towards supporting Twig 2 ;) )_

emodric:
If KNP Menu Bundle is used in some closed source packages (eZ Studio bundles or similar), then this constraint should be specified there, together with KNP Menu Bundle constraint (since KNP constraint for Twig is wrong), or ideally, fixed upstream in Menu Bundle 1.x itself.

We haven't used KNP Menu Bundle 1.x in ages, so we have no problems using Twig 2. In that case, limiting Twig to 1.x here is wrong IMHO.

webhdx:
I don't think `composer.json` from KnpMenu bundle can be fixed - we are using 4 years old release, they are not supporting it anymore. We will update demo to use KnpMenu 2.x which will be fine with both Twig 1.x and 2.x.

The problem I see there is that next EZP release is going to be installed with Twig 2 (as opposed to Twig 1.x in the previous release) but if someone needs 1.x it can be simply required in their package.

andrerom:
> The problem I see there is that next EZP release is going to be installed with Twig 2 (as opposed to Twig 1.x in the previous release) but if someone needs 1.x it can be simply required in their package.

well the temporary fix should then be done on meta repos affected by this, this repo is not, but I guess both platform-demo and platform-ee-demo is then or?

emodric:
> well the temporary fix should then be done on meta repos affected by this

@andrerom Why not in the demo bundles themselves? If KNP Menu Bundle 1.x requirement is specified there, there's no reason why Twig should not be there too.

andrerom:
> @andrerom Why not in the demo bundles themselves? If KNP Menu Bundle 1.x requirement is specified there, there's no reason why Twig should not be there too.

true, but in case of ezplatform-demo, that is for the most part in the meta repo itself currently.

emodric:
> true, but in case of ezplatform-demo, that is for the most part in the meta repo itself.

Right, okay. Haven't looked there in a while ;)

webhdx:
I've checked quickly and `ezplatform-demo` is not affected. It looks like only `ezplatform-ee-demo` has this issue due to an old `knplabs/knp-menu-bundle` package. You can close this PR. I will create a new one in the affected repository.

bdunogier:
Okay, great, thank you @webhdx.

PR #177, opened 2017-05-03
[WIP] Hybrid ui integration

bdunogier:
Integrates [ezsystems/hybrid-platform-ui](http://github.com/ezsystems/hybrid-platform-ui) into ezplatform.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -69,6 +69,8 @@ twig:
 assetic:
     debug: '%kernel.debug%'
     use_controller: '%kernel.debug%'
+    bundles: []
+    #java: /usr/bin/java
     filters:
         cssrewrite: ~
 
@@ -79,3 +81,6 @@ swiftmailer:
     username: '%env(MAILER_USER)%'
     password: '%env(MAILER_PASSWORD)%'
     spool: { type: memory }
+
+fos_js_routing:
+    routes_to_expose: [_ez_content_view]
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "ezsystems/ezpublish-kernel": "7.0.x-dev as 6.99.x-dev",
         "ezsystems/repository-forms": "^1.7@dev",
         "ezsystems/ezplatform-solr-search-engine": "^1.3@dev",
-        "ezsystems/platform-ui-bundle": "^1.9@dev",
+        "ezsystems/platform-ui-bundle": "2.0.x-dev",
         "ezsystems/ez-support-tools": "~0.1.0",
         "ezsystems/ezplatform-http-cache": "^0.1@dev",
         "ezplatform-i18n/ezplatform-i18n-ach_ug": "^1.1@dev",
@@ -47,6 +47,9 @@
         "sensio/framework-extra-bundle": "^3.0.2",
         "ezsystems/content-on-the-fly-prototype": "^0.1.7",
         "willdurand/js-translation-bundle": "^2.6.4"
+        "composer/installers": "dev-ezplatform_assets as 1.3.x-dev",
+        "ezsystems/hybrid-platform-ui": "dev-master as 1.0.x-dev",
+        "ezsystems/hybrid-platform-ui-core-components": "^0.1@dev"
     },
     "require-dev": {
         "ezsystems/ezplatform-i18n": "^1.0@dev",
@@ -100,5 +103,11 @@
         "branch-alias": {
             "dev-master": "2.0.x-dev"
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/dpobel/installers.git"
+        }
+    ]
 }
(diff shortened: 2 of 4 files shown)


andrerom:
Add this on 2.0 branch?

bdunogier:
> Add this on 2.0 branch?

Why not. It will require platform-ui 2.0 anyway.

composer.json
```diff
@@ -14,13 +14,19 @@
         "ezsystems/ezpublish-community": "*"
     },
     "autoload": {
+<<<<<<< HEAD
```
dpobel: oops :p

composer.json
```diff
@@ -44,8 +50,14 @@
...
         "doctrine/doctrine-bundle": "~1.5",
         "sensio/framework-extra-bundle": "~3.0",
-        "ezsystems/content-on-the-fly-prototype": "~0.1.7",
         "willdurand/js-translation-bundle": "^2.6.4"
+
+        ,"friendsofsymfony/jsrouting-bundle": "^1.6",
+        "composer/installers": "dev-ezplatform_assets as 1.3.x-dev",
+        "dpobel/ez-field-edit": "dev-master as 1.0.x-dev",
+        "dpobel/ez-map": "dev-master as 1.0.x-dev",
+        "dpobel/hybrid-platformui-assets": "~0.3",
```
dpobel: I've created and published version 0.1.0 of ezsystems/hybrid-platform-ui-assets which can replace this one. That also means this dependency can (and should) be moved tohybrid-platform-ui composer.json.

composer.json
```diff
@@ -44,8 +50,14 @@
         "hautelook/templated-uri-bundle": "~2.0",
         "doctrine/doctrine-bundle": "~1.5",
         "sensio/framework-extra-bundle": "~3.0",
-        "ezsystems/content-on-the-fly-prototype": "~0.1.7",
         "willdurand/js-translation-bundle": "^2.6.4"
+
+        ,"friendsofsymfony/jsrouting-bundle": "^1.6",
+        "composer/installers": "dev-ezplatform_assets as 1.3.x-dev",
+        "dpobel/ez-field-edit": "dev-master as 1.0.x-dev",
```
dpobel: given content edit is not supposed to work yet, I would remove dpobel/ez-field-edit and dpobel/ez-map

composer.json
```diff
@@ -44,8 +50,14 @@
         "hautelook/templated-uri-bundle": "~2.0",
         "doctrine/doctrine-bundle": "~1.5",
         "sensio/framework-extra-bundle": "~3.0",
-        "ezsystems/content-on-the-fly-prototype": "~0.1.7",
         "willdurand/js-translation-bundle": "^2.6.4"
+
+        ,"friendsofsymfony/jsrouting-bundle": "^1.6",
```
dpobel: this one should rather be in PlatformUIBundle composer.json (and most likely at some point in   ezsystems/hybrid-platform-ui as well)

bdunogier: This one can indeed be moved. It needs to be in hybrid-platform-ui, as the layout.html.twig includes a link to it. I'll add it.

composer.json
bdunogier: Don't be sad, @andrerom, we are just keeping it manageable. The prototypes are still around :)

composer.json
bdunogier: Great, I'll clean that up too.

composer.json
andrerom: I know. And it's not a sad face, it's I want an ice cream face, but I understand it's to cold outside today so I can wait :)

bdunogier:
Should be all cleaned up and installable.

app/config/config.yml
```diff
@@ -67,8 +67,11 @@ twig:
 
 # Assetic Configuration
 assetic:
-    debug: '%kernel.debug%'
+    debug:          '%kernel.debug%'
```
andrerom: can skip this :)

app/config/config.yml
```diff
@@ -67,8 +67,11 @@ twig:
...
     use_controller: '%kernel.debug%'
+    use_controller: "%kernel.debug%"
```
andrerom: duplicate

composer.json
```diff
@@ -47,6 +47,9 @@
         "sensio/framework-extra-bundle": "^3.0.2",
         "ezsystems/content-on-the-fly-prototype": "^0.1.7",
         "willdurand/js-translation-bundle": "^2.6.4"
+        "composer/installers": "dev-ezplatform_assets as 1.3.x-dev",
```
andrerom: something strange here it seems

andrerom:
But maybe rebase away the fixup commits when merging with merge commit _(to be able to revert the disabling of multi upload & cotf later)_

composer.json
bdunogier: `composer validate` is okay with it. Tried to replace it with a fresh `:`, no diff. We'll see :)

bdunogier:
And merged to 2.0.

PR #178, opened 2017-05-04
[Symfony 3.3] Remove trusted_proxies config

emodric:
In Symony 3.3, `trusted_proxies` config has been removed and results in the following exception:

```
[InvalidArgumentException]                                                   
The "framework.trusted_proxies" configuration key has been removed in Symfony
3.3. Use the Request::setTrustedProxies() method in your front controller instead.
```


andrerom:
Does this have a negative impact on symfony 2.x?

emodric:
@andrerom Nope, default value for `trusted_proxies` in Symfony 2.8 is an empty array, whether this config exists with a default value or not.

PR #181, opened 2017-06-07
Fix EZP-27285: Prevent access to website with direct usage of app.php in URL (404)

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27285

# Description

The main propose of this PR is to open discussion about default configuration of URL Rewrite Engine which currently allows access to EZP based websites with direct usage of app.php in URL e.g. 

* http://exemple.com/app.php
* http://example.com/anything/you/want/app.php

Besides of non-loaded assets, it may impact on SEO because of [content duplication](https://support.google.com/webmasters/answer/66359?hl=en).

# TODO: 

- [X] Change vhost template for Apache
- [x] Change vhost template for Nginx  

# Update

This PR solves EZP-27286 issue with a 404 "Not found" response. Alternative solution with 301 redirect: https://github.com/ezsystems/ezplatform/pull/193

### DIFF
--- a/doc/nginx/ez_params.d/ez_rewrite_params
+++ b/doc/nginx/ez_params.d/ez_rewrite_params
@@ -18,7 +18,6 @@ rewrite "^/bundles/(.*)" "/bundles/$1" break;
 rewrite "^/assets/(.*)" "/assets/$1" break;
 
 # Prevent access to website with direct usage of app.php in URL
-# More info here: https://jira.ez.no/browse/EZP-27285
 if ($request_uri ~ "^/(.+/)?app\.php") {
     return 404;
 }
--- a/doc/nginx/ez_params.d/ez_rewrite_params
+++ b/doc/nginx/ez_params.d/ez_rewrite_params
@@ -17,5 +17,11 @@ rewrite "^/w3c/p3p\.xml" "/w3c/p3p.xml" break;
 rewrite "^/bundles/(.*)" "/bundles/$1" break;
 rewrite "^/assets/(.*)" "/assets/$1" break;
 
+# Prevent access to website with direct usage of app.php in URL
+# More info here: https://jira.ez.no/browse/EZP-27285
+if ($request_uri ~ "^/(.+/)?app\.php") {
+    return 404;
+}
+
 rewrite "^(.*)$" "/app.php$1" last;
 
--- a/doc/apache2/vhost.template
+++ b/doc/apache2/vhost.template
@@ -102,6 +102,10 @@
         RewriteCond %{ENV:SYMFONY_ENV} !^(dev)
         RewriteRule ^/(css|js|fonts?)/.*\.(css|js|otf|eot|ttf|svg|woff) - [L]
 
+        # Prevent access to website with direct usage of app.php in URL
+        # More info here: https://jira.ez.no/browse/EZP-27285
+        RewriteRule ^/(.+/)?app\.php - [R=404,L]
+
         RewriteRule .* /app.php
     </IfModule>
 
--- a/doc/apache2/vhost.template
+++ b/doc/apache2/vhost.template
@@ -103,7 +103,6 @@
         RewriteRule ^/(css|js|fonts?)/.*\.(css|js|otf|eot|ttf|svg|woff) - [L]
 
         # Prevent access to website with direct usage of app.php in URL
-        # More info here: https://jira.ez.no/browse/EZP-27285
         RewriteRule ^/(.+/)?app\.php - [R=404,L]
 
         RewriteRule .* /app.php


glye:
Does it behave the same way with or without app.php in the url (apart from assets)? I.e. are there no other known concerns than what you've already mentioned?

We might have rewrite rules that redirects to correct URLs instead of 404, but may not be worthwhile (hardly a common mistake). PR seems ok to me.

adamwojs:
@glye Thank you for your comment.

> Does it behave the same way with or without app.php in the url (apart from assets)? I.e. are there no other known concerns than what you've already mentioned?

In first example (http://exemple.com/app.php) frontend works correctly, but I unable to sign in to Administration Panel. 

I've got malformed API call with double app.php in URL: `http://192.168.56.104/app.php/app.php/api/ezp/v2/user/sessions/5f4ba53c3ca644587b27e7a872f52e96/refresh`, but this is it a separate issue. 

* The second example is relevant only for homepage i.e http://example.com/anything/you/want/app.php/existing-article will return 404 as expected. 

> We might have rewrite rules that redirects to correct URLs instead of 404, but may not be worthwhile (hardly a common mistake).

Yes, this solution is also taken into consideration. Default .htaccess from Symfony standard edition could be an example https://github.com/symfony/symfony-standard/blob/2.8/web/.htaccess#L49

emodric:
We're using the following `.htaccess` rule without any issues:

```
RewriteCond %{ENV:REDIRECT_STATUS} ^$
RewriteRule ^app(_.+)?\.php(/(.*)|$) /$3 [R=301,L]
```

andrerom:
@glye @vidarla was there similar issues solved in legacy at some point?

glye:
As I remember, legacy allows index.php in the URL if you set it up that way, and it will work fine.

vidarl:
In legacy we had the issue that if you where running the site with rewrite rules, but the user still added index.php to the URL, the request would work but the view cache would be polluted with urls also containing  the index.php. That caused problems for users visiting the same page but without index.php in the url. ( I don't remember the exact details of the problem but the cache ended up with links which was not working ).

That is why the ForceVirtualHost ( settings was named something like that, I don't remember the exact naming) setting was introduced. It would make sure legacy did not jump out of "virtual host mode" even if the url contained index.php.

This mess happened because
- We first thought we would be able to autodetect if we where in virtualhost mode or not. Thus initially there was not configuration setting for this.
- Turned out we where not able to detect this in 100% of the cases, so we had to add this ForceVirtualHost setting later on, when this problem was discovered.
- End result is that I guess there are quite a few sites out there which does not have ForceVirtualHost=enabled even though they should....


IMO opinion, we should give 404 if the url is incorrect ( for instance if it contains app.php when it should not .... ). We should at least make sure with don't end up with the same autodetection mess we got into in legacy

vidarl:
With the proposed fix ( rewrite rule with R=404 ) apache will return the 404 ( without site design )... It would look a lot better if ezp generates the 404....

adamwojs:
> With the proposed fix ( rewrite rule with R=404 ) apache will return the 404 ( without site design )... It would look a lot better if ezp generates the 404....

@vidarl Definitely, but this mean we must redirect user to corret URL (without app.php). As long as we don't have dedicated route to display 404 error page, the only option is something like this (based on @emodric snippet, thx!):

```
RewriteCond %{ENV:REDIRECT_STATUS} ^$
RewriteRule ^/(.+/)?app(_.+)\.php(/?.*)?$ $3 [R=301,L]
```
but is little bit complicated in compare to just return 404 by HTTP server. 

I found two additional cases:

1. http://example.com/app.phpsome/valid/id/without/slash
2. http://example.com/anything/you/want/app.php/some/article/id 

In second case application will return response with 404 status code and error:  `No route found for "GET rticle/id"` :confused:

andrerom:
> I found two additional cases:

Do the two approaches suggested that involves solving this in rewrite rules (404 / 301) solve these?

adamwojs:
I've added commit with modified vhost template for Nginx, so comments are welcome.

@andrerom First solution with 404 covers all cases, second solution with 301 will fail on these additional cases. Probably this will require more than one RewriteRule, but I'm still work on this issue.

doc/nginx/ez_params.d/ez_rewrite_params
```diff
@@ -17,5 +17,11 @@ rewrite "^/w3c/p3p\.xml" "/w3c/p3p.xml" break;
 rewrite "^/bundles/(.*)" "/bundles/$1" break;
 rewrite "^/assets/(.*)" "/assets/$1" break;
 
+# Prevent access to website with direct usage of app.php in URL
+# More info here: https://jira.ez.no/browse/EZP-27285
```
andrerom: don't need link to issue btw, it's available via blame. Keep the explanation above

adamwojs: fixed for Apache & Nginx template

andrerom:
Given the `two additional cases` above, does the change here looks sane to you @emodric and @vidarl ?

emodric:
@andrerom Shouldn't we match any `app_*.php` as well as `app.php`?

andrerom:
@emodric given the rewrite rules re passing on to app.php I don't see the need for that, but can maybe add remark inline on how it needs to be adapted of you use different index file in given vh? _(I assume you are thinking about vh for dev use where requests are piped to app_dev.php)_

emodric:
> I assume you are thinking about additional vh for dev use where requests are piped to app_dev.php

Yep :)

andrerom:
@adamwojs could you adapt the inline doc here? basically if user has any other index file like app_dev.php or even index.php, the line will need to be adapted accordingly.

adamwojs:
After further investigation of this issue, I've found a correct solution with a permanent redirect for Apache: https://gist.github.com/adamwojs/809fadddb3390749dd05b3c98d222290

Compared to the original template:
1. Moved and adapted RewriteRules into `<Directory>` directive
2. Added redirection rule based on .htaccess file from Symfony Standard Edition:

```
# Redirect to URI without front controller to prevent duplicate content
# (with and without `/app.php`). Only do this redirect on the initial
# rewrite by Apache and not on subsequent cycles. Otherwise we would get an
# endless redirect loop (request -> rewrite to front controller ->
# redirect -> request -> ...).
# So in case you get a "too many redirects" error or you always get redirected
# to the start page because your Apache does not expose the REDIRECT_STATUS
# environment variable, you have 2 choices:
# - disable this feature by commenting the following 2 lines or
# - use Apache >= 2.3.9 and replace all L flags by END flags and remove the
#   following RewriteCond (best solution)
RewriteCond %{ENV:REDIRECT_STATUS} ^$
RewriteRule ^app\.php(/?(.*)|$) /$2 [R=301,L]
``` 
3. Added disabling MultiViews option if mod_negotiation is enabled 

```
# Disabling MultiViews prevents unwanted negotiation, e.g. "/app" should not resolve
# to the front controller "/app.php" but be rewritten to "/app.php/app".
<IfModule mod_negotiation.c>
    Options -MultiViews
</IfModule>
```
The problem in the previous solution was a context of pattern matching, which is different depending on RewriteRules location: http://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriterule

@andrerom Which solution do we choose?

andrerom:
> @andrerom Which solution do we choose?

Again, I'm not the rewrite expert. So more of an question to others that know ow this a bit, like @vidarl @emodric and others

emodric:
@andrerom Not really much of an expert here too :) My example was also taken from Symfony Standard and modified a bit to match our needs..

adamwojs:
If the decision would be up to me, I would choose a solution with 301 redirect. What do you think @vidarl, @glye? 404 or 301?

andrerom:
I'm indifferent on this, afaik we don't care about site design here if someone on purpose types in app.php in url or similar, so I'm more then fine with that, on the other side I prefer aligning with symfony standard, so if they do redirect on this, so should we.

bdunogier:
I don't have strong feelings about this. I'd say the simpler the better. 
Redirections feel a bit more complicated, and I'm not sure why they're necessary: what are the cases where you actually _end-up_ on an URI with `app.php` in it ? If it is an accidental case, I think it's fine if we get a browser's 404. You can also customize those, and probably should anyway.

PR #182, opened 2017-06-13
Fix EZP-27454: Add the ez-design-engine to ezplatform by default

lolautruche:
> https://jira.ez.no/browse/EZP-27454

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ composer.lock
 behat.yml
 .php_cs.cache
 auth.json
+ide-twig.json
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -48,6 +48,7 @@ class AppKernel extends Kernel
             new EzSystems\RepositoryFormsBundle\EzSystemsRepositoryFormsBundle(),
             new EzSystems\EzPlatformSolrSearchEngineBundle\EzSystemsEzPlatformSolrSearchEngineBundle(),
             new EzSystems\EzContentOnTheFlyBundle\EzSystemsEzContentOnTheFlyBundle(),
+            new EzSystems\EzPlatformDesignEngineBundle\EzPlatformDesignEngineBundle(),
             new Bazinga\Bundle\JsTranslationBundle\BazingaJsTranslationBundle(),
             new JMS\TranslationBundle\JMSTranslationBundle(),
             new AppBundle\AppBundle(),
--- a/app/config/ezplatform_behat.yml
+++ b/app/config/ezplatform_behat.yml
@@ -9,3 +9,7 @@ ezpublish:
         default_siteaccess: "site"
         match:
             URIElement: 1
+
+ezdesign:
+    phpstorm:
+        enabled: false
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "ezsystems/ezplatform-http-cache": "^0.1@dev",
         "ezplatform-i18n/ezplatform-i18n-ach_ug": "^1.1@dev",
         "ezsystems/ezplatform-multi-file-upload": "^0.1.0@dev",
+        "ezsystems/ezplatform-design-engine": "^1.0@dev",
         "egulias/listeners-debug-command-bundle": "~1.9",
         "white-october/pagerfanta-bundle": "1.0.*",
         "hautelook/templated-uri-bundle": "~1.0 | ~2.0",


lolautruche:
Review ping @bdunogier @andrerom @glye

lolautruche:
Not sure Travis failure is because of me

andrerom:
> Not sure Travis failure is because of me

hmm, it might be, that is a new one

andrerom:
@lolautruche It might be you somehow, can't reproduce locally on master. What I can see from travis output is that the application responds with html and not xml, so probably error page. I'll keep on digging a few different directions.

lolautruche:
OK I'll have a look.
Anything I need to know to be able to launch Behat tests locally?

andrerom:
follow steps in .travis.yml, typically:
```bash
export COMPOSE_FILE="doc/docker-compose/base-dev.yml:doc/docker-compose/selenium.yml" SYMFONY_ENV=behat SYMFONY_DEBUG=1

docker-compose -f doc/docker-compose/install.yml up --abort-on-container-exit

# Will be super slow on mac, & will fail on first couple of tests until cache has been warmed up
docker-compose up -d --force-recreate

# First part is not needed on second run against running container
docker-compose exec --user www-data app sh -c "php /scripts/wait_for_db.php; php bin/behat -vv --profile=rest --suite=fullXml --tags=~@broken"
```


Currently expected failure is on just one of the sceanrios:
```
    Then response status code is "201"                                # eZ\Bundle\EzPublishRestBundle\Features\Context\RestContext::assertStatusCode()
      Expected status code '201' found '406'
      Client Exception (406): Argument 'password' is invalid: 'NULL' is wrong value in class 'UserCreateStruct'
```

But not the one that came here:
```
      Behat\Testwork\Call\Exception\FatalThrowableError: Fatal error: [] operator not supported for strings in vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Handler/Xml.php:164
      Stack trace:
      #0 vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Handler/Xml.php(164): eZ\Publish\Core\REST\Common\Input\Handler\Xml->convertDom(Object(DOMElement))
      #1 vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Handler/Xml.php(155): eZ\Publish\Core\REST\Common\Input\Handler\Xml->convertDom(Object(DOMElement))
      #2 vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Handler/Xml.php(155): eZ\Publish\Core\REST\Common\Input\Handler\Xml->convertDom(Object(DOMElement))
      #3 vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Handler/Xml.php(155): eZ\Publish\Core\REST\Common\Input\Handler\Xml->convertDom(Object(DOMElement))
      #4 vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Handler/Xml.php(155): eZ\Publish\Core\REST\Common\Input\Handler\Xml->convertDom(Object(DOMElement))
      #5 vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Handler/Xml.php(155): eZ\Publish\Core\REST\Common\Input\Handler\Xml->convertDom(Object(DOMElement))
      #6 vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Handler/Xml.php(117): eZ\Publish\Core\REST\Common\Input\Handler\Xml->convertDom(Object(DOMDocument))
      #7 vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/REST/Common/Input/Dispatcher.php(101): eZ\Publish\Core\REST\Common\Input\Handler\Xml->convert('<!DOCTYPE html>...')
```

.gitignore
```diff
@@ -48,3 +48,4 @@ composer.lock
 behat.yml
 .php_cs.cache
 auth.json
+ide-twig.json
```
andrerom: I guess this should be separate, or is it related?

bdunogier: It is related. The file is generated by the bundle to enhance completion for what it handles.

bdunogier:
I'm tempted to tag a beta2 with this. Any counter opinion ?

.gitignore
lolautruche: Definitely related 😃 .
This file is generated for completion & goto in PHPStorm

lolautruche:
@bdunogier Please let me first fix the tests :-)

bdunogier:
> @bdunogier Please let me first fix the tests :-)

Sure thing. But I'm still tempted ;)

andrerom:
>  so I'll probably try to dump out the error response code from Travis 😝

yeah, the code there obviously need some more error checks, it seems it assumes it is xml when it is actually html from the looks of it. And getting dump of that html would give a clue to what we are dealing with.

emodric:
Just to be clear on some points:

Even if the bundle is active, it does not change anything concerning current way frontend templates are implemented? That is, when we upgrade, old templates and design will keep working as is, without any issues?

Can we "activate" this bundle and use it for new templates, while keeping the old ones the way they are?

What needs to be done to "activate" the design support in an existing project?

lolautruche:
@emodric 

There is no impact on old templates, so there is no risk upgrading.
To use the new design engine, you just need to start using `@ezdesign` Twig namespace, and `ezdesign` asset package.

lolautruche:
@andrerom I don't manage to reproduce the fatal locally. I do have the `406` error, but not the fatal.

andrerom:
ah, so you are trying to write to root _(ide-twig.json)_, that won't work in prod indeed.

lolautruche:
The error, finally: `Unable to write to the /var/www directory.`

lolautruche:
I don't, but anyway I'm gonna disable `ide-twig.json` generation for tests

lolautruche:
> that won't work in prod indeed.

Default value is `%kernel.debug%`... My guess is to disable it in `ezplatform_behat.yml`

glye:
> ah, so you are trying to write to root (ide-twig.json), that won't work in prod indeed.

Platform.sh doesn't like that either, dev or prod :)

andrerom:
> Default value is %kernel.debug%... My guess is to disable it in ezplatform_behat.yml

you can also change SYMFONY_DEBUG on the given .travis.yml line where we use prod container, it will make it harder to find issues if ther eis one there next time, but it would be slightly more correct prod setup.

lolautruche:
Well, I think the best fix here is to prevent `ide-twig.json` generation, as it's useless here

andrerom:
> Well, I think the best fix here is to prevent ide-twig.json generation, as it's useless here

ah, in the behat file, yes indeed.

lolautruche:
Travis seems to hang on the last test suite... Could you please relaunch it?

lolautruche:
Finally passing 🎉

lolautruche:
@bdunogier @andrerom Could you please merge this into `2.0` branch?

PR #183, opened 2017-06-13
[TEST] Testing if issue in #182 affects master

andrerom:
Can't reproduce issue in #182 locally, so this is purely to see if master is affected without design bundle.

Turns on debug, if this passes _(fails on password atm)_, then should be rerurn with debug off to make sure that is not relevant.


andrerom:
_(it did not, debugging continues in linked PR)_

PR #186, opened 2017-06-16
[WIP] EZP-27499: updated platform.sh hooks for v2.x

SylvainGuittard:
Jira issue: https://jira.ez.no/browse/EZP-27499
# Description
eZ Platform v2 is using Symfony 3, so we have to update the hooks (build and deploy).

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -36,21 +36,21 @@ disk: 2048
 
 # The mounts that will be performed when the package is deployed.
 mounts:
-    "/app/cache": "shared:files/cache"
-    "/app/logs": "shared:files/logs"
+    "/var/cache": "shared:files/cache"
+    "/var/logs": "shared:files/logs"
+    "/var/sessions": "shared:files/sessions"
     "/web/var": "shared:files/files"
 
 # The hooks that will be performed when the package is deployed.
 hooks:
     build: |
         set -e
-        composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader
-        rm web/app_dev.php
+        composer install --prefer-dist --no-progress --no-interaction --optimize-autoloader
         . ./.env
         if [ -z "$SYMFONY_ENV" ]; then
             export SYMFONY_ENV=prod
         fi
-        app/console --env=$SYMFONY_ENV assetic:dump
+        bin/console --env=$SYMFONY_ENV assetic:dump
     deploy: |
         set -e
         . ./.env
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -46,8 +46,7 @@
         "sensio/framework-extra-bundle": "^3.0.2",
         "willdurand/js-translation-bundle": "^2.6.4",
         "composer/installers": "dev-ezplatform_assets as 1.3.x-dev",
-        "ezsystems/hybrid-platform-ui": "dev-master as 1.0.x-dev",
-        "ezsystems/hybrid-platform-ui-core-components": "^0.1@dev"
```
bdunogier: Why do you remove `"ezsystems/hybrid-platform-ui-core-components` ? I'm pretty sure we do need it :)

.platform.app.yaml
```diff
@@ -19,6 +19,7 @@ build:
 # side is in the form `<service name>:<endpoint name>`.
 relationships:
     database: "mysqldb:mysql"
+    cache: "cache:memcache"
```
andrerom: Should not deal with memcached here _(it's not used as it's not configured)_. This should anyway should be separate PR, and most likely towards master or 1.7 ad not limited to 2.0.

.platform.app.yaml
```diff
@@ -36,32 +37,33 @@ disk: 2048
...
         if [ -z "$SYMFONY_ENV" ]; then
             export SYMFONY_ENV=prod
         fi
         if [ ! -f web/var/.platform.installed ]; then
-            php -d memory_limit=-1 app/console ezplatform:install --env=$SYMFONY_ENV $INSTALL_EZ_INSTALL_TYPE
+            php -d memory_limit=-1 bin/console ezplatform:install --env=$SYMFONY_ENV $INSTALL_EZ_INSTALL_TYPE
             touch web/var/.platform.installed
         fi
-        app/console --env=$SYMFONY_ENV cache:clear
+        composer update --no-dev --prefer-dist
```
andrerom: why was this added? is it that you you needed to clear cache and re-generate parameters, if so `composer run-script post-update-cmd` should do

.platform/services.yaml
```diff
@@ -1,3 +1,5 @@
 mysqldb:
-    type: mysql:10.0
+    type: 'mysql:10.0'
```
andrerom: could mope this change to 1.7 to so we merge it up instead of ending up with potentially future conflicts in 2.0 branch.

composer.json
SylvainGuittard: It's just a test because, the bundle appears as installed in the build log. But not available in the vendor directory and I have a 404 in the admin interface...
Because on platform.sh, you need a commit to rebuild / redeploy the website, I decided to remove temporary `ezsystems/hybrid-platform-ui-core-components` and see what happens... Did not have time to continue. Maybe tonight :D

.platform.app.yaml
SylvainGuittard: Agree, I will remove it

.platform.app.yaml
SylvainGuittard: Actually, it's related to https://github.com/ezsystems/ezplatform/pull/186#discussion_r123078326
I wanted to force the install / update of `ezsystems/hybrid-platform-ui-core-components`

andrerom: Ok, but it kind of make sense that you need to do it the post update scripts, given the usage of env variables to parameters.yml mapping in composer.json atm.

SylvainGuittard:
Closing so you won't receive a lot of notifications.
Working on my fork. I will submit another PR when everything will be ready.

PR #188, opened 2017-06-21
EZP-27542: Add support for using Varnish with Docker

andrerom:
> Issue: https://jira.ez.no/browse/EZP-27542

Alternative try to #143.

Adds:
- possibility to inject http cache config by env variables
- varnish container and docker compose setup taking advantage of this
- test it on travis using our existing behat setup
- doc updates for this


Plopix:
Hello @andrerom,

I think this PR should be split into 2, the 3 files for the configuration injection on one side and the docker/automated test complexity on this other.

```
app/config/default_parameters.yml
app/config/env/docker.php
app/config/ezplatform.yml
```

I have a use case in PROD, on AWS where I would love to just have to:
- change a ENV variable
- clear the cache

Even better and more globally it will be great to use that:
http://symfony.com/blog/new-in-symfony-3-2-runtime-environment-variables

It would avoid a cache clear.

My scenario if for any reason I am adding a Varnish, I need to update the config of the web server.

Let me know I can PR something especially for that.

++

andrerom:
> Even better and more globally it will be great to use that:

This is 1.7 (symfony 2.8), implied is that it will need to be adjusted when merged into 2.0 branch

For the rest, I'll see if I can get the container running, if not I'll split up.

andrerom:
Passing 🎉 

atm only platformui profile failing _(not setup to use varnish)_ because of good old issue with stash when deleting files _([random bug](https://github.com/tedious/Stash/issues/359))_, will attempt to get that one to run with redis _(and fallback to memcached if that does not work_).

andrerom:
Ready for review


## Review notes
The VCL added here is maybe something that could be part of the main one, but the difference is basically:
```diff
diff --git a/doc/varnish/vcl/varnish4.vcl b/doc/varnish/vcl/varnish4.vcl
index 8d743b6..7be2e95 100644
--- a/doc/varnish/vcl/varnish4.vcl
+++ b/doc/varnish/vcl/varnish4.vcl
@@ -6,20 +6,21 @@ vcl 4.0;
 // Our Backend - Assuming that web server is listening on port 80
 // Replace the host to fit your setup
 backend ezplatform {
-    .host = "127.0.0.1";
+    .host = "web";
     .port = "80";
 }
 
 // ACL for invalidators IP
 acl invalidators {
     "127.0.0.1";
-    "192.168.0.0"/16;
+    "172.16.0.0"/20;
+    "app";
 }
 
 // ACL for debuggers IP
 acl debuggers {
     "127.0.0.1";
-    "192.168.0.0"/16;
+    "172.16.0.0"/20;
 }
```
So it's a bit specific.

Plopix:
+1 for the PR

About the Varnish config, to avoid 2 files,  we could:
a). do a script like for the vhost.sh that would make the changes and we would have only one VCL. (sounds complex)
b) split the VCL in different files and define the backend in a specific file (maybe the good one)
c) directly change the main doc/varnish/vcl/varnish4.vcl, and a mention in the doc to explain.

c) is my preferred way

andrerom:
Thanks for the fixes @glye.

@plopix I'm also fine with c) or b), but b) would bump requirement to 5.x right? @bdunogier what's your take/pref?

Plopix:
@andrerom include works on 4.x ;)

andrerom:
@Plopix Actually include also works on 3.x, so adapted to get rid of the vcl duplication. Ok @Plopix / @glye ?

glye:
:+1: Red code is dead code is good code.

Plopix:
👍  yes that is cool!

PR #192, opened 2017-07-12
EZP-27788: Update and cleanup docker config for Docker 17.06 features

andrerom:
> Issue: https://jira.ez.no/browse/EZP-27788
> Epic: https://jira.ez.no/browse/EZP-26310

Docker 17.06 came out in June, and with it these notable features are now in stable channel which we haven't started using yet:
- `:cached`, partly improving mac/windows performance _(as tested in #173)_
- multi stage builds, letting us end up with much smaller builds _(as tested in ezsystems/ezplatform-ee-demo#33)_
- healtcheck, allowing us to make sure a given container is running before we try to access it

### Todo / TBD
- [x] ~~Currently blocked on bug in Docker 17.06 that is solved in 17.06.01-rc~~ _false poitive during testing, the issue was something else_ 😕 
- [x] ~~Update doc to refer to launchpad for simple use cases~~_commited to active branches ahead of 1.11 release in 7f2abb5acd555a0c70a8a5c77c04814be2e3af41_
- [x] Rename doc/docker-compose to `doc/docker` and adapt yml/doc given it is about more then just docker-compose now _( and shorter paths ;) )_
- [x] Test on kernel repo to make sure this does not break sub repo testing: _[Passing](https://github.com/ezsystems/ezpublish-kernel/pull/2059)_

### Possible followups:
- Docker Swarm work for use on own demo systems _( ezsystems/ezplatform-ee-demo#33)_
- Assuming https://github.com/docker/compose/pull/5064 is merged and released, solve issue where project name is same on every platform checkout as it's taken from "docker" folder
- #189 Either as is, or proposing further fixes to compose to have variable for project name available, or check if we can omit setting a image name if that also works with swarm somehow
- Heath checks should not use "example" containers from Docker but rather extend the official once and add logic ourselves.
  - For all shell approach to not have to install bash on alpine see: https://github.com/dkarlovi/symfony-flex-docker/tree/master/.infra/docker/db
  - While on it, we can also consider adding some sort of health check for app container, however unsure if we should check db given that is done in db container: https://github.com/phudinsky/symfony-docker/blob/master/bin/healthcheck


doc/docker/Dockerfile-varnish
```diff
@@ -0,0 +1,17 @@
...
+    curl -s https://packagecloud.io/install/repositories/varnishcache/varnish5/script.deb.sh | bash && \
+    apt-get install -q -y --allow-unauthenticated --no-install-recommends varnish && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY doc/varnish/vcl/varnish4.vcl /etc/varnish/default.vcl
+COPY doc/docker/entrypoint/varnish/parameters.vcl /etc/varnish/parameters.vcl
+
+EXPOSE 80 6082
+
+CMD ["varnishd", "-F", "-a", ":80", "-T", ":6082", "-f", "/etc/varnish/default.vcl", "-s", "malloc,256M"]
```
vidarl: Would it not be nice if "256M" would be defined in a variable in .env ?
This is a setting which is typically customized....

andrerom: true, can try to use env variables in build args.

PR #195, opened 2017-07-18
Excluded fos_js_routing from the firewall (EZP-27675)

bdunogier:
> Fixes [EZP-27675](https://jira.ez.no/browse/EZP-27675)

Prevents it from being used as a redirect URI after login on the v2 UI.


webhdx:
@bdunogier Maybe something more specific like `(js\/routing)` perhaps?

adamwojs:
As same as @webhdx, I think that when dealing with disabling security for something, is good to be as precise as possible. Does `fos_js_routing` is used outside administration UI? If not, it can be even  an `admin\/js`.

SylvainGuittard:
@adamwojs It cannot be `admin\/js` because `admin` is a `siteaccess`.

PR #196, opened 2017-07-28
Fix-EZP-27724: Unable to run dockerized Solr

kmadejski:
JIRA issue: [https://jira.ez.no/browse/EZP-27724](https://jira.ez.no/browse/EZP-27724)

# Description
As described in issue, run Solr is impossible because in the newest version of Solr ``6.6.0`` field ``defaultSearchField`` is deprecated. 
This PR changes dockerized version of Solr to ``6.5``.


kmadejski:
Another solution was chosen: https://github.com/ezsystems/ezplatform-solr-search-engine/pull/97

PR #197, opened 2017-08-15
Docker workflow upgrade

fchris82:
It is better way when the developer can use custom `.env`, and can change eg. the `HTTP_HOST` and `HTTP_PORT`.
For example I am using the `8080` port by other program and I prefer the unique domain per project because of cookies and remembered usernames/passwords.

So I want to use:
```env
HTTP_PORT=80
HTTP_HOST=concrate-ez-project.loc
```


fchris82:
I have been trying the Launchpad. I like the concept, I wanted to similar :) I'm reviewing and when I miss something, I will make pull request there or jira ticket.

PR #198, opened 2017-08-24
Updated platform.sh settings

L0rdJ:
- Hardcoded cluster DB name
- Cache settings
- Disabling PHPStormPass (makes impossible to run ezplatform in dev mode on platform.sh)

### DIFF
--- a/app/config/env/platformsh.php
+++ b/app/config/env/platformsh.php
@@ -33,6 +33,15 @@ if (isset($relationships['cache'])) {
         $container->setParameter('cache_host', $endpoint['host']);
         $container->setParameter('cache_memcached_port', $endpoint['port']);
     }
+} elseif (isset($relationships['redis'])) {
+    foreach ($relationships['redis'] as $endpoint) {
+        if ($endpoint['scheme'] !== 'redis') {
+            continue;
+        }
+
+        $container->setParameter('cache_host', $endpoint['host']);
+        $container->setParameter('cache_redis_port', $endpoint['port']);
+    }
 }
 
 // Disable PHPStormPass
--- a/app/config/env/platformsh.php
+++ b/app/config/env/platformsh.php
@@ -19,7 +19,24 @@ foreach ($relationships['database'] as $endpoint) {
     $container->setParameter('database_user', $endpoint['username']);
     $container->setParameter('database_password', $endpoint['password']);
     $container->setParameter('database_path', '');
+
+    // Cluster DB name is hardcoded. It will have no any effect if cluster is disabled
+    $container->setParameter('cluster_database_name', 'cluster');
 }
+
+if (isset($relationships['cache'])) {
+    foreach ($relationships['cache'] as $endpoint) {
+        if ($endpoint['scheme'] !== 'memcached') {
+            continue;
+        }
+
+        $container->setParameter('cache_host', $endpoint['host']);
+        $container->setParameter('cache_memcached_port', $endpoint['port']);
+    }
+}
+
+// Disable PHPStormPass
+$container->setParameter('ezdesign.phpstorm.enabled', false);
 
 // Store session into /tmp.
 ini_set('session.save_path', '/tmp/sessions');


app/config/env/platformsh.php
```diff
@@ -19,7 +19,24 @@
...
+}
+
+if (isset($relationships['cache'])) {
+    foreach ($relationships['cache'] as $endpoint) {
+        if ($endpoint['scheme'] !== 'memcached') {
+            continue;
+        }
+
+        $container->setParameter('cache_host', $endpoint['host']);
+        $container->setParameter('cache_memcached_port', $endpoint['port']);
```
andrerom: while at it, can you also add redis port support? see docker.php

app/config/env/platformsh.php
```diff
@@ -19,7 +19,24 @@
     $container->setParameter('database_user', $endpoint['username']);
     $container->setParameter('database_password', $endpoint['password']);
     $container->setParameter('database_path', '');
+
+    // Cluster DB name is hardcoded. It will have no any effect if cluster is disabled
+    $container->setParameter('cluster_database_name', 'cluster');
```
andrerom: where is this defined? Or do you use it in custom config?

L0rdJ:
Redis settings are added

L0rdJ:
Regarding cluster settings, we have following:
```
doctrine:
    dbal:
        connections:
            dfs:
                driver: %database_driver%
                host: %database_host%
                dbname: %cluster_database_name%
                user: %database_user%
                password: %database_password%
                charset: UTF8
            default:
                driver: %database_driver%
                host: %database_host%
                port: %database_port%
                user: %database_user%
                password: %database_password%
                dbname: %database_name%
                charset: UTF8
```

Clustering is not enabled by default, and we are assuming cluster and main databases are using the same entry point on platform.sh side.

L0rdJ:
So database services on platfrom.sh is following
```
mysqldb:
    type: mysql:10.0
    disk: 512
    configuration:
        schemas:
            - main
            - cluster
        endpoints:
            mysql:
                default_schema: main
                privileges:
                    main: admin
                    cluster: admin
```
`.platform.app.yaml`:
```
relationships:
    database: "mysqldb:mysql"
```
And it has following JSON in it's `PLATFORM_RELATIONSHIPS` env variable:
```
{
   "database" : [
      {
         "rel" : "mysql",
         "username" : "mysql",
         "path" : "main",
         "query" : {
            "is_master" : true
         },
         "cluster" : "osub47flwvg6u-master-7rqtwti",
         "service" : "mysqldb",
         "scheme" : "mysql",
         "host" : "database.internal",
         "port" : 3306,
         "password" : "ef310cf4ee0c69a6efdc692f4b1c6e6e",
         "ip" : "246.0.176.199"
      }
   ]
}
So in this case it impossible to extract cluster DB name without adding additional relationship.

**Comment by andrerom** [comment]
Thanks, asking since we would kind of need to add defaults for this that makes sense for both normal use _(so it won't be used)_ and perhaps change default platform.sh config to always take advantage.

**Comment by L0rdJ** [comment]
Another option would be to add "cluster" relationship and extract cluster DB parameters from there. 
So services definition on platfrom.sh would be something like:
```
mysqldb:
    type: mysql:10.0
    disk: 512
    configuration:
        schemas:
            - main
            - cluster
        endpoints:
            main:
                default_schema: main
                privileges:
                    main: admin
            cluster:
                default_schema: cluster
                privileges:
                    cluster: admin
```
`.platform.app.yaml:`
```
relationships:
    database: "mysqldb:main"
    cluster: "mysqldb:cluster"
```
And extract cluster DB name (and probably host/user/pass/port) from `$relationships['cluster']`

But i`m not really sure if there is any sense of doing that. What do you think about it?

andrerom:
> But i`m not really sure if there is any sense of doing that. What do you think about it?

no opinion TBH

app/config/env/platformsh.php
vidarl: FYI : I have changed this approach in https://github.com/ezsystems/ezplatform/pull/263

PR #204, opened 2017-08-29
EZP-27499: updated platform.sh hooks for v2.x

SylvainGuittard:
Jira issue: https://jira.ez.no/browse/EZP-27499

# Description
eZ Platform v2 is using Symfony 3, so we have to update the hooks (build and deploy).

Based on the previous PR #186


andrerom:
I guess there needs to be some adoptions on `app/config/env/platformsh.php` to right? We need to somehow set the env variables needed from the format provided by platform.sh

.platform.app.yaml
```diff
@@ -36,32 +36,34 @@ disk: 2048
...
         if [ -z "$SYMFONY_ENV" ]; then
             export SYMFONY_ENV=prod
         fi
         if [ ! -f web/var/.platform.installed ]; then
-            php -d memory_limit=-1 app/console ezplatform:install --env=$SYMFONY_ENV $INSTALL_EZ_INSTALL_TYPE
+            php -d memory_limit=-1 bin/console ezplatform:install --env=$SYMFONY_ENV $INSTALL_EZ_INSTALL_TYPE
             touch web/var/.platform.installed
         fi
-        app/console --env=$SYMFONY_ENV cache:clear
+        composer update --no-dev --prefer-dist
```
andrerom: Do you know why you added this btw?

andrerom:
> I guess there needs to be some adoptions on `app/config/env/platformsh.php` to right? We need to somehow set the env variables needed from the format provided by platform.sh

Aha, the env variables added here are hardcoding env variables instead right? So fixing `app/config/env/platformsh.php` will remove the need to do that. I can have look, but please reply to question below.

.platform.app.yaml
SylvainGuittard: Oups, I don't remember... 😄 
It might be a mistake.

SylvainGuittard:
> Aha, the env variables added here are hardcoding env variables instead right?
Yes. I agree, fixing `app/config/env/platformsh.php` is a better option.

andrerom:
Closing as work here was pulled into master (cabfe49cd74dcb4ba111bd3b9bf04c8787adfac5, ddba987e516b74aaad589ce4865a16b2ef206b4b, 2677e498701e3d9122ec874a4844685ed65160d3, f5d64edb110998f97bf8b1b98406b73b30ae4ba8) / 2.0 (fa4aa9b) where applicable

PR #205, opened 2017-08-30
Docker stack

vidarl:
This PR adds examples on how to use ezplatform with docker stack (swarm).

It is basically a rebase of https://github.com/ezsystems/ezplatform-ee-demo/pull/33 which was made against ee-demo for testing purposes.

However, it should be reviewed again as there has been some changes since ee-demo PR...


Dockerfile
```diff
@@ -13,14 +10,28 @@ COPY . /var/www
 RUN if [ -d .git ]; then echo "ERROR: .dockerignore folders detected, exiting" && exit 1; fi
 
 # Install and prepare install
-RUN mkdir -p web/var \
-    && composer install --optimize-autoloader --no-progress --no-interaction --no-suggest --prefer-dist \
+RUN mkdir -p web/var
+# For now, only run composer in order to generate parameters.yml
+RUN composer run-script post-install-cmd --no-interaction
```
vidarl: When work was based on ezplatform-ee-demo, I instead had:

    RUN composer run-script build --no-interaction
But that results in this failure on ezplatform:

    Script "build" is not defined in this package

I guess running post-install-cdm is just as good?

andrerom: yes, it is the safest across versions. master now renamed it to `symfony-scripts` to align with symfony standard.

doc/docker/my-ez-full-app-stack.yml
```diff
@@ -0,0 +1,80 @@
...
+     - HTTPCACHE_PURGE_SERVER=http://varnish
+     - PHP_INI_ENV_session.save_handler=redis
+     - PHP_INI_ENV_session.save_path="tcp://redis:6379?weight=1"
+    volumes:
+     - vardir:/var/www/web/var
+    networks:
+     - stack-db_db
+     - my-ez-app
+    deploy:
+      replicas: 5
```
andrerom: Any specific reason why we went for 5 here and 2 for web?

doc/docker/db-stack.yml
```diff
@@ -0,0 +1,26 @@
+version: '3.2'
```
andrerom: 3.3

doc/docker/create-dataset.yml
```diff
@@ -0,0 +1,14 @@
+version: '2'
```
andrerom: for consistency we can move all to 3.3 right?

.env
```diff
@@ -19,6 +19,15 @@ REDIS_IMAGE=healthcheck/redis
 # App image name for use if you intend to push it to docker registry/hub.
 APP_PROD_IMAGE=my-ez-app
 APP_DOCKER_FILE=Dockerfile
+NGINX_PROD_IMAGE=my-ez-web
```
andrerom: I'm a bit unsure about these, because of #189, and would avoid having to always customize this per project.

What I was aiming to try top get docker-compose to get support for:
- COMPOSE_PROJECT_DIRECTORY  https://github.com/docker/compose/pull/5064
- COMPOSE_PROJECT_NAME always available https://github.com/docker/compose/pull/5026

However the last PR hints docker folks don't like this, so workaround would be if we have one `.env` variable setting COMPOSE_PROJECT_NAME, and use this in the pattern for all  images we need to name, removing all `_PROD_IMAGE` env variables.

Either way benefit is that both service names and the image names will be prefixed with project name avoid conflicts when running several projects at the same time.

doc/docker/my-ez-full-app-stack.yml
vidarl: Docker stack does not support pods; you can't bind phpfpm and nginx containers in pairs.
Therefore, any nginx container can hit any phpfpm container.
And since nginx has `worker_connections=1000` and phpfpm has `pm.max_children=5` you will typically need more phpfpm containers than nginx. It does not make much sense to have same amount of both....
Yes, nginx will also serve static content without using phpfpm, but the 1000 to 5 ratio still a bit much.

But the 2 to 5 ratio was just a hunch. I haven’t benchmarked anything in order to give an ideal number.

ps : I think I will make separate PRs for customizing those nginx and phpfpm settings

.env
vidarl: I suggest we then leave this as-is and once we have a decision on what to do, we remove NGINX_PROD_IMAGE and APP_PROD_IMAGE at the same time.

Note that .env file is ignored by docker stack. So, once we rely on a dynamic project name, the docker stack instructions must be changes as user would then likely need to substitute some image names before being able to get it to work.

doc/docker/create-dataset.yml
vidarl: Fixed

doc/docker/db-stack.yml
vidarl: fixed

.env
andrerom: 👍

doc/docker/my-ez-full-app-stack.yml
andrerom: ok, clear

vidarl:
I intend to use this on sales demo too, yes. But that requires an additional PR which I am working on

vidarl:
Don't merge this yet. I don't thing we need both `doc/docker/Dockerfile-vardirdump` and `doc/docker/Dockerfile-demo-vardir`

vidarl:
I have fixed the `doc/docker/Dockerfile-demo-vardir` issue + also cleaned up quite a few things. For instance, I got rid the a few other settings in .env

doc/docker/README.md
```diff
@@ -81,6 +87,9 @@ by default under the hood, which leads to much slower IO performance.*
 
 From root of your projects clone of this distribution, [setup composer auth.json](#composer) and execute the following:
 ```sh
+# If you already have followed the intructions in the 'demo "image" usage' chapter, run
+unset COMPOSE_FILE
```
andrerom: this is covered in the end of the file, unsure if we need to mentioned it in beginning of each section

vidarl: Ok, removed again

vidarl:
Change base branch to 1.7 and squashed
Branch should  be ready for merge

Dockerfile
vidarl: Actually, post-install-cmd seems to fail in 1.7
So, should I use build there, or should post-install-cmd be backported?

Dockerfile
```diff
@@ -13,14 +10,28 @@ COPY . /var/www
...
+# urls like http://localhost:8080/_ezcombo?/bundles/ezstudioui/js/views/ezs-landingpageview.js&/tpl/handlebars/studiolandingpageconfigview-ez-template.js&/bundles/ezstudioui/js/views/ezs-landingpageconfigview.js&/tpl/handlebars/studiolayoutselectorview-ez-template.js&/bundles/ezstudioui/js/views/ezs-layoutselectorview.js&/tpl/handlebars/studiolandingpageconfigpopupformview-ez-template.js&/bundles/ezstudioui/js/views/forms/ezs-landingpageconfigpopupformview.js&/tpl/handlebars/landingpagecreatorview-ez-template.js&/bundles/ezsystemsformbuilder/js/models/fb-formfield-model.js&/bundles/ezsystemsformbuilder/js/lists/fb-formfields-modellist.js&/bundles/ezsystemsformbuilder/js/models/fb-formpage-model.js&/bundles/ezsystemsformbuilder/js/lists/fb-formpages-modellist.js&/bundles/ezsystemsformbuilder/js/models/fb-form-model.js&/tpl/handlebars/fbbasetabview-ez-template.js&/bundles/ezsystemsformbuilder/js/tabs/fb-base-tabview.js&/tpl/handlebars/fbpanelview-ez-template.js&/bundles/ezsystemsformbuilder/js/panels/fb-panelview.js
+# will not work when loading http://localhost:8080/ez
+RUN rm -rf web/css web/fonts web/js web/var
```
andrerom: Maybe add comment on why this is removed, is it because it is served by ngninx in prod ?

vidarl: Added

vidarl:
With latest fixup (https://github.com/ezsystems/ezplatform/pull/205/commits/ca6b4b54739f84c6413832d6c79ad88310bcbe05) travis should pass

If you prefer I can copy the code in create_dataset.yml into demo.yml to save a `-f ...`  when doing "demo use".. It is just a few lines of code which then would be duplicated.

.travis.yml
```diff
@@ -16,7 +16,7 @@ env:
     - SYMFONY_DEBUG=1
   # list of behat arguments to test
   matrix:
-    - TEST_CMD="bin/behat -vv --profile=rest --suite=fullJson --tags=~@broken" COMPOSE_FILE="doc/docker/base-prod.yml:doc/docker/varnish.yml:doc/docker/selenium.yml" WEB_HOST="varnish"
+    - TEST_CMD="bin/behat -vv --profile=rest --suite=fullJson --tags=~@broken" COMPOSE_FILE="doc/docker/base-dev.yml:doc/docker/varnish.yml:doc/docker/selenium.yml" WEB_HOST="varnish"
```
andrerom: why did you have to change this one?

vidarl: Because base-prod.yml now also requires -f `doc/docker/create-dataset.yml -f doc/docker/demo.yml` or some other means of importing the database and vardir, ref. https://github.com/ezsystems/ezplatform/pull/205/files#diff-ec06f6fb92ec24b756398a53162b1b54R62

As I tried to communicate in the previous post I could duplicate the code in create-dataset.yml into demo.yml so that  base-prod.yml would only require demo.yml

andrerom: ok, missed that part. but 1. then not really any point calling anything `base` anymore if it can't be used that way as drop in replacement for base dev. 2. this will then break https://github.com/ezsystems/PlatformUIBundle/blob/master/.travis.yml#L20

vidarl: Previously, base-prod kinda did two things:
* set SYMFONY_ENV=prod
* created production images of php and nginx, so doc-root was not bind-mounted

So, in my mind, the old base-prod was intended for production use, but it was not quite up to the task. Thus, I changed it. However, there is maybe a need for running tests with SYMFONY_ENV=prod too?

So we basically needs these use cases? :
1. demo use
2. dev use ( SYMFONY_ENV=dev ), with bind mount
3. dev use ( SYMFONY_ENV=prod ), with bind mount
4. production use

`#3` is ATM not supported ( and never was )

andrerom: @vidarl The linked PlatformUIBundle PR has been merged, but it's made the behat tests run slower. Probably because dev uses xdeub and configures php for dev use. So we are not done here I'm afraid.

andrerom: PR linked to above has been merged, it's slow _(>= 20min)_, but seems to have been the case for a long time, so something to look into separately.

PR #207, opened 2017-09-01
[Docker] Removing image tags : my-ez-app and my-ez-web

vidarl:
This PR is crated as a spin-off of https://github.com/ezsystems/ezplatform/pull/205#pullrequestreview-59508996 and the discussion started there. 

FYI : Only the last commit matters. The other commits are part of #205


vidarl:
Basically, there are no benefits of explicitly naming the php or nginx image explicit IMO.
So, with this PR, docker-compose decides the naming, and user can somewhat influence that using COMPOSE_PROJECT_NAME

andrerom:
@vidarl can you make this part of #205 also now that it passes?

vidarl:
ok, made this a part of #205

PR #208, opened 2017-09-01
Docker distribution images

vidarl:
Alternative way of building docker production images where ezplatform distribution files are reparated from the php image

Depends on #205


andrerom:
So this both depends on 205, and serves as an alternative to what is in it hence blocking it, right?

vidarl:
It is just one commit more, compared to 205. The reason why this is in separate PR is that this commit was made after 205 was created.
I can merge it into 205 if you like

The stack file in 205 and 208 solves two different usecases.
If you only have a few sites or every site has different code/customization, then stack example in 205 is the way to go. If you have many 100% identically (codewise ) sites, then 208 is the way to go

andrerom:
Unsure if there is any point in having several ways of doing stuff here, so lets deal with this later and focus on #205

vidarl:
This PR would be required by the demos.
If we don't want the code here, the code would need to live somewhere else; Possible, but not very convenient.

andrerom:
why is it requriment for demo? to be able to update php but not the code?
If so, let's make this the only documented way for "production use" then, we don't need doc for something we don't do.

However would maybe rename the cases to something like:
- Local using Docker Compose: Development "mount" use 
- Loca using Docker Compose: Demo image use
- Server using Docker Swarm: Demo image use

vidarl:
> If so, let's make this the only documented way for "production use" then, we don't need doc for something we don't do.

Sure about that? We only document the particular use-cases we ATM uses internally? and not what maybe most others would need ?

I also realize that this PR should be renamed as it doesn't use docker stack, only docker-compose.
This is because stack has no support for pods so you would then need to use a non-default volume driver in order to get  this working with docker stack on a multi node swarm cluster.

andrerom:
> Sure about that? We only document the particular use-cases we ATM uses internally? and not what maybe most others would need ?

most others? As of https://github.com/ezsystems/ezplatform/commit/7f2abb5acd555a0c70a8a5c77c04814be2e3af41 this stuff is not really meant for direct end developer use.

vidarl:
Closing this PR, it will instead be part of #205

PR #209, opened 2017-09-05
Enable roave/security-advisories

glye:
> Ready for review

Purpose: Prevent accidental require/update to known vulnerabilities.
See https://github.com/Roave/SecurityAdvisories which also explains why `dev-master` is used.

Manual test that a vulnerable package is refused:
```
$ composer require zendframework/zendframework:2.3.1
Your requirements could not be resolved to an installable set of packages.
  Problem 1
    - zendframework/zendframework 2.3.1 conflicts with roave/security-advisories[dev-master].
```

Manual test that Symfony's security checker does its thing:
```
$ php bin/security-checker security:check composer.lock
Symfony Security Check Report
=============================
 // Checked file: /home/gl/git/ezplatform/composer.lock
 [OK] No packages have known vulnerabilities.
 ! [NOTE] This checker can only detect vulnerabilities that are referenced in the SensioLabs security advisories
 !        database. Execute this command regularly to check the newly discovered vulnerabilities.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
         "symfony/expression-language": "~2.4",
         "sensio/framework-extra-bundle": "~3.0",
         "ezsystems/content-on-the-fly-prototype": "~0.1.7",
-        "willdurand/js-translation-bundle": "^2.6"
+        "willdurand/js-translation-bundle": "^2.6",
+        "roave/security-advisories": "dev-master"
     },
     "require-dev": {
         "behat/behat": "~3.0",


andrerom:
nitpick: don't add at the end of the array, it causes conflicts for merges up to 2.0, maybe at the top 

Also as we generate releases using dev, we could consider adding this to dev section so it does not interfere with updates on prod for end users. wdyt @yannickroger? You know the current maintenance process best here.

Note: Implied here is to update release process to always update this package to, and add packages  it conflicts with _(hence which has security updates)_ until it passes.

glye:
Moved up top. As for interfering with prod updates, this would only happen when someone either works only in prod, or if they insist on using a vulnerable version, right? Both of which we'd presumably recommend not doing. Anyway, I'll bow to superior knowledge here :)

andrerom:
@glye @bdunogier @yannickroger This merge implies this package should always also be updated when updating lock files for releases , and if it complains about security conflicts, then also update the packages it conflicts with. If this does not fit, like for instance on micro releases we'll use our best judgment to get releases out, and further improve the release process with the outcome.

ACK by reaction? :)

glye:
@andrerom Thanks for merging. I reckon this should also go into 2.0 (uncertain now..?) and ezpublish-platform (separate PR/commit?)

andrerom:
> I reckon this should also go into 2.0 (uncertain now..?)

it will on next rebase

> and ezpublish-community / ezpublish-platform (separate PR/commit?)

unsure if applicable, but yeah cherry pick to ezpublish-platform:5.4

PR #211, opened 2017-09-15
[WIP] EZEE-1710: Check if shm size is too low for platform-ui non-edge

mnocon:
### Do not merge

I'm checking if the tests that are run in PlatformUI repository can be fixed with specifying greater shm-size in selenium docker container.

If this build fails (as this one: https://travis-ci.org/ezsystems/PlatformUIBundle/jobs/275153347) with:
```
When I create a content of this Type                                                             
# EzSystems\PlatformUIBundle\Features\Context\Fields::createAContentOfThisType()
      WebDriver\Exception\UnknownError: unknown error: session deleted because of page crash
```
I will create another commit increasing the shm-size. If it passes the test command change will be reverted.


mnocon:
Well, my hunch was wrong - I'm closing this PR

PR #216, opened 2017-10-13
EZP-28126: Change default Varnish VCL to not cache user varying content in proxies or browsers

andrerom:
> Issue: https://jira.ez.no/browse/EZP-28126
> Epic: https://jira.ez.no/browse/EZP-27624

Improvement to our VCL in terms of not caching responses in Browser and shared proxy caches by default. And with inline suggestions for those that want to adapt it to be less strict or apply to all responses.

### DIFF
--- a/doc/varnish/vcl/varnish4_xkey.vcl
+++ b/doc/varnish/vcl/varnish4_xkey.vcl
@@ -69,6 +69,9 @@ sub vcl_recv {
     // Retrieve client user context hash and add it to the forwarded request.
     call ez_user_context_hash;
 
+    // Sort the query string for cache normalization.
+    set req.url = std.querysort(req.url);
+
     // If it passes all these tests, do a lookup anyway.
     return (hash);
 }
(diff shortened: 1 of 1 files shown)


doc/varnish/vcl/varnish4_xkey.vcl
```diff
@@ -203,16 +203,17 @@ sub vcl_deliver {
...
     // If we get here, this is a real response that gets sent to the client.
 
-    // Remove the vary on user context hash, this is nothing public. Keep all
-    // other vary headers.
-    set resp.http.Vary = regsub(resp.http.Vary, "(?i),? *X-User-Hash *", "");
-    set resp.http.Vary = regsub(resp.http.Vary, "^, *", "");
-    if (resp.http.Vary == "") {
-        unset resp.http.Vary;
-    }
+    // Remove the private vary on user context hash, and set caches varying to privat for end clients
```
bdunogier: Typo: `private`

andrerom:
> We agree that everything done in the vcl will only apply to the Response that is sent by varnish, do we ?

yes, so if this is to be solved overall we would need to solve for Symfony proxy as well before going ahead and call this solved :)

bdunogier:
> we would need to solve for Symfony proxy as well

Good point yes. This must be added to the JIRA issue. Can you do it if you still have its reference ? I can look it up otherwise.

andrerom:
done

andrerom:
@bdunogier Updated to allow some caching on non session based traffic in proxies by default, it's getting a bit complex but feedback from @kmadejski's review was that it was pretty clear and also clear how you can tune it for custom needs.

PR #218, opened 2017-10-19
[composer] Added oneup/flysystem-bundle dependency

webhdx:
# Description
Currently `2.0` cannot be installed due to the version constraint on `oneup/flysystem-bundle` in `ezsystems/ezpublish-kernel` which allows to install either `^1.0` or `^2.0`. Composer tends to install lower version which makes installation fail due to the BC break in `symfony/symfony` 3.4.0-BETA1. This PR forces `oneup/flysystem-bundle` installation in version 2.x.


andrerom:
Hmm.. Normally it does not prefer lower version, anyway, 👍  but would be good to know why.

PR #227, opened 2017-11-24
Add WhiteOctoberPagerfantaBundle

mikadamczyk:
Add WhiteOctoberPagerfantaBundle to implement pagination feature in new Admin UI

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -27,6 +27,7 @@ class AppKernel extends Kernel
             new Knp\Bundle\MenuBundle\KnpMenuBundle(),
             new Bazinga\Bundle\JsTranslationBundle\BazingaJsTranslationBundle(),
             new FOS\JsRoutingBundle\FOSJsRoutingBundle(),
+            new WhiteOctober\PagerfantaBundle\WhiteOctoberPagerfantaBundle(),
             // eZ Systems
             new EzSystems\PlatformHttpCacheBundle\EzSystemsPlatformHttpCacheBundle(),
             new eZ\Bundle\EzPublishCoreBundle\EzPublishCoreBundle(),
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,7 @@
         "willdurand/js-translation-bundle": "^2.6.4",
         "oneup/flysystem-bundle": "^2.0",
         "friendsofsymfony/jsrouting-bundle": "^1.4",
+        "white-october/pagerfanta-bundle": "^1.1",
         "roave/security-advisories": "dev-master"
     },
     "require-dev": {


composer.json
```diff
@@ -48,7 +48,8 @@
         "willdurand/js-translation-bundle": "^2.6.4",
         "oneup/flysystem-bundle": "^2.0",
         "friendsofsymfony/jsrouting-bundle": "^1.4",
-        "roave/security-advisories": "dev-master"
+        "roave/security-advisories": "dev-master",
+        "white-october/pagerfanta-bundle": "^1.1"
```
andrerom: could you add it before instead, simplifies rebases :)

app/AppKernel.php
```diff
@@ -37,10 +37,10 @@ public function registerBundles()
             new EzSystems\PlatformInstallerBundle\EzSystemsPlatformInstallerBundle(),
             new EzSystems\RepositoryFormsBundle\EzSystemsRepositoryFormsBundle(),
             new EzSystems\EzPlatformSolrSearchEngineBundle\EzSystemsEzPlatformSolrSearchEngineBundle(),
-            // new EzSystems\EzPlatformDesignEngineBundle\EzPlatformDesignEngineBundle(),
             new EzSystems\EzPlatformAdminUiBundle\EzPlatformAdminUiBundle(),
             new EzSystems\EzPlatformAdminUiModulesBundle\EzPlatformAdminUiModulesBundle(),
             new EzSystems\EzPlatformAdminUiAssetsBundle\EzPlatformAdminUiAssetsBundle(),
+            new WhiteOctober\PagerfantaBundle\WhiteOctoberPagerfantaBundle(),
```
adamwojs: IMHO this should be placed under "Dependencies" section

app/AppKernel.php
```diff
@@ -37,10 +37,10 @@ public function registerBundles()
             new EzSystems\PlatformInstallerBundle\EzSystemsPlatformInstallerBundle(),
             new EzSystems\RepositoryFormsBundle\EzSystemsRepositoryFormsBundle(),
             new EzSystems\EzPlatformSolrSearchEngineBundle\EzSystemsEzPlatformSolrSearchEngineBundle(),
-            // new EzSystems\EzPlatformDesignEngineBundle\EzPlatformDesignEngineBundle(),
```
andrerom: maybe skip removing this, we will, just like this one, probably re add it anyway, so good to keep track of what was disabled / removed on the initial admin ui import ;)

composer.json
mikadamczyk: ok :)

mikadamczyk:
Fixed :)

PR #228, opened 2017-11-24
Added Fastly params in config

vidarl:
This PR makes it possible to set Fastly configuration via environment variables, prepares the semantic configuration and make it a minimal effort for users to configure it once the fastly bundle has been installed

It also renames app/config/env/docker.php → app/config/env/generic.php as the file contains no docker specific stuff.

It might be a discussion if we should put this in platform or platform-ee as fastly will anyway not be available in platform. But recommendation is already partly here anyway ( I added the missing parts too ).

### DIFF
--- a/.env
+++ b/.env
@@ -29,6 +29,12 @@ WEB_HOST=web
 SELENIUM_HOST=selenium
 
 # Enable recommendations by setting valid id, key and uri
+# In order for this to work you also need to have the EzSystemsRecommendationBundle installed
 #RECOMMENDATIONS_CUSTOMER_ID=""
 #RECOMMENDATIONS_LICENSE_KEY=""
 #PUBLIC_SERVER_URI=""
+
+# Enable fastly by setting valid service_id and key
+# In order for this to work you also need to have the EzSystemsPlatformFastlyCacheBundle installed
+#FASTLY_SERVICE_ID=""
+#FASTLY_KEY=""
--- a/.env
+++ b/.env
@@ -35,6 +35,6 @@ SELENIUM_HOST=selenium
 #PUBLIC_SERVER_URI=""
 
 # Enable fastly by setting valid service_id and key
-# In order for this to work you also need to have the EzSystemsPlatformFastlyCacheBundle installed
+# In order for this to work you also need to have the EE bundle EzSystemsPlatformFastlyCacheBundle installed
 #FASTLY_SERVICE_ID=""
 #FASTLY_KEY=""
--- a/app/config/env/generic.php
+++ b/app/config/env/generic.php
@@ -107,19 +107,6 @@ if ($value = getenv('HTTPCACHE_DEFAULT_TTL')) {
     $container->setParameter('httpcache_default_ttl', $value);
 }
 
-// EzSystemsRecommendationsBundle settings
-if ($value = getenv('RECOMMENDATIONS_CUSTOMER_ID')) {
-    $container->setParameter('ez_recommendation.default.yoochoose.customer_id', $value);
-}
-
-if ($value = getenv('RECOMMENDATIONS_LICENSE_KEY')) {
-    $container->setParameter('ez_recommendation.default.yoochoose.license_key', $value);
-}
-
-if ($value = getenv('PUBLIC_SERVER_URI')) {
-    $container->setParameter('ez_recommendation.default.server_uri', $value);
-}
-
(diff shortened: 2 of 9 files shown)


app/config/ezplatform.yml
```diff
@@ -37,6 +37,6 @@ ezpublish:
               # As we by default enable EzSystemsPlatformHttpCacheBundle which is designed to expire all affected cache
               # on changes, and as error / redirects now have separate ttl, we easier allow ttl to be greatly increased
               default_ttl: '%httpcache_default_ttl%'
-            # HttpCache purge server(s) setting, eg Varnish, for when ezpublish.http_cache.purge_type is set to 'http'.
             http_cache:
+                # HttpCache purge server(s) setting, eg Varnish, for when ezpublish.http_cache.purge_type is set to 'varnish'.
                 purge_servers: ['%purge_server%']
```
vidarl: I am considering also adding the following after this line:

```
                # Enable the following line if you have installed the EzSystemsPlatformFastlyCacheBundle bundle
                # fastly:
                #     service_id: '%fastly_service_id%'
                #     key: '%fastly_key%'
```

.env
```diff
@@ -29,6 +29,12 @@ WEB_HOST=web
 SELENIUM_HOST=selenium
 
 # Enable recommendations by setting valid id, key and uri
+# In order for this to work you also need to have the EzSystemsRecommendationBundle installed
 #RECOMMENDATIONS_CUSTOMER_ID=""
 #RECOMMENDATIONS_LICENSE_KEY=""
 #PUBLIC_SERVER_URI=""
+
+# Enable fastly by setting valid service_id and key
+# In order for this to work you also need to have the EzSystemsPlatformFastlyCacheBundle installed
```
andrerom: As it's not apparent by the name, maybe mention it is a EE bundle here. Unlike Reco, it's not public, so no way to look it up and discover you need a license for it on github / packagist.

app/config/env/generic.php
```diff
@@ -90,12 +90,41 @@
...
     $container->setParameter('purge_server', $purgeServer);
 }
 
 if ($value = getenv('HTTPCACHE_DEFAULT_TTL')) {
     $container->setParameter('httpcache_default_ttl', $value);
 }
+
+// EzSystemsRecommendationsBundle settings
+if ($value = getenv('RECOMMENDATIONS_CUSTOMER_ID')) {
+    $container->setParameter('ez_recommendation.default.yoochoose.customer_id', $value);
```
adamwojs: Nitpick: EzSystemsRecommendationsBundle parameters are out of commit/PR scope. Could you please move those changes to separate commit?

bdunogier:
> It might be a discussion if we should put this in platform or platform-ee as fastly will anyway not be available in platform

We could clarify that by specifying in the file that Fastly support can be obtained with the Enterprise Edition of eZ Platform.

app/config/env/generic.php
bdunogier: An important nitpick :)

app/config/ezplatform.yml
bdunogier: Is it really needed ? Uncommenting these lines shouldn't have direct benefits. It uses default settings that come from the built-in parameters.yml.

app/config/env/generic.php
vidarl: Moved to separate commit ( need to rebase in order to get ridd of the fixup!s though)

.env
vidarl: Fixed

vidarl:
> Unsure about having this in ezsystems/ezplatform as well. Since it requires EE, isn't it easier to have it over there ?
Yeah, maybe it is.. I added here since reco bundle was done like this ( partly at least )...

But if you prefer, I might just rename docker.php -> generic.php here in ezplatform, and do the rest in a platform-ee PR

andrerom:
> But if you prefer, I might just rename docker.php -> generic.php here (+ the RECOMENDATION fix) in ezplatform, and do the rest in a platform-ee PR

maybe that is cleaner indeed, then my comment on clarifying it is for ee only goes somewhat away.

vidarl:
Ok. Closing this PR and creating two new onces...

vidarl:
New PRs:
platform : https://github.com/ezsystems/ezplatform/pull/231
platform-ee : https://github.com/ezsystems/ezplatform-ee/pull/68

PR #229, opened 2017-11-26
EZP-28183: Always refresh stale cache for logged in users

andrerom:
> Issue: https://jira.ez.no/browse/EZP-28183

When editing Landing pages @mnocon are getting stale content when viewing it afterwards, which I think is because we on purpose allowed that here for view, only exception was REST routes when used from platform ui. So  given we need to ship with sensible defaults, making sure users with sessions sees the latests changes all over seems like a safer choice* _(editors, forum users, admins, shoppers, ..)_.

\* _Only exception kept here is if backend is not healthy, which should not happen here as we have not configured a [probe](http://varnish-cache.org/trac/wiki/BackendPolling) but left that up to the integrator._

### DIFF
--- a/doc/varnish/vcl/varnish4_xkey.vcl
+++ b/doc/varnish/vcl/varnish4_xkey.vcl
@@ -79,18 +79,18 @@ sub vcl_recv {
 // Called when a cache lookup is successful. The object being hit may be stale: It can have a zero or negative ttl with only grace or keep time left.
 sub vcl_hit {
    if (obj.ttl >= 0s) {
-       // A pure unadultered hit, deliver it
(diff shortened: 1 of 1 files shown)


adamwojs:
@andrerom Change is quite simple, but could you please provide more information about issue which you refer to ? Ticket from PR description is very generic :wink:

andrerom:
@adamwojs added :)

andrerom:
@Plopix FYI on launchpad vcl

Plopix:
got it @andrerom thx!
https://github.com/ezsystems/launchpad/commit/9882513c68274f611c603e8aa968933fe276f292

PR #231, opened 2017-11-28
EZP-28336 : app/config/env/docker.php is actually non-docker specific

vidarl:
This PR does the following:
- renames app/config/env/docker.php → app/config/env/generic.php as the file contains no docker specific stuff.
- adds possibility to set purge_type via environment variable HTTPCACHE_PURGE_TYPE
- adds possibility to set recommendation settings via environment variables

Note:
When merging this to ezplatform-ee-demo, please note that the recommencation settings are already in docker.php (generic.php) ... Make sure that code is not duplicated, and that is is located  at the same spot in the file to avoid further merge conflicts.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -2,7 +2,7 @@ imports:
     - { resource: default_parameters.yml }
     - { resource: parameters.yml }
     - { resource: security.yml }
-    - { resource: env/docker.php }
+    - { resource: env/generic.php }
     - { resource: env/platformsh.php }
 
 # Configuration for Database connection, can have several connections used by eZ Repositories in ezplatform.yml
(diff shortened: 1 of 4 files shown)


andrerom:
Missing issue and such, but +1

vidarl:
Added jira issue : https://jira.ez.no/browse/EZP-28336 as requested

vidarl:
so, do we have a 1.13 branch which this one can be merged to ?

vidarl:
well, there is no BC so I guess it may also be merged to master

andrerom:
This should go into 1.13 if there is a branch for that, it will then go into 1.13 -> master -> 2.0 _(if still separate branch)_

PR #233, opened 2017-12-07
Refactor param names to keep BC with eZ Platform 1.x while still allowing fallback to env vars

emodric:
This changes param names back to "human readable" format to keep BC with eZ Platform 1.x, while still allowing for fallback to environment variables or if environment variables do not exist, fallback to values specified with `'%env(ENV_NAME)%'` format.

This is the same what Sylius does with their parameters: https://github.com/Sylius/Sylius/blob/master/app/config/parameters.yml.dist

### DIFF
--- a/app/config/cache_pool/cache.memcached.yml
+++ b/app/config/cache_pool/cache.memcached.yml
@@ -7,4 +7,4 @@ services:
               clearer: cache.app_clearer
               # Example from vendor/symfony/symfony/src/Symfony/Component/Cache/Traits/MemcachedTrait.php:
               # memcached://user:pass@localhost?weight=33'
-              provider: 'memcached://%env(CACHE_DSN)%'
+              provider: 'memcached://%cache_dsn%'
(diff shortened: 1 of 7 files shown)


app/config/parameters.yml.dist
```diff
@@ -24,3 +24,13 @@ parameters:
 
     # Purge type used by HttpCache system
     purge_type: local
+
+    # Fallback values for when environment variables do not exist
+    env(SYMFONY_SECRET): ThisEzPlatformTokenIsNotSoSecretChangeIt
```
andrerom: this is kind of a problem, user will be prompted for variables twice here right?

emodric: Yep. We can leave these in here, and move the `secret` and `database_*` to `default_parameters.yml` maybe?

andrerom: yes

emodric: Okay. I'll try to get it done today.

emodric:
@andrerom How about now?

andrerom:
You can remove your changes in app/config/parameters.yml.dist now. Also; does it work? :P

emodric:
Removed the changes from `parameters.yml.dist`. As for does it work, can't test now, I'm in some meetings, but will let you know.

emodric:
@andrerom Tested, works fine!

Plopix:
Nice!

PR #234, opened 2017-12-07
Cleanup/Improve Platform.sh config for 1.13 (and up)

andrerom:
Suggestions for improving the default config to be more cluster aware, and try to reduce how much cache we put on shared disk as it is shared _(for now, plan is to move it to local disk once p.sh config allows us to later)_.

Todo:

- [x] Rebase on top of Crell's session PR and adjust for that and other feedback he has
- [x] Cleanup? _(including drop some of the stuff in the newly added platformsh.yml file for now)_


.platform.app.yaml
```diff
@@ -36,7 +45,6 @@ disk: 2048
 
 # The mounts that will be performed when the package is deployed.
 mounts:
-    "/app/cache": "shared:files/cache"
```
vidarl: We cannot remove this line I think.
Well, maybe for professional subscription cause that is atm a single host environment ( I believe platform.sh might introduce multi nodes in the biggest subscription plans shortly though)

On enterprise there will be no convenient way of clearing the cache unless this is shared among the nodes

.platform.app.yaml
```diff
@@ -64,7 +72,9 @@ hooks:
             php -d memory_limit=-1 app/console ezplatform:install --env=$SYMFONY_ENV $INSTALL_EZ_INSTALL_TYPE
             touch web/var/.platform.installed
         fi
-        app/console --env=$SYMFONY_ENV cache:clear
```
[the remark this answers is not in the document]
andrerom: Isn't deploy effectively a new node with empty cache?

My assumption was that, and that the cache clear was needed since we shared it on NFS.

.platform.app.yaml
andrerom: But this is an major anti pattern.

For eZ Publish 5.x and eZ Platform we clearly state app/cache is not meant to be shared among severs.
--

But there is an blocker here that I forgot about when opening this: The reason why this line is here is because Symfony needs to write to cache folder. And if this is removed it will not be writeable after build stage

This seems to have been fixed in Symfony 4, but does not seems to have been fixed in earlier versions.

.platform.app.yaml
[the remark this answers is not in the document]
vidarl: Nope, it is the same node every time. So you do for instance reuse the same composer cache during build....
But if don't declare it as a share, it should in practice be empty on deployment. But it will also be read-only.... If application want to write something, it must be in a shared mount...

AFAIK, with symfony 3.x, the cache might be committed to git as it has no DB info etc, right ?
but for 2.x, this must be a volume.

But also note that shared mounts are not NFS..... it is a shared block device something. I don't know the technical details. I have not performance tested it, but I assume it outperforms NFS

.platform.app.yaml
```diff
@@ -64,7 +72,9 @@ hooks:
             php -d memory_limit=-1 app/console ezplatform:install --env=$SYMFONY_ENV $INSTALL_EZ_INSTALL_TYPE
```
andrerom: This will fail now, as it requires write access to clear cache.

Might be that it just need to clear persistence cache, so that could be fixed.

andrerom: Fixed in: https://github.com/ezsystems/ezpublish-kernel/pull/2180

andrerom:
@vidarl you're right, app/cache needs to be writable, it's not safe to assume something else for now. Only Symfony 4 supports otherwise.

So instead made changes here to minimize amount of stuff is placed in the shared app/cache, minimizing the issue. In parallel will be looking with p.sh towards allowing local writable mount for use for caching so it's not shared but rather per host.

Also added session storage in redis here, but that might be something @glye already tested back during the spring and found issues with.

glye:
> already tested back during the spring and found issues with.

I don't think so. It's so long ago that it would need new testing now, anyway.

app/config/env/platformsh.php
```diff
@@ -47,5 +51,7 @@
 // Disable PHPStormPass
 $container->setParameter('ezdesign.phpstorm.enabled', false);
 
-// Store session into /tmp.
-ini_set('session.save_path', '/tmp/sessions');
+// Load platform.sh specifc settings
```
bdunogier: typo on "specific"

app/config/env/platformsh.yml
```diff
@@ -0,0 +1,16 @@
+# To make sure as littel as possible cache is written to the shared mount we use apc and redis (for cache that needs to be shared)
```
bdunogier: "as little cache as possible"

andrerom:
@SylvainGuittard Looks ok? And I assume when platformsh job here says it's green the config works right?

SylvainGuittard:
@andrerom If it's green it's a good sign regarding the deploy script. And it is. Here is the output:
```
Found 2 new commits.

Building application 'app' (runtime type: php:7.1, tree: a417c19)
  Generating runtime configuration.
  
  Executing build hook...
    W: Loading composer repositories with package information
    W: Updating dependencies
    W: Package operations: 79 installs, 0 updates, 0 removals
    W:   - Installing ocramius/package-versions (1.2.0): Loading from cache
    W:   - Installing symfony/polyfill-mbstring (v1.6.0): Loading from cache
    W:   - Installing twig/twig (v2.4.4): Loading from cache
    W:   - Installing symfony/security-acl (v3.0.0): Loading from cache
    W:   - Installing symfony/symfony (v2.8.32): Loading from cache
    W:   - Installing symfony/polyfill-util (v1.6.0): Loading from cache
    W:   - Installing paragonie/random_compat (v2.0.11): Loading from cache
    W:   - Installing symfony/polyfill-php70 (v1.6.0): Loading from cache
    W:   - Installing symfony/polyfill-php56 (v1.6.0): Loading from cache
    W:   - Installing ircmaxell/password-compat (v1.0.4): Loading from cache
    W:   - Installing symfony/polyfill-php55 (v1.6.0): Loading from cache
    W:   - Installing symfony/polyfill-php54 (v1.6.0): Loading from cache
    W:   - Installing psr/log (1.0.2): Loading from cache
    W:   - Installing symfony/polyfill-intl-icu (v1.6.0): Loading from cache
    W:   - Installing doctrine/lexer (v1.0.1): Loading from cache
    W:   - Installing doctrine/inflector (v1.2.0): Loading from cache
    W:   - Installing doctrine/collections (v1.5.0): Loading from cache
    W:   - Installing doctrine/cache (v1.7.1): Loading from cache
    W:   - Installing doctrine/annotations (v1.6.0): Loading from cache
    W:   - Installing doctrine/common (v2.8.1): Loading from cache
    W:   - Installing symfony/polyfill-apcu (v1.6.0): Loading from cache
    W:   - Installing psr/cache (1.0.1): Loading from cache
    W:   - Installing guzzle/guzzle (v3.9.3): Loading from cache
    W:   - Installing friendsofsymfony/http-cache (1.4.4): Loading from cache
    W:   - Installing friendsofsymfony/http-cache-bundle (1.3.12): Loading from cache
    W:   - Installing zetacomponents/base (1.9.1): Loading from cache
    W:   - Installing zetacomponents/mail (1.8.3): Loading from cache
    W:   - Installing tedivm/stash (v0.14.2): Loading from cache
    W:   - Installing tedivm/stash-bundle (v0.6.2): Loading from cache
    W:   - Installing symfony-cmf/routing (1.4.1): Loading from cache
    W:   - Installing sensio/framework-extra-bundle (v3.0.28): Loading from cache
    W:   - Installing composer/ca-bundle (1.1.0): Loading from cache
    W:   - Installing sensiolabs/security-checker (v4.1.6): Loading from cache
    W:   - Installing sensio/distribution-bundle (v5.0.21): Loading from cache
    W:   - Installing qafoo/rmf (1.0.0): Loading from cache
    W:   - Installing pagerfanta/pagerfanta (v1.0.5): Loading from cache
    W:   - Installing league/flysystem (1.0.41): Loading from cache
    W:   - Installing oneup/flysystem-bundle (2.0.1): Loading from cache
    W:   - Installing zendframework/zend-eventmanager (3.2.0): Loading from cache
    W:   - Installing zendframework/zend-code (3.3.0): Loading from cache
    W:   - Installing ocramius/proxy-manager (2.1.1): Loading from cache
    W:   - Installing nelmio/cors-bundle (1.5.4): Loading from cache
    W:   - Installing imagine/imagine (v0.7.1): Loading from cache
    W:   - Installing liip/imagine-bundle (1.9.1): Loading from cache
    W:   - Installing kriswallsmith/buzz (v0.15.2): Loading from cache
    W:   - Installing nikic/php-parser (v3.1.2): Loading from cache
    W:   - Installing jms/translation-bundle (1.3.2): Loading from cache
    W:   - Installing hautelook/templated-uri-router (2.0.2): Loading from cache
    W:   - Installing hautelook/templated-uri-bundle (2.0.1): Loading from cache
    W:   - Installing doctrine/doctrine-cache-bundle (1.3.2): Loading from cache
    W:   - Installing jdorn/sql-formatter (v1.2.17): Loading from cache
    W:   - Installing doctrine/dbal (v2.6.3): Loading from cache
    W:   - Installing doctrine/doctrine-bundle (1.8.1): Loading from cache
    W:   - Installing ezsystems/ezpublish-kernel (dev-master cfe8d73): Loading from cache
    W:   - Installing ezsystems/ezplatform-http-cache (dev-master aede123): Loading from cache
    W:   - Installing ezsystems/ezplatform-design-engine (v1.1.0): Loading from cache
    W:   - Installing willdurand/js-translation-bundle (2.6.5): Loading from cache
    W:   - Installing zetacomponents/system-information (1.1.1): Loading from cache
    W:   - Installing ezsystems/ez-support-tools (v0.1.3): Loading from cache
    W:   - Installing ezsystems/repository-forms (1.10.x-dev b6b0a30): Loading from cache
    W:   - Installing ezsystems/platform-ui-assets-bundle (v4.0.0): Loading from cache
    W:   - Installing ezsystems/platform-ui-bundle (dev-master f2bda5b): Loading from cache
    W:   - Installing ezsystems/content-on-the-fly-prototype (v0.1.11): Loading from cache
    W:   - Installing twig/extensions (v1.5.1): Loading from cache
    W:   - Installing kriswallsmith/assetic (v1.4.0): Loading from cache
    W:   - Installing symfony/assetic-bundle (v2.8.2): Loading from cache
    W:   - Installing swiftmailer/swiftmailer (v5.4.8): Loading from cache
    W:   - Installing symfony/swiftmailer-bundle (v2.6.7): Loading from cache
    W:   - Installing monolog/monolog (1.23.0): Loading from cache
    W:   - Installing symfony/monolog-bundle (v3.1.2): Loading from cache
    W:   - Installing sensio/generator-bundle (v3.1.7): Loading from cache
    W:   - Installing incenteev/composer-parameter-handler (v2.1.2): Loading from cache
    W:   - Installing netgen/query-translator (1.0.2): Loading from cache
    W:   - Installing ezsystems/ezplatform-solr-search-engine (dev-master 8fe61c2): Loading from cache
    W:   - Installing ezsystems/ezplatform-multi-file-upload (v0.1.3): Loading from cache
    W:   - Installing ezplatform-i18n/ezplatform-i18n-ach_ug (v1.4.0): Loading from cache
    W:   - Installing egulias/listeners-debug-command-bundle (1.9.1): Loading from cache
    W:   - Installing white-october/pagerfanta-bundle (v1.0.8): Loading from cache
    W: Package guzzle/guzzle is abandoned, you should avoid using it. Use guzzlehttp/guzzle instead.
    W: Writing lock file
    W: Generating optimized autoload files
    ocramius/package-versions:  Generating version class...
    ocramius/package-versions: ...done generating version class
    W: > Incenteev\ParameterHandler\ScriptHandler::buildParameters
    Creating the "app/config/parameters.yml" file
    W: > Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::buildBootstrap
    W: > eZ\Bundle\EzPublishCoreBundle\Composer\ScriptHandler::clearCache
    
     // Clearing the cache for the prod environment with debug false                                                        
    
     [OK] Cache for the "prod" environment (debug=false) was successfully cleared.                                          
    
    W: > Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::installAssets
    
     Trying to install assets as relative symbolic links.
    
     --- -------------------------------- ------------------ 
          Bundle                           Method / Error    
     --- -------------------------------- ------------------ 
      ✔   FrameworkBundle                  relative symlink  
      ✔   EzSystemsMultiFileUploadBundle   relative symlink  
      ✔   eZPlatformUIAssetsBundle         relative symlink  
      ✔   eZPlatformUIBundle               relative symlink  
      ✔   WhiteOctoberPagerfantaBundle     relative symlink  
      ✔   ContentOnTheFlyBundle            relative symlink  
      ✔   BazingaJsTranslationBundle       relative symlink  
      ✔   JMSTranslationBundle             relative symlink  
     --- -------------------------------- ------------------ 
    
     [OK] All assets were successfully installed.                                                                           
    
    W: > eZ\Bundle\EzPublishCoreBundle\Composer\ScriptHandler::dumpAssets
    Dumping all prod assets.
    Debug mode is off.
    
    11:53:36 [dir+] web/css
    11:53:36 [file+] web/css/64dd8c5.css
    11:53:36 [dir+] web/js
    11:53:36 [file+] web/js/d0ac14a.js
    W: > Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::installRequirementsFile
  
  Executing pre-flight checks...

  Compressing application.
  Beaming package to its final destination.

Provisioning certificates:
  Reusing existing certificates.


Re-deploying environment figyhb64ptrpy-pr-234-vpaqoby.
  Environment configuration:
    app (type: php:7.1, size: S, disk: 2048)
    mysqldb (type: mysql:10.1, size: S, disk: 2048)
    rediscache (type: redis:3.2, size: S)
```

When you click on the link, the server returns an internal server error 500.
I had a look at the Symfony logs:
```
[2017-12-08 12:00:23] request.CRITICAL: Uncaught PHP Exception RedisException: "Redis server went away" at /app/vendor/tedivm/stash/src/Stash/Driver/Redis.php line 270 {"exception":"[object] (RedisException(code: 0): Redis server went away at /app/vendor/tedivm/stash/src/Stash/Driver/Redis.php:270)"} []
[2017-12-08 12:00:23] request.CRITICAL: Exception thrown when handling an exception (InvalidArgumentException: The directory "/app/app/cache/prod/annotations" is not writable. at /app/vendor/doctrine/cache/lib/Doctrine/Common/Cache/FileCache.php line 92) {"exception":"[object] (InvalidArgumentException(code: 0): The directory \"/app/app/cache/prod/annotations\" is not writable. at /app/vendor/doctrine/cache/lib/Doctrine/Common/Cache/FileCache.php:92)"} []
[2017-12-08 12:00:23] request.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "The directory "/app/app/cache/prod/annotations" is not writable." at /app/vendor/doctrine/cache/lib/Doctrine/Common/Cache/FileCache.php line 92 {"exception":"[object] (InvalidArgumentException(code: 0): The directory \"/app/app/cache/prod/annotations\" is not writable. at /app/vendor/doctrine/cache/lib/Doctrine/Common/Cache/FileCache.php:92, RedisException(code: 0): Redis server went away at /app/vendor/tedivm/stash/src/Stash/Driver/Redis.php:270)"} []
```
So, it seems that Redis is not happy and I don't understand why `annotations` is not writable because we have `"/app/cache": "shared:files/cache"`...

nmeirik:
We are running netmaking.no on Platform.sh and are in need of search (for instance Solr). If needed, I'd be happy to do som testing when and if needed.

.platform.app.yaml
```diff
@@ -19,6 +19,21 @@ build:
...
+    # Set in php variables below
+    sessionstorage: "redisdata:redis"
+    #solr: "solr:endpoint0"
+
+
+variables:
+    php:
+        "display_errors": "Off"
+        session.save_handler: redis
+        session.save_path: "tcp://sessionstorage.internal:6379"
```
andrerom: todo, this needs to be kept in app/config/env/platformsh.php so it can be properly extracted from the relationships data which sometimes will differ from deploy to deploy.

.platform/services.yaml
```diff
@@ -16,3 +16,15 @@ rediscache:
...
+# Example for setting up Solr, place solr config in .platform/solr/config/solr (NB! remember to set autoSoftCommit)
+#solr:
+#    type: solr:6.6
+#    disk: 1024
+#    configuration:
+#        cores:
+#            core1:
+#                conf_dir: !archive "solr/config/solr"
+#        endpoints:
+#            endpoint0:
```
Crell: Is `endpoint0` a significant name in eZ?  If not, something like "main" seems like it would be more user-friendly.

.platform.app.yaml
```diff
@@ -40,46 +52,52 @@ disk: 2048
 
 # The mounts that will be performed when the package is deployed.
 mounts:
+    # NOTE: Once possible, cache folder will rather be configured as local writable mount
```
Crell: This comment is PE-specific, and this configuration applies to PS.  It can be removed.

.platform.app.yaml
```diff
@@ -40,46 +52,52 @@ disk: 2048
...
-            php -d memory_limit=-1 app/console ezplatform:install --env=$SYMFONY_ENV $INSTALL_EZ_INSTALL_TYPE
+            php -d memory_limit=-1 app/console ezplatform:install $INSTALL_EZ_INSTALL_TYPE
             touch web/var/.platform.installed
         fi
-        app/console --env=$SYMFONY_ENV cache:clear
+        # As we are deploying code changes we need to clear any kind of code/system cache
+        app/console cache:clear
+        # If this is not a "fast-deploy" then typically you would want to migrate your data as well here
+        #app/console doctrine:migrations:migrate --no-interaction --allow-no-migration
+        #app/console kaliop:migration:migrate --no-interaction --no-debug
```
Crell: Remember, this is on PS...  I don't know what you mean by "fast-deploy" here.  (It was an older feature on PE that isn't really a thing on Wings now, IIRC.)

Something to consider: Move the entire deploy hook to a `deploy.sh` file, which we can then reference from the PE setup as well.  That way users can still customize their deploy hook without having to notify the Platform.sh support team.  The commented out lines here suggest that you intend for users to be editing these possibly often, in which case that would help eliminate a step.

.platform/routes.yaml
```diff
@@ -4,6 +4,7 @@
     cache:
         # As this does not support Vary, and purging, we can't use this as Sf Proxy drop in.
         # However it is possible to enable this for anonymous traffic when backend sends expiry headers.
+        # Note: Recommended option for Enterprise use is to use ezplatform-http-cache-fastly bundle
```
Crell: Grammar suggestion: "eZ Platform Enterprise users are encouraged to use Fastly and enable the ezplatform-http-cache-fastly bundle."

.platform/services.yaml
```diff
@@ -1,5 +1,5 @@
 mysqldb:
-    type: mysql:10.0
+    type: mysql:10.1
```
Crell: You can go all the way to 10.2 if you want. :smile:

app/config/config_prod.yml
```diff
@@ -2,15 +2,25 @@ imports:
     - { resource: config.yml }
     - { resource: ezplatform_prod.yml }
 
+# If you are on platform.sh/docker or similar where app/cache is not writable during  runtime
```
Crell: Please capitalize Platform.sh

.platform/services.yaml
andrerom: We have reports on some issues with 10.2 for those still using eZ Publish legacy with this, so this was on purpose not bumped up to that yet. Also we have no tests, automated or manual running on it yet.

.platform.app.yaml
andrerom: didn't you or someone mention there will be a new mount format for PS (gen2) which also allows to specify a mount as local, which hopefully eventually will be supported by PE (wings) as well so the fomat can stay the same between the two?

.platform.app.yaml
andrerom: >  Move the entire deploy hook to a deploy.sh file, 

good suggestion.
Same for build?

>  I don't know what you mean by "fast-deploy" here

from what I been told there is a PS (gen2) option for this involving putting incomming requests on hold during deploy hook time.

.platform/services.yaml
andrerom: it's also used in default config: https://github.com/ezsystems/ezplatform/blob/master/app/config/config.yml#L36

.platform/routes.yaml
andrerom: I actually picked that wording specifically not to make distinction between eZ Platform Enterprise and Platform.sh Enterprise here, given you'll need both in order to be able to take advantage of it.

And once you can, it's strongly recommended, and not just encouraged. As we only support usage of Varnish/Fastly as a shared proxy on cluster installs, we don't support the file based Symfony proxy in this scenario, which would involve sharing cache over nfs, and niether does Symfony _(they recommend Varnish, and discourages use of nfs for cache, including HttpCache)_.

.platform/services.yaml
Crell: If it works it works; I just prefer less machine-y names in a user-visible file most of the time is all.

.platform.app.yaml
Crell: The build hook is controlled from the `.platform.app.yaml` file regardless of PS or PE, so there's no need to split it off.  You could if you want to but there's no compelling reason to.

We already put incoming requests on hold during deploy on PS.  On PE the VMs just go offline entirely for a moment, but with a CDN in front of them 99.9% of users won't notice.  There was a flag on PE called `fast-deploy` that I think skipped the deploy hook, or reopened the site before the deploy hook finished, or something, but it doesn't apply anymore in Wings so you can/should ignore its existence.

Just assume that the deploy hook will always run, the same, and block the site coming back up regardless of whether you're on PS or PE, so keep it short.

.platform.app.yaml
Crell: There is a new mount format for PS only that's already been deployed. With Wings, PE can be set to pull its mount information from the `.platform.app.yaml` file iff:

* They're in the old-style format
* No mounts on PE need to be `local`, they can all be shared by all webheads.

Since point 2 is not true for eZ, we cannot rely on that and the mount section here will be for PS only and will be ignored on PE.  I'm working on internal documentation for us on how to setup the PE mounts for eZ by default.  There's no need to reference that here.

.platform/routes.yaml
Crell: Well, the current text isn't a valid grammatical sentence in English so we should tweak it regardless. 😄 

```
We strongly recommend that users of eZ Platform Enterprise on Platform.sh Enterprise use the ezplatform-http-cache-fastly bundle and set this value to false.
See: *link to the docs for how to set that up*
```

Since eZ Platform on PS and eZ Platform Community on PE are both still legit things for people to do, even if we want them to go double-Enterprise.

.platform/services.yaml
Crell: Ah, no 1.0.2 then.  👍

app/config/env/platformsh.yml
```diff
@@ -0,0 +1,19 @@
+# To make sure as little cache as possible is written to the shared mount we use apc and redis (for cache that needs to be shared)
+
+imports:
+    - { resource: ../cache_pool/singleredis.yml }
```
andrerom: We can skip this one now I guess, as it's loaded in platformsh.php, cc @vidarl

.platform.app.yaml
andrerom: We can remove the comment for now, but we need to have one config that works on both and works good (shows best practice). Currently that is impossible since Wings does not support new things coming in PS, is there a chance this is solved by fall?

PR #238, opened 2017-12-13
[Travis] Use .env file for picking install

andrerom:
Also `[Travis] For now disable testing on varnish`.


andrerom:
@alongosz Merging this for now, lets sync up tomorrow.

NOTE: The Varnish error might also be because of the new travis images: https://docs.travis-ci.com/user/build-environment-updates/2017-12-12/

PR #239, opened 2017-12-14
[Docker/Travis] Retry composer install on issues & update docker-compose

andrerom:
Also see if this might fix varnish segmentation fault also.

- [ ] Varnish still gives segmentation fault on the new travis images

*NB: This is Stuck! Ideally should report this to [travis](https://github.com/travis-ci/travis-ci/issues/8870), however I have unsuccessfully been stuck at segmentation fault before so any kind of report will need a reproducible case for them to look into it...*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ branches:
     - master
     - /^\d.\d+$/
 
+# Update Docker and Docker Compose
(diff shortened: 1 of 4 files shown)


andrerom:
Reported: https://github.com/travis-ci/travis-ci/issues/8914

PR #242, opened 2017-12-19
fix: Remove hardcoded session name, it's obsolete.

Nattfarinn:


lserwatka:
You need to change a target branch on 2.0

PR #244, opened 2018-01-04
Fix Platform Admin UI assets package version

emodric:
^0.5 does not install anything newer than 0.5.99


emodric:
@andrerom Ideally, this requirement should be moved to Admin UI repo, right?

andrerom:
probably, but I'm not aware why it is like ti is so I'll leave it for @lserwatka to answer on if we can do that now that package is stable.

PR #246, opened 2018-01-04
[WIP] Fixed Dockerfiles to work with Symfony 3

vidarl:
Some minor changes forgotten in 8065c6faf4d7615f413f014b9ac86efb3ffb7cd1

### DIFF
--- a/doc/docker/Dockerfile-distribution
+++ b/doc/docker/Dockerfile-distribution
@@ -12,8 +12,8 @@ COPY --from=builder /var/www /var/www
 WORKDIR /var/www
 
 # Fix permissions for www-data
-RUN chown -R www-data:www-data app/cache app/logs \
-    && find app/cache app/logs -type d -print0 | xargs -0 chmod -R 775 \
-    && find app/cache app/logs -type f -print0 | xargs -0 chmod -R 664
+RUN chown -R www-data:www-data var/cache var/logs \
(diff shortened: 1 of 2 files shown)


vidarl:
FYI : I marked this as [WIP]  as I would like to do final testing on ezplatform-demo v2.0 once that is ready for testing

doc/docker/Dockerfile-distribution
```diff
@@ -12,8 +12,8 @@ COPY --from=builder /var/www /var/www
 WORKDIR /var/www
 
 # Fix permissions for www-data
-RUN chown -R www-data:www-data app/cache app/logs \
-    && find app/cache app/logs -type d -print0 | xargs -0 chmod -R 775 \
-    && find app/cache app/logs -type f -print0 | xargs -0 chmod -R 664
+RUN chown -R www-data:www-data var/cache var/logs \
```
andrerom: should probably align this and case above with doc and do this for var and web/var

Cause there might be var/session(s) in some cases, and binary files are still placed in web/var.

andrerom:
@vidarl @SylvainGuittard Any update on the testing and progress of demo? Aka if stabilizing right now, then we can bundle it with 2.0.1, otherwise we can do a 2.0.2 once demo is ready. /cc @lserwatka

PR #248, opened 2018-01-09
Fix SYMFONY_TMP_DIR when defined

Plopix:
dirname of /tmp would return /, native dirname(__DIR__) function return the parent of app which is good. But not in the SYMFONY_TMP_DIR case.


andrerom:
@Plopix could you rebase for 2.0?

Plopix:
done @andrerom ;)

PR #251, opened 2018-01-09
Add gzip encoding by default

peterkeung:
Page speed best practices means we should add gzip encoding by default

- [ ] nginx conf
-  ~varnish vcl~ Should work by default: https://varnish-cache.org/docs/trunk/users-guide/compression.html

### DIFF
--- a/doc/apache2/vhost.template
+++ b/doc/apache2/vhost.template
@@ -114,4 +114,11 @@
             ExpiresDefault "now plus 10 years"
         </LocationMatch>
     </IfModule>
+
+    # Enable gzip encoding
+    <IfModule mod_deflate.c>
+        AddOutputFilterByType DEFLATE text/css application/x-javascript application/javascript text/javascript text/plain text/xml application/xml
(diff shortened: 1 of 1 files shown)


andrerom:
+1, but we should add to nginx as well if possible

doc/apache2/vhost.template
```diff
@@ -114,4 +114,9 @@
             ExpiresDefault "now plus 10 years"
         </LocationMatch>
     </IfModule>
+
+    # Enable gzip encoding
+    <IfModule mod_deflate.c>
+        AddOutputFilterByType DEFLATE text/css application/x-javascript application/javascript text/javascript text/plain text/html text/xml application/xml
```
andrerom: Besides lack of config for nginx, should also omit html here as it should not be used in combination with https. _(ref: BREACH/CRIME)_

peterkeung: Thanks. I've broken this out with a comment.

andrerom:
Closed in favour of https://github.com/ezsystems/ezplatform/pull/375

PR #257, opened 2018-01-29
[WIP][Docker] Add SSL proxy for being able to run using http/2

andrerom:
In order to be able to do http/2 and server push testing with docker containers:
- Adds a ssl proxy using nginx _(and perl module to be able to set public port in env variable)_
- Adjusts Varnish vcl to be able to pass in port number from trusted proxy
- Adjust app.php in oder to correctly translate the hostnames used by docker to actually trusted proxy ip's

Known issues:
- Default self signed cert will show error in browser, there is no easy way around that, once proceeded it will be hidden until container is restarted. _However does support mounting own cert_

Todo:
- [ ] Wait for incomming nginx release with server push support
- [ ] Check with @vidarl where this is best placed, can also be separate yml file that would need to be appended to varnish or similar.

### DIFF
--- a/doc/docker/Dockerfile-nginx-ssl-proxy
+++ b/doc/docker/Dockerfile-nginx-ssl-proxy
@@ -1,0 +1,17 @@
+FROM nginx:stable
+# Based on: https://github.com/clamorisse/nginx-ssl-container
+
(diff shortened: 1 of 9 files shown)


andrerom:
On the part on configuring and matching proxies in VCL, see also: https://github.com/ezsystems/ezplatform/pull/93

PR #261, opened 2018-02-08
Fix indentation for the default_ttl config var

crevillo:
An easy one. Indetation is 4 spaces for all the file except for this part i'm changing now. Just in order to follow the same rule for everything.

ping @andrerom

### DIFF
--- a/app/config/ezplatform.yml
+++ b/app/config/ezplatform.yml
@@ -34,9 +34,9 @@ ezpublish:
             # all eng-GB data to other locales first.
             languages: [eng-GB]
             content:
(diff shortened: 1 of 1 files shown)


andrerom:
Does it also affect 1.13?

crevillo:
yes, seems so. https://github.com/ezsystems/ezplatform/blob/1.13/app/config/ezplatform.yml#L39

andrerom:
ok, then please rebase this for 1.13 and change base branch here, and we'll take care about conflicts when merging up.

crevillo:
Done. let me know if i did it ok. no conflicts here...

PR #262, opened 2018-02-09
EZP-28800 : Make it possible to store session in redis when using platform.sh

vidarl:
This one depends on https://github.com/ezsystems/ezpublish-kernel/pull/2245

See the [Jira ticket](https://jira.ez.no/browse/EZP-28800) for further info

### DIFF
--- a/app/config/env/platformsh.php
+++ b/app/config/env/platformsh.php
@@ -55,8 +55,8 @@ if (isset($relationships['redissession'])) {
             continue;
         }
 
-        ini_set('session.save_handler', 'redis');
-        ini_set('session.save_path', sprintf('%s:%d', $endpoint['host'], $endpoint['port']));
+        $container->setParameter('ezplatform.session.handler_id', 'ezplatform.core.session.handler.native_redis');
+        $container->setParameter('session.save_path', sprintf('%s:%d', $endpoint['host'], $endpoint['port']));
     }
 } elseif (isset($relationships['rediscache'])) {
     foreach ($relationships['redissession'] as $endpoint) {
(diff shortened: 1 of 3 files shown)


app/config/env/platformsh.php
```diff
@@ -55,21 +55,21 @@
...
         }
 
-        ini_set('session.save_handler', 'redis');
-        ini_set('session.save_path', sprintf('%s:%d', $endpoint['host'], $endpoint['port']));
+        $container->setParameter('ezplatform.session.handler_id', 'ezplatform.core.session.handler.native_redis' );
+        $container->setParameter('session.save_path', sprintf('%s:%d', $endpoint['host'], $endpoint['port']));
     }
 } else {
-    // Store session into /tmp.
-    ini_set('session.save_path', '/tmp/sessions');
```
vidarl: @Crell : Is it really needed to explicitly set save_path to /tmp/session ?

In my platform.sh envirment, `/tmp/session` is the default path defined in php.ini in the php container.
I assume it is safe to assume this is the case for all php projects on Platform.sh?

vidarl: The reason I ask is that I would prefer to avoid dependency to `session.handler.native_file` if it anyway is unnecessary.

Crell: Yes, that would be the same system-wide.  While hypothetically we could change that system-wide, I can't imagine why we'd ever want to.  I included it in my code because it was included in the code I was modifying from elsewhere.  If it works without it then it's probably safe to omit.

vidarl: Ok, I removed this couple of lines.

vidarl:
squashed and rebased

andrerom:
Note: Merge up will wait until 2.0.2 is put.

PR #263, opened 2018-02-09
EZP-28844: Fix cluster support in docker and Platform.sh

vidarl:
This PR makes it convenient to config cluster settings when using docker or platform.sh
by cluster config I mean:
- session stored in redis
- cache stored in redis
- enabling dfs file handler

### DIFF
--- a/app/config/env/platformsh.php
+++ b/app/config/env/platformsh.php
@@ -18,12 +18,43 @@ foreach ($relationships['database'] as $endpoint) {
     $container->setParameter('database_name', $endpoint['path']);
     $container->setParameter('database_user', $endpoint['username']);
     $container->setParameter('database_password', $endpoint['password']);
-    $container->setParameter('database_path', '');
 
+    // 'cluster_database_name' is deprecated in eZ Platform 1.13.1/2.1
     // Cluster DB name is hardcoded. It will have no any effect if cluster is disabled
     $container->setParameter('cluster_database_name', 'cluster');
 }
 
+// PLATFORMSH_DFS_NFS_PATH is different compared to DFS_NFS_PATH in the sense that it is relative to ezplatform dir
+// DFS_NFS_PATH is an absolute path
+if ($dfsNfsPath = getenv('PLATFORMSH_DFS_NFS_PATH')) {
+    $container->setParameter('dfs_nfs_path', sprintf('%s/%s', getcwd(), $dfsNfsPath));
+
+    if (array_key_exists('dfs_database', $relationships)) {
+        foreach ($relationships['dfs_database'] as $endpoint) {
+            if (empty($endpoint['query']['is_master'])) {
+                continue;
+            }
+
+            $container->setParameter('dfs_database_driver', 'pdo_' . $endpoint['scheme']);
+            $container->setParameter('dfs_database_host', $endpoint['host']);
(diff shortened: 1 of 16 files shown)


app/config/env/platformsh.php
```diff
@@ -18,12 +18,44 @@
     $container->setParameter('database_name', $endpoint['path']);
     $container->setParameter('database_user', $endpoint['username']);
     $container->setParameter('database_password', $endpoint['password']);
-    $container->setParameter('database_path', '');
 
+    // 'cluster_database_name' is deprecated in eZ Platform 2.1
```
vidarl: Yeah, I want to deprecate this one. I guess 2.1 is the first upcoming version where we may do so?

app/config/env/platformsh.php
```diff
@@ -18,21 +21,56 @@
...
         if ($endpoint['scheme'] !== 'redis') {
             continue;
         }
 
+        $container->setParameter('cache_pool', 'singleredis');
         $container->setParameter('cache_host', $endpoint['host']);
         $container->setParameter('cache_redis_port', $endpoint['port']);
+
+        $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__ . '/../cache_pool'));
+        $loader->load('singleredis.yml');
```
andrerom: prior attempt was to let generic.php take care of the loading by specifying the needed env variables.

app/config/env/platformsh.php
andrerom: Should be fine, can also write 1.13.1/2.1

vidarl: ok. May I add, ", removed in 2.2" ?

app/config/env/platformsh.php
vidarl: well.
In generic.php it loads the yml if `$CUSTOM_CACHE_POOL` is set, which is fine.

So, we could require platform.sh users to also set that environment variable in order to load the yaml, but the rest of the logic will anyway be different as `cache_host` does not have a fixed value on platform.sh. It needs to be fetched from `$PLATFORM_RELATIONSHIPS` anyway.

Thus, I think it is better that we also autodetect the whole thing on platform.sh using `$PLATFORM_RELATIONSHIPS`

app/config/env/platformsh.php
andrerom: Then you can just as well remove it in master (2.1)

vidarl:
> Besides inline comment, do we also need change on apache vhost for the dfs change?

apache template fixed

andrerom:
Looks good, should there perhaps be a JIRA issue number mentioned in the PR title and linked to in description?

app/config/env/platformsh.php
andrerom: ok, then make sure we are consistent on this.

vidarl: so, you approve then?

We could keep the load() code only in generic.php by changing the condition for it's execution. Something like `if $loadSingleRedisYml then ....`, and then set `$loadSingleRedisYml = true;` when needed in both generic.php and platformsh.php

andrerom: yes, it's ok as is, we can cleanup if needed once the other things are in.

vidarl:
Squashed commit and added jira issue

PR #264, opened 2018-02-11
Wrong number in LimitRequestBody Apache directive

lserwatka:
48MB is 50331648 bytes (in binary) not 49152 bytes. This breaks almost all REST based file uploads.
Ref. to http://httpd.apache.org/docs/2.0/mod/core.html#limitrequestbody


andrerom:
Manually cherry picked in f3f99e0

PR #265, opened 2018-02-15
Remove Twig dependency as it is already specified by kernel

emodric:
Small continuation to #247


emodric:
> On second thought, this repo uses twig, especially demo variants, so having dependency makes sense here.

Well, this repo does not use Twig, only demo variants do :) I'm trying to avoid discrepancies between required versions that build up over time.

PR #274, opened 2018-03-08
EZP-26809: unifying install instructions

mere-teresa:
See https://github.com/ezsystems/developer-documentation/pull/203

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -18,7 +18,7 @@ If you don't have it already, install Composer, the command-line package manager
 php -r "readfile('https://getcomposer.org/installer');" | php
 ```
 
-For further information about Composer usage see [Using Composer](https://doc.ezplatform.com/en/latest/getting_started/about_composer/).
+For further information about Composer usage see the [Using Composer](https://doc.ezplatform.com/en/latest/getting_started/about_composer/) section.
 
 ### Install eZ Platform
 
@@ -45,7 +45,7 @@ php bin/console assetic:dump
 
 #### Installing another version
 
-The instructions above show how to install the latest stable version, however with Composer you can specify the version and stability if you want to install something else.
+The instructions above show how to install the latest stable version, however with Composer you can specify the exact version and stability level you want to install.
 
 Versions [can be expressed in many ways in Composer](https://getcomposer.org/doc/articles/versions.md), but the ones we recommend are:
 
@@ -73,15 +73,15 @@ php bin/console ezplatform:install clean
 
 ## Setting up directory permissions
 
-Directories `var`, `web/var` need to be writable by cli and web server user
+Directories `var`, `web/var` need to be writable by CLI and web server user
 (see [Symfony documentation](http://symfony.com/doc/3.4/setup/file_permissions.html))
 
 Furthermore, future files and directories created by these two users will need to inherit those permissions.
 
-*For security reasons, in production there is no need for web server to have access to write to other directories.*
+*For security reasons, in production there is no need for web server to have write permission to other directories.*
 
 For development setup you may change your web server config to use the same user as the owner of a directory.
-What follows below is mainly for production setup, and like Symfony we first and foremost recommend using an approach using ACL.
+What follows below is mainly for production setup, and like Symfony we first and foremost recommend using an ACL.
 
 ### Using ACL on a Linux/BSD system that supports chmod +a
 
@@ -98,7 +98,7 @@ sudo chmod +a "$(whoami) allow delete,write,append,file_inherit,directory_inheri
 
 ### Using ACL on a Linux/BSD system that does not support chmod +a
 
-Some systems don't support chmod +a, but do support another utility called setfacl. You may need to
+Some systems don't support `chmod +a`, but do support another utility called setfacl. You may need to
 [enable ACL support](https://help.ubuntu.com/community/FilePermissionsACLs) on your partition and install setfacl
 before using it *(as is the case with Ubuntu)*. The example below uses a command to try to determine
 your web server user and set it as `HTTPDUSER`, alternatively change to your actual web user if non standard:
@@ -127,14 +127,14 @@ sudo find web/var var -type f | xargs sudo chmod -R 664
 ### Using chmod on a Linux/BSD/OS X system where you can't change owner
 
 If you can't use ACL and aren't allowed to change owner, you can use chmod, making the files writable by everybody.
-Note that this method really isn't recommended as it allows any user to do anything.
+**Note that this method really isn't recommended as it allows any user to do anything.**
 
 ```bash
 sudo find web/var var -type d | xargs sudo chmod -R 777
 sudo find web/var var -type f | xargs sudo chmod -R 666
 ```
 
-When using chmod, note that newly created files (such as cache) owned by the web server's user may have different/restrictive permissions.
+When using `chmod`, note that newly created files (such as cache) owned by the web server's user may have different/restrictive permissions.
 In this case, it may be required to change the umask so that the cache and log directories will be group-writable or world-writable (`umask(0002)` or `umask(0000)` respectively).
 
 It may also be possible to add the group ownership inheritance flag so new files inherit the current group, and use `775`/`664` in the command lines above instead of world-writable:
@@ -143,7 +143,7 @@ It may also be possible to add the group ownership inheritance flag so new files
 sudo chmod g+s web/var var
 ```
 
-### Set up directory permissions on Windows
+### Setting up directory permissions on Windows
 
 For your choice of web server you'll need to make sure web server user has read access to `<root-dir>`, and
 write access to the following directories:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,90 +1,61 @@
-# Installation instructions
+# Installation
 
-NOTE: *For simplified installation, consider using [eZ Launchpad](https://ezsystems.github.io/launchpad/) which takes care about the whole setup for you.*
+NOTE: *For simplified installation, consider using [eZ Launchpad](https://ezsystems.github.io/launchpad/) which takes care of the whole setup for you.*
 
-These installation instructions are kept current in the online docs here:
-https://doc.ezplatform.com/en/latest/getting_started/install_using_composer/
+These installation instructions are kept current [in the online docs](https://doc.ezplatform.com/en/latest/getting_started/install_using_composer/).
 
-## Prerequisite
+## Prerequisites
 
-  These instructions assume you have technical knowledge and have already installed PHP, web server & *a database server* needed for this software. For further information [on requirements see online doc](https://doc.ezplatform.com/en/latest/getting_started/requirements_and_system_configuration/).
+These instructions assume you have already installed PHP, a web server and a database server needed for eZ Platform. For further information [on requirements see online doc](https://doc.ezplatform.com/en/latest/getting_started/requirements_and_system_configuration/).
 
-# Installation Using Composer
+## Installation
 
-## Get Composer
+### Get Composer
 
-If you don't have it already, install Composer, the command-line package manager for PHP. You'll have to have a copy of Git installed on your machine. The following command uses PHP to download and run the Composer installer, and should be entered on your terminal and executed by pressing Return or Enter:
+If you don't have it already, install Composer, the command-line package manager for PHP. You need a copy of git installed on your machine. The following command uses PHP to download and run the Composer installer:
 
 ``` bash
 php -r "readfile('https://getcomposer.org/installer');" | php
 ```
 
-For further information about Composer usage see [Using Composer](about_composer.md).
+For further information about Composer usage see [Using Composer](https://doc.ezplatform.com/en/latest/getting_started/about_composer/).
 
-## eZ Platform Installation
+### Install eZ Platform
 
-The commands below assume you have Composer installed globally, a copy of git on your system, and your **MySQL/MariaDB server *already set up* with a database**. Once you've got all the required PHP extensions installed, you can get eZ Platform up and running with the following commands:
+The commands below assume you have Composer installed globally, a copy of git on your system, and your **MySQL/MariaDB server already set up with a database**. Once you've got all the required PHP extensions installed, you can get eZ Platform up and running with the following commands:
 
 ``` bash
-composer create-project --no-dev --keep-vcs ezsystems/ezplatform ezplatform
+composer create-project --keep-vcs ezsystems/ezplatform ezplatform
 cd ezplatform
+```
 
-#at the end of the install process, you will be told to perform the following commands:
-export SYMFONY_ENV="prod"
+At this point you need to [set up directory permissions](#setting-up-directory-permissions).
+
+Next, use the installation command:
+
+``` bash
 php bin/console ezplatform:install clean
+```
+
+Finally, dump assets:
+
+``` bash
 php bin/console assetic:dump
 ```
 
-!!! note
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -5,178 +5,99 @@ NOTE: *For simplified installation, consider using [eZ Launchpad](https://ezsyst
...
 
-       ```bash
-       $ HTTPDUSER=$(ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1)
-       # if this does not work, try adding '-n' option
-       $ sudo setfacl -dR -m u:"$HTTPDUSER":rwX -m u:$(whoami):rwX var web/var
-       $ sudo setfacl -R -m u:"$HTTPDUSER":rwX -m u:$(whoami):rwX var web/var
-       ```
+#at the end of the install process, you will be told to perform the following commands:
+export SYMFONY_ENV="prod"
+php bin/console ezplatform:install clean
```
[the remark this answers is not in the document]
DominikaK: Also, does this command completely replace the `CREATE DATABASE <database> CHARACTER SET utf8;` from the old version?

INSTALL.md
```diff
@@ -5,178 +5,99 @@ NOTE: *For simplified installation, consider using [eZ Launchpad](https://ezsyst
...
-       - `<directory>`: Folder to extract to, if omitted same as package name. In example specified to `ezplatform`.
-       - `<version>`: Optional, *when omitted you'll get latest stable*. Examples:
-        - `^2.0@beta`: To pick latests 2.x beta release, to pick latests 2.0.x release use `~2.0.0`.
-        - `v2.0.1` : To pick a specific tag, could also have been `v2.0.0-beta4`
-        - `dev-master` : Picks a development version from a branch, here `master`.
-
-     Further reading: https://getcomposer.org/doc/03-cli.md#create-project
+  These instructions assume you have technical knowledge and have already installed PHP, web server & *a database server* needed for this software. For further information [on requirements see online doc](https://doc.ezplatform.com/en/latest/getting_started/requirements_and_system_configuration/).
 
-2. **Setup folder rights**<a name="install-2-folder-rights"></a>:
```
DominikaK: If the folder rights section is removed, maybe link explicitly to this page in the main doc?

INSTALL.md
```diff
@@ -5,178 +5,99 @@ NOTE: *For simplified installation, consider using [eZ Launchpad](https://ezsyst
...
 
-       B. **Using ACL on a *Linux/BSD* system that does not support chmod +a**
+The commands below assume you have Composer installed globally, a copy of git on your system, and your **MySQL/MariaDB server *already set up* with a database**. Once you've got all the required PHP extensions installed, you can get eZ Platform up and running with the following commands:
 
-       Some systems don't support chmod +a, but do support another utility called setfacl. You may need to
-       [enable ACL support](https://help.ubuntu.com/community/FilePermissionsACLs) on your partition and install setfacl
-       before using it *(as is the case with Ubuntu)*. With it installed example below uses a command to try to determine
-       your web server user and set it as ``HTTPDUSER``, alternatively change to your actual web user if non standard:
+``` bash
+composer create-project --no-dev --keep-vcs ezsystems/ezplatform ezplatform
```
DominikaK: The other PR (to developer-documentation) removes `--no-dev` here. Also, just making sure that removing `php -d memory_limit=-1` is confirmed as the recommended way.

bdunogier:
Out of curiosity, have we considered having a simple link to the online documentation for installation ? It is very common, and avoids duplicating information.

andrerom:
We probably should, so it can be made part of README.md like in demo repos so they can be kept in sync better. But as long as online doc is dumbing things down and skipping important considerations for production it is not really best place to link to yet. But yes, let's fix that.

bdunogier:
> But as long as online doc is dumbing things down and skipping important considerations for production

I agree, simplification isn't always the ultimate goal.
I'd expect that the link goes to a cover page for installation, with links to both simple instructions for the most common dev cases as well as production considerations.

DominikaK:
@andrerom @bdunogier, so to confirm: would you be fine with this INSTALL.md containing only a link to the developer doc (ultimately, assuming the install guide over there is full and not "dumbed down")? One disadvantage is there would be no installation guide cloned with the repo, buth other than that it's definitely better to maintain.
So we could finish and merge this PR for the time being, complete and polish the dev-doc installation, and then make one more PR replacing INSTALL.md with minimal info + link.

bdunogier:
> would you be fine with this INSTALL.md containing only a link to the developer doc (ultimately, assuming the install guide over there is full and not "dumbed down")? One disadvantage is there would be no installation guide cloned with the repo, buth other than that it's definitely better to maintain.

Yes. The only doubt is that maybe a simple link to the doc would be better in the README.md, in an Installation chapter. That way, it would be visible directly from https://github.com/ezsystems/ezplatform.

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
-    B. **Install via Composer**
+``` bash
+php bin/console assetic:dump
+```
 
-     You can get eZ Platform using Composer with the following commands:
+#### Installing another version
 
-     *Note: composer will take its time to download all libraries and when done you will be asked to fill in some settings, including database settings.*
+The instructions above show how to install the latest stable version, however with Composer you can specify the version and stability if you want to install something else.
```
MagdalenaZuba: can specify the exact version and stability level you want to install.

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
+php bin/console ezplatform:install clean
+```
 
-       A. **Using ACL on a *Linux/BSD* system that supports chmod +a**
+## Setting up directory permissions
 
-       Some systems, primarily Mac OS X, supports setting ACL using a `+a` flag on `chmod`. Example uses a command to
-       try to determine your web server user and set it as ``HTTPDUSER``, alternatively change to your actual web server
-       user if non standard:
+Directories `var`, `web/var` need to be writable by cli and web server user
```
MagdalenaZuba: CLI

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
-       ```bash
-       $ rm -rf var/cache/* var/logs/* var/sessions/*
-       $ HTTPDUSER=$(ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1)
-       $ sudo chmod +a "$HTTPDUSER allow delete,write,append,file_inherit,directory_inherit" var web/var
-       $ sudo chmod +a "$(whoami) allow delete,write,append,file_inherit,directory_inherit" var web/var
-       ```
+Furthermore, future files and directories created by these two users will need to inherit those permissions.
 
-       B. **Using ACL on a *Linux/BSD* system that does not support chmod +a**
+*For security reasons, in production there is no need for web server to have access to write to other directories.*
```
MagdalenaZuba: to have write permission to other directories.

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
 
-       Some systems don't support chmod +a, but do support another utility called setfacl. You may need to
-       [enable ACL support](https://help.ubuntu.com/community/FilePermissionsACLs) on your partition and install setfacl
-       before using it *(as is the case with Ubuntu)*. With it installed example below uses a command to try to determine
-       your web server user and set it as ``HTTPDUSER``, alternatively change to your actual web user if non standard:
+For development setup you may change your web server config to use the same user as the owner of a directory.
+What follows below is mainly for production setup, and like Symfony we first and foremost recommend using an approach using ACL.
```
MagdalenaZuba: we recommend using an ACL.

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
-       ```bash
-       $ HTTPDUSER=$(ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1)
-       $ sudo chown -R "$HTTPDUSER":"$HTTPDUSER" var web/var
-       $ sudo find web/var var -type d | xargs sudo chmod -R 775
-       $ sudo find web/var var -type f | xargs sudo chmod -R 664
-       ```
+### Using ACL on a Linux/BSD system that does not support chmod +a
 
-       D. **Using chmod on a *Linux/BSD/OS X* system where you can't change owner**
+Some systems don't support chmod +a, but do support another utility called setfacl. You may need to
```
MagdalenaZuba: `chmod +a`

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
+If you can't use ACL and aren't allowed to change owner, you can use chmod, making the files writable by everybody.
+Note that this method really isn't recommended as it allows any user to do anything.
 
+```bash
+sudo find web/var var -type d | xargs sudo chmod -R 777
+sudo find web/var var -type f | xargs sudo chmod -R 666
+```
 
-3. **Run installation command**<a name="install-4-db-setup"></a>:
+When using chmod, note that newly created files (such as cache) owned by the web server's user may have different/restrictive permissions.
```
MagdalenaZuba: `chmod`

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
 
-       E. **Setup folder rights on Windows**
+### Using chmod on a Linux/BSD/OS X system where you can't change owner
 
-       For your choice of web server you'll need to make sure web server user has read access to `<root-dir>`, and
-       write access to the following directories:
-       - web/var
-       - var
+If you can't use ACL and aren't allowed to change owner, you can use chmod, making the files writable by everybody.
+Note that this method really isn't recommended as it allows any user to do anything.
```
MagdalenaZuba: putting this sentence in a colorful warning frame would be more effective

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
 
-    ```bash
-    $ php -d memory_limit=-1 bin/console ezplatform:install --env prod clean
-    ```
+```bash
+sudo chmod g+s web/var var
+```
 
-    **Note**: Password for the generated `admin` user is `publish`, this name and password is needed when you would like to login to backend Admin UI where you can and should change this.
+### Set up directory permissions on Windows
```
MagdalenaZuba: Setting up

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
-       - `<directory>`: Folder to extract to, if omitted same as package name. In example specified to `ezplatform`.
-       - `<version>`: Optional, *when omitted you'll get latest stable*. Examples:
-        - `^2.0@beta`: To pick latests 2.x beta release, to pick latests 2.0.x release use `~2.0.0`.
-        - `v2.0.1` : To pick a specific tag, could also have been `v2.0.0-beta4`
-        - `dev-master` : Picks a development version from a branch, here `master`.
+-   Exact git tag: `v1.3.1`
+-   Tilde for locking down the minor version: `~1.3.0`
+    -   Equals: `1.3.*``
+-   Caret for allowing all versions within a major: `^1.3.0`
+    -   Equals: `1.* <= 1.3.0`
```
andrerom: I really love the new composer section here, great job on making this understandable. But do we need to duplicate this across our repos as opposed to drill it down to one relevant example and link to our online doc for more?

INSTALL.md
```diff
@@ -1,182 +1,151 @@
...
-       ```bash
-       $ cd /<directory>/
-       $ curl -sS https://getcomposer.org/installer | php
-       $ php -d memory_limit=-1 composer.phar run-script post-install-cmd
-       ```
+Finally, dump assets:
 
-    B. **Install via Composer**
+``` bash
+php bin/console assetic:dump
```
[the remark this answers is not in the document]
andrerom: Because if we manage to shorten this, we can make it part of README.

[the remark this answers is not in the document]
DominikaK: Just checked that demo installed correctly without adding assetic:dump, so I'll remove this.

But still, as we need the directory permissions section, I'm not sure it can be short enough to include in readme (unless we can just link to the "set up permissions" page in main doc, which is by the way also being modified here: https://github.com/ezsystems/developer-documentation/pull/203)

INSTALL.md
DominikaK: That actually comes from the "Install using composer" page in the main doc, so I guess I can replace this here with a link as well.

INSTALL.md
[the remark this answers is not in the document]
andrerom: > But still, as we need the directory permissions section, I

depends, the install doc here now in this state is completely lacking instructions on setting up web server, if you intend to do like I did in the simplied doc: https://github.com/ezsystems/developer-documentation/pull/228/files

then we don't need it.

Sum:
- when using what the _Welcome to eZ Platform message_ suggest you to do, namely `server:start/run/stop`, then no need to cover permissions, as it runs using the same user as you executes it with, so same access you have as a user
- permissions should be covered in the place we describe how to setup Apache / Nginx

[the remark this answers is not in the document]
DominikaK: Thanks, I removed the directory permission as you wrote, and added a link to the main doc (which doesn't work as yet, but will once the PR over there is merged).
As for "setup for production", having a page like this is definitely a plan ofr (relatively near) future.

INSTALL.md
```diff
@@ -1,182 +1,42 @@
...
-     _Once installed, for additional settings you may configure here, see also `app/config/default_parameters.yml`._
+The commands below assume you have Composer installed globally, a copy of git on your system, and your **MySQL/MariaDB server already set up with a database**. Once you've got all the required PHP extensions installed, you can get eZ Platform up and running with the following commands:
 
+``` bash
+composer create-project --keep-vcs ezsystems/ezplatform ezplatform
+cd ezplatform
+```
 
-    A. **Extract archive** (tar/zip)
+At this point you need to [set up directory permissions](https://doc.ezplatform.com/en/latest/getting_started/set_up_directory_permissions/).
```
andrerom: I'm more or less +1 now, one last thing feels like it's missing here to make it complete tough.

The `composer create-project` command above will ask for misc parameters and in the end output the following:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Composer/ScriptHandler.php#L109-L135

So we can rewrite everything after that _(including this line)_ to reflect this the same way as we did in composer install doc recently:
https://doc.ezplatform.com/en/2.1/getting_started/install_using_composer/#ez-platform-installation

But, we can even simplify that further,  if install command is changed to
``` bash
composer create-project --keep-vcs ezsystems/ezplatform ezplatform ^2
cd ezplatform
```

Then we don't need to mention PHP 7.1, as composer will complain for us if user does not have right PHP version, up to you if that makes sense here, but saves us having to explain why they might get v1.

So with that we can simplify what we write after the command to something like _(but feel free to write it differently)_:
```

The composer command above will instruct you on what to do next:
- During `create-project` command, you will be asked to input things like database host name, login, password and so on. This will be generated to `<ezplatform>/app/config/parameters.yml`.
- Once done you will be given further instructions on how to install data into the database, and finally how to run a simplified dev server using `server:run` Symfony command to get access. _Note: For a more complete and far better performing setup using Apache or Nginx, read up on how to [install eZ Platform manually](https://doc.ezplatform.com/en/latest/getting_started/install_manually/)_.
´´´

That, I think sums up, and syncs this up with `install_using_composer` doc. I have further ideas on how to make this shorter and as part of Readme, but I'll spare you the iterations and rather do PR for you to review as followup ti this

PR #275, opened 2018-03-12
EZP-26491: Remove "Forwarded" header in the VCL to avoid ConflictingHeadersException

natanael89:
> JIRA issue: [EZP-26491](https://jira.ez.no/browse/EZP-26491)

This PR adds the following line: https://github.com/ezsystems/ezplatform/blob/1.13/doc/varnish/vcl/varnish4_xkey.vcl#L31 to our other VCL templates. This is done to avoid ConflictingHeadersException, which is thrown by Symfony when there are both "X-Forwarded-For" and "Forwarded" headers set.
This is also suggested in the Symfony documentation: https://symfony.com/doc/2.8/http_cache/varnish.html#make-symfony-trust-the-reverse-proxy.

### DIFF
--- a/doc/varnish/vcl/varnish3.vcl
+++ b/doc/varnish/vcl/varnish3.vcl
@@ -13,6 +13,10 @@ sub vcl_recv {
     // Advertise Symfony for ESI support
     set req.http.Surrogate-Capability = "abc=ESI/1.0";
 
+    // Varnish, in its default configuration, sends the X-Forwarded-For header but does not filter out Forwarded header
+    // To be removed in Symfony 3.3
+    unset req.http.Forwarded;
+
     // Add a unique header containing the client address (only for master request)
     // Please note that /_fragment URI can change in Symfony configuration
     if (!req.url ~ "^/_fragment") {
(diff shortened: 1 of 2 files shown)


davidsayre:
Thank you for looking into this ticket.
I really appreciate that someone was able to replicate and confirm this.

andrerom:
You're welcome @davidsayre!

Note: Be aware that the `ConflictingHeadersException` hints that for some reason something is setting conflicting `Forward` header. I can think of two possible cases, 1. load balancer injecting header change but not injecting similar change on X-Forward, if so this is worth looking into 2. Browsers/users trying to spoof this header for some reason, if so then this fixes it.

Additional side note: Symfony 3 does not suggest unsetting this any longer, so might be Symfony 3 is handling this differently.

natanael89:
FYI: `Forward` header is set by the browsers in at least two cases (that I know of):
* when using Opera Mini proxy servers: https://forums.opera.com/topic/17082/conflicting-headers
* on Chrome Mobile, when using Google Data Compression Proxy: https://github.com/symfony/symfony/issues/20215

The issue was solved in Symfony 3.3 by not trusting any header by default: https://github.com/symfony/symfony/issues/20178 and https://github.com/symfony/symfony/pull/22238.

PR #276, opened 2018-03-16
EZEE-1910: [Platform.sh] Speed up deployment when in production to avoid downtime

andrerom:
> Issue: https://jira.ez.no/browse/EZEE-1910

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -81,10 +81,14 @@ hooks:
             php -d memory_limit=-1 app/console ezplatform:install $INSTALL_EZ_INSTALL_TYPE
             touch web/var/.platform.installed
         fi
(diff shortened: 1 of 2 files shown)


andrerom:
@vidarl / @damianz5 Updated based on point made by @vidarl.

But to be honest I'm a bit torned on it. as re-deployment should normally result in same composer packages as people are supposed to use composer.lock. 

Maybe the code here should only execute if it is missing? And if it is, it will have to clear the whole folder, it can do this by renaming it to something that will be removed in post_deploy.

What do you think?

andrerom:
@vidarl / @damianz5  Updated again, looks ok?

vidarl:
yes

PR #278, opened 2018-03-19
EZP-28950: MySQL UTF8 doesn't support 4-byte chars

glye:
> Target version: 2.2
> Depends on https://github.com/ezsystems/ezpublish-kernel/pull/2277

We should use the `utf8mb4` mysql charset which supports all of unicode, instead of `utf8` which does not support 4-byte characters (which includes some Chinese, emojis, and other things).

### DIFF
--- a/app/config/dfs/dfs.yml
+++ b/app/config/dfs/dfs.yml
@@ -9,7 +9,7 @@ doctrine:
                 user: "%dfs_database_user%"
                 password: "%dfs_database_password%"
                 dbname: "%dfs_database_name%"
-                charset: UTF8
+                charset: utf8mb4
 
 # define the flysystem handler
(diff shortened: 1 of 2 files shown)


app/config/dfs/dfs.yml
```diff
@@ -9,7 +9,7 @@ doctrine:
                 user: "%dfs_database_user%"
                 password: "%dfs_database_password%"
                 dbname: "%dfs_database_name%"
-                charset: UTF8
+                charset: utf8mb4
```
andrerom: note: does this work with postgres / sqlite ?

glye: Right, probably doesn't, in which case I'll have to put it in the upgrade doc instead.

PR #280, opened 2018-03-21
Simplify Platform.sh configuration

Plopix:
Hello this PR removes a lot of things, let me explain (and start the discussion)

```diff
-        if [ "$SYMFONY_ENV" = "dev" ] ; then
-            composer install --prefer-dist --no-progress --no-interaction --optimize-autoloader
-        else
-            composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader
-        fi
+        composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader
```
Why would use use Platform.sh in "dev" mode? Even if we could find reason, I don't think we should add this in the `.platform.app.yaml` it complexifies for nothing.

```diff
-        # When deploying changes to existing cluster, clear all cache now that we have shared mounts available
-        if [ "$SYMFONY_ENV" != "prod" ] ; then
-            # Clear class cache before we boot up symfony in case of interface changes on classes cached
-            rm -Rf var/cache/$SYMFONY_ENV/*.*
-            bin/console cache:clear
-        elif [ -d "var/cache/prod/$PLATFORM_TREE_ID" ] ; then
-            # Clear cache on re-deploy when the folder exits, move folder so post_deploy can cleanup
-            mv -f var/cache/prod/$PLATFORM_TREE_ID var/cache/prod/old_deploy
-        fi
+        # if you re build with the same sha, you still want to clear the cache
+        mv var/cache/$SYMFONY_ENV var/cache/old_deployed_cache 2>/dev/null
+        bin/console cache:clear
```

This seems related to https://jira.ez.no/browse/EZEE-1910, where on PE (Platform.sh Enterprise) the mount points are NFS mount points (which is not true in PS)
I think this code is not good, for a common mistake. 
**`cache` and `logs` in multi-node configuration MUST NOT be shared**

Then in PE, things have to be handled differently. We need a cache and a logs folder that are writable but not shared between nodes.

So the correction in EZEE-1910 is a false positive fix, it corrects a situation that should not exist.

We should probably have a `.platform.app.yaml` for PS and one for PE to simplify each of them.

Also about:
```
    ##bin/console doctrine:migrations:migrate --no-interaction --allow-no-migration
    ##bin/console kaliop:migration:migrate --no-interaction --no-debug
```

Should we leverage `composer run-scripts` option? to make it better and more abstract to the deployment architecture?

What do you think?

cc @andrerom @vidarl

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -68,11 +68,7 @@ hooks:
     build: |
         set -e
         rm web/app_dev.php
-        if [ "$SYMFONY_ENV" = "dev" ] ; then
-            composer install --prefer-dist --no-progress --no-interaction --optimize-autoloader
-        else
-            composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader
-        fi
+        composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader
     # Deploy hook, access to services & done once (per cluster, not per node), only mounts are writable at this point
     # Note: Http traffic is paused while this is running, so for prod code this should finish as fast as possible, < 30s
     deploy: |
@@ -83,23 +79,16 @@ hooks:
             php -d memory_limit=-1 bin/console ezplatform:install $INSTALL_EZ_INSTALL_TYPE
             touch web/var/.platform.installed
         fi
-        # When deploying changes to existing cluster, clear all cache now that we have shared mounts available
-        if [ "$SYMFONY_ENV" != "prod" ] ; then
-            # Clear class cache before we boot up symfony in case of interface changes on classes cached
-            rm -Rf var/cache/$SYMFONY_ENV/*.*
-            bin/console cache:clear
-        elif [ -d "var/cache/prod/$PLATFORM_TREE_ID" ] ; then
-            # Clear cache on re-deploy when the folder exits, move folder so post_deploy can cleanup
-            mv -f var/cache/prod/$PLATFORM_TREE_ID var/cache/prod/old_deploy
-        fi
+        # if you re build with the same sha, you still want to clear the cache
+        mv var/cache/$SYMFONY_ENV var/cache/old_deployed_cache 2>/dev/null
+        bin/console cache:clear
         # Example of additional deploy hooks if you use doctrine and/or kaliop migration bundle
         ##bin/console doctrine:migrations:migrate --no-interaction --allow-no-migration
         ##bin/console kaliop:migration:migrate --no-interaction --no-debug
-    # Post deploy hook, like deploy but after being deployed and live, for deploy tasks we can do asynchronously
     post_deploy: |
         set -e
         # Cleanup old prod cache folders, basically all except current $PLATFORM_TREE_ID folder.
-        find var/cache/prod -mindepth 1 -maxdepth 1 -type d \! \( -name "$PLATFORM_TREE_ID" \) -exec rm -rf '{}' \;
+        find var/cache/$SYMFONY_ENV -mindepth 1 -maxdepth 1 -type d \! \( -name "$PLATFORM_TREE_ID" \) -exec rm -rf '{}' \;
 
 # The configuration of scheduled execution.
 # see http://symfony.com/doc/current/components/console/introduction.html
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -69,27 +69,27 @@ class AppKernel extends Kernel
         return __DIR__;
     }
 
+    private function getSymfonyTempDirPrefix()
+    {
+        return !empty($_SERVER['SYMFONY_TMP_DIR']) ? rtrim($_SERVER['SYMFONY_TMP_DIR'], '/') : dirname(
+            $this->getRootDir()
+        );
+    }
+
     public function getCacheDir()
     {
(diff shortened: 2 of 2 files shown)


.platform.app.yaml
```diff
@@ -83,24 +79,11 @@ hooks:
...
-        if [ "$SYMFONY_ENV" != "prod" ] ; then
-            # Clear class cache before we boot up symfony in case of interface changes on classes cached
-            rm -Rf var/cache/$SYMFONY_ENV/*.*
-            bin/console cache:clear
-        elif [ -d "var/cache/prod/$PLATFORM_TREE_ID" ] ; then
-            # Clear cache on re-deploy when the folder exits, move folder so post_deploy can cleanup
-            mv -f var/cache/prod/$PLATFORM_TREE_ID var/cache/prod/old_deploy
-        fi
+        rm -Rf var/cache/$SYMFONY_ENV
+        bin/console cache:clear
```
andrerom: Yoou are breaking changes we do in #276

andrerom:
I'm unsure this simplifies anything, what are you actually trying to accomplish?

> cache and logs in multi-node configuration MUST NOT be shared

I know, but platform.sh does not offer anything else. This is what Symfony itself is doing in their p.sh config. So until p.sh provides local writable mount points we can reliably clear across cluster, OR we once we are on Symfony 4 which no longer needs to write to cache folder on run time, this is what we have atm.

> Why would use use Platform.sh in "dev" mode?

Unsure, @vidarl / @damianz5 maybe you can shed some light on that and if we need it in any use cases / demos?

vidarl:
> Unsure, @vidarl / @damianz5 maybe you can shed some light on that and if we need it in any use cases / demos?

I have never used dev on Platform.sh honestly... Not sure what we want to do when deploying PRs automatically on platform.sh though.... Then it might come handy

Plopix:
my main goal (and it my own opinion) is to reduce the number of lines in this file and remove complexity.
"dev" on Platform.sh does not make a lot of sense to me. You want Platform.sh for your UAT env and you want them as close as the PROD full cache and services enabled etc. (to avoid #EZP-28995 or #EZP-28995 for instance ;)). If you want to enable "dev" mode then, you do your own things but I don't see why we would do all those `if` around that in the provided default.

> I'm unsure this simplifies anything, what are you actually trying to accomplish?
> > cache and logs in multi-node configuration MUST NOT be shared
> I know, but platform.sh does not offer anything else. This is what Symfony itself is doing in their p.sh config. So until p.sh provides local writable mount points we can reliably clear across cluster, OR we once we are on Symfony 4 which no longer needs to write to cache folder on run time, this is what we have atm.

Here I am not sure, I have a project on PE, eZ 1.13.x, and I have specifically asked to remove this "shared" caches/logs over NFS, because it is bad, and should not be shared as you know. (for concurrency etc. etc.)
Cache has to be local.

This code, as a user/dev, scares me:
```
        # When deploying changes to existing cluster, clear all cache now that we have shared mounts available
        if [ "$SYMFONY_ENV" != "prod" ] ; then
            # Clear class cache before we boot up symfony in case of interface changes on classes cached
            rm -Rf var/cache/$SYMFONY_ENV/*.*
            bin/console cache:clear
        elif [ -d "var/cache/prod/$PLATFORM_TREE_ID" ] ; then
            # Clear cache on re-deploy when the folder exits, move folder so post_deploy can cleanup
            mv -f var/cache/prod/$PLATFORM_TREE_ID var/cache/prod/old_deploy
        fi
```

The `if` with `SYMFONY_ENV` is not required.
The `elif` either.

Plus it tends to make me think that is logical to have a shared cache in cluster mode, which again is wrong.

> I have never used dev on Platform.sh honestly... Not sure what we want to do when deploying PRs automatically on platform.sh though.... Then it might come handy

You cannot *just* enable `dev` on PR. It is part of the source code, it would mean that you would enable `dev` in the commit being deploy. (then you would need another commit to remove it and then again to go through the PR etc.)

> You could achieve that with '.environnement` but I don't think it is a good idea.

All of that is what I think and what I do ;-) We all own our `.platform.app.yam` I just wanted to share that.

SylvainGuittard:
I agree with @Plopix about `dev` mode. I don't see the point and it will simplify the file
What do you think @damianz5 ?

andrerom:
On dev: we can remove it if everyone here don't see the need, I don't know where it came from. Question is if we should allow custom env or not? I know someone uses it for custom logic, just not who and how.

In regards of the cache folder change we can rather take that ourselves, the patch is incomplete on that, and like dev change should be applied to 1.13 and up. Basically @Plopix suggested we can just rename the folder early on deploy and cleanup old folder in post_deploy. From what I can see that suggestion is valid, and should work without need for setting custom cache folder. But it does need a extra `mv` during deploy, so question is if that takes noticeable time on nfs (EE) or not.

@antoligy What do you think about these two topics?

andrerom:
Fixed in:
https://github.com/ezsystems/ezplatform/commit/a96931e5801640a375098c94c70e08f6961a1e35#diff-fe14053ff51cb358f8b709415b2a2d17
https://github.com/ezsystems/ezplatform/commit/b3e9460240ab78515ac6fe89c16670a4a7dcebcf#diff-fe14053ff51cb358f8b709415b2a2d17

PR #282, opened 2018-03-22
EZEE-1903: Added solr service template to platform.sh config

vidarl:
This PR adds support for injecting solr config on Platform.sh

TODO:
- [ ] It depends on https://github.com/ezsystems/ezplatform-solr-search-engine/pull/120

### DIFF
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -31,3 +31,21 @@ rediscache:
 #redissession:
 #    type: redis-persistent:3.2
 #    disk: 1024
+
+# If you wish to use solr, uncomment this service and the corresponding relationship in .platform.app.yaml.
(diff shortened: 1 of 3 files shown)


andrerom:
I'm a bit unsure about this, as there is no way for us to automate testing to make sure we verify that the script don't break.

.platform/services.yaml
```diff
@@ -31,3 +31,21 @@ rediscache:
 #redissession:
 #    type: redis-persistent:3.2
 #    disk: 1024
+
+# If you wish to use solr, uncomment this service and the corresponding relationship in .platform.app.yaml.
+# Also, you need to generate  the config using:
```
glye: Hardcore nitpick: Double space after "generate".

PR #283, opened 2018-03-31
Platform.sh conf - Real and more useful example of crontab setting

gauthiergarnier:
I propose to replace the cron example by the one that should be activated by default.
If I'm not wrong, the "ezpublish:cron:run" is dedicated to eZ Platform EE only for the Date-based publishing.

Note: I did not find any details on this crontab (check-urls) in the Installation documentation.

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -105,8 +105,8 @@ hooks:
 # see http://symfony.com/doc/current/components/console/introduction.html
 #crons:
 #    symfony:
-#        spec: "*/20 * * * *"
-#        cmd: "php bin/console ezpublish:cron:run"
+#         spec: "0 0 * * 0"
+#         cmd: "php bin/console ezplatform:check-urls --quiet --env=prod > var/logs/ezp_cron.txt"
 
 runtime:
     extensions:


andrerom:
@adamwojs Is ezplatform:check-urls setup to be run as cron on ee? (should it be? or maybe it rather should be as soon as we get around to open source it so all can use it and help improve it)

adamwojs:
@andrerom It should be as soon as we open source it. Right now it need to be added this manually to CRONtab.

andrerom:
ok, clear.

@vidarl ok with this one in 1.13 and up?

As for master, any chance we can go ahead and open source cron package in order to get this right as of 2.2? @bdunogier @lserwatka

.platform.app.yaml
```diff
@@ -105,8 +105,8 @@ hooks:
 # see http://symfony.com/doc/current/components/console/introduction.html
 #crons:
 #    symfony:
-#        spec: "*/20 * * * *"
-#        cmd: "php bin/console ezpublish:cron:run"
+#         spec: "0 0 * * 0"
+#         cmd: "php bin/console ezplatform:check-urls --quiet --env=prod > var/logs/ezp_cron.txt"
```
vidarl: I am not too happy about the redirect. I think the log should end up in Platform.sh's cron log by default

andrerom:
Closing, added this in https://github.com/ezsystems/ezplatform/commit/ddf654b01a939295f89d0f0e5ad4607baae6aa6f

PR #284, opened 2018-04-06
Use default session handler

mnocon:
Setting the value to the same as in 1.13 branch (https://github.com/ezsystems/ezplatform/blob/1.13/app/config/default_parameters.yml#L4), so that the default session handler will be used.

This is required by our Docker stack: the changes introduced by redis-session container (https://github.com/ezsystems/ezplatform/blob/master/doc/docker/redis-session.yml) have no effect as long as this setting is in place (local value is used instead of master one).

### DIFF
--- a/app/config/default_parameters.yml
+++ b/app/config/default_parameters.yml
@@ -1,7 +1,7 @@
 # This file contains defaults for optional parameters, which you can override in parameters.yml
 parameters:
     locale_fallback: en
-    ezplatform.session.handler_id: session.handler.native_file
+    ezplatform.session.handler_id: ~
 
     # A secret key that's used to generate certain security-related tokens
     secret: '%env(SYMFONY_SECRET)%'


mnocon:
@andrerom @vidarl what do you think?

vidarl:
This fix looks correct to me. I think we should default to PHP's session setting, not default to save sessions to file.

Another question is why this was never merged to 2.0 ? It is a bit awkward that this is in 1.13 and 2.1, but not 2.0 ?

mnocon:
I've chosen 2.1 as base because I guess we are not maintaining 2.0 after 2.1 is released? I don't have anything against merging it there if it's not the case.

andrerom:
> I've chosen 2.1 as base because I guess we are not maintaining 2.0 after 2.1 is released? I don't have anything against merging it there if it's not the case.

Partly correct, as of updated release cycle introduced in Barcelona Partner Conf we support FT releases for 3 months for maintenance and +2 months for security and support.

---

Be aware that the reason why handler is set is because we merged in changes from symfony, so this PR breaks what has been done in https://github.com/symfony/symfony-standard/pull/904

We can consider doing this change but having inline comment about setting handler to this in order to be able to use save_path (it that is relevant), or we can change this to use a parameter which can be adjusted.

Plopix:
Hey I am pasting the message from the other PR here:

I think we missed something here.

Now the default config is:
```yaml
framework:
    session:
        save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'

parameters:
    ezplatform.session.handler_id: session.handler.native_file
```
=> Meaning that by default it is ALWAYS file

If I want eZ to use the php.ini I have to change the default
```yaml
parameters:
    ezplatform.session.handler_id: ~
```
That works, it is fine, but that does not invalidate the logic of this PR.

eZ is looking to the ENV vars for platform.sh why not in the Docker context?

To me there is not reason to treat it differently and would avoid to touch anything we would have the same config locally and in prod using `ezplatform.core.session.handler.native_redis`

I think the session storage should be configured in php.ini. Unfortunately that was not possible at Platform.sh

I agree @vidarl `php.ini` is better but more and more forgotten and making it globally ENV related would make more sense to me. And on Platform.sh it is possible it is just we don't know the HOST as it is dynamic... that could be the case with Docker too.

I think we need to counter part of the `platformsh.php` configuration in the `generic.php` it would be consistent.

Also to answer your question @andrerom 
> If anyone has suggestions for how to set ~ and still use save path when file is used (or remove support for save path) please add input in #284 which is exactly on this topic

Not sure I get it right, but we could have
```yaml
framework:
    session:
        save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'

parameters:
    ezplatform.session.handler_id: ~
```
and if someone change `ezplatform.session.handler_id`  `save_path` will still be OK.
to me `save_path` is not used if `handler_id` is null

andrerom:
@vidarl maybe you can look into this once you are done with current issue? And get this moving forward with @mnocon and @Plopix.

vidarl:
So, since symfony in 3.x defaults to `handler_id:  session.handler.native_file`, I think we should do the same.

And that means that we'll need to change `app/config/env/generic.php` and `doc/docker/redis-session.yml` so that we use ENVIRONMENT variables in order to inject session handler. All agree?

I'll make a new pr....

vidarl:
new pr : https://github.com/ezsystems/ezplatform/pull/304

vidarl:
#304 is now merged.. closing this one....

mnocon:
Thanks @vidarl for taking over this PR!

PR #288, opened 2018-04-27
[Platform.sh] Manually clear Redis cache on deploy

kmadejski:
Related to: https://jira.ez.no/browse/EZP-28816

Tested on: https://master-7rqtwti-u52o3swc3hqj6.eu.platform.sh/

Todo:
- [ ] Once Varnish changes are in, revise (simplify) this + /tmpfs change on top (probably as one PR)


vidarl:
It just hit me that if you don't have a separate redis_session instance, we'll also clear all sessions on deploy with this approach ....

kmadejski:
@vidarl yes, you are right, but I don't see any other way to do that. Maybe we should uncomment second Redis instance for sessions by default?

vidarl:
Maybe it is possible to delete all keys **not** starting with `PHPREDIS_SESSION` ?

It seems not to be straight forward, ref https://stackoverflow.com/questions/4006324/how-to-atomically-delete-keys-matching-a-pattern-using-redis

But deleting everything, including session is better than current solution

andrerom:
Can't we just ship with always using own instance for session? It's way better anyway _(in regards to config of redis being relevant for the job, and the separation)_.

If so, we can btw move all the deploy cache logic to this script.

kmadejski:
> Maybe it is possible to delete all keys not starting with `PHPREDIS_SESSION`?

I believe that it is possible, but might be a little tricky (https://stackoverflow.com/questions/29942541/how-to-get-keys-which-does-not-match-a-particular-pattern-in-redis).

> Can't we just ship with always using own instance for session?

It was my first thought and I also think that this is a better approach. 

> If so, we can btw move all the deploy cache logic to this script.

You meat that part: https://github.com/ezsystems/ezplatform/blob/1.13/.platform.app.yaml#L84-#L92, right?

andrerom:
> You meat that part: https://github.com/ezsystems/ezplatform/blob/1.13/.platform.app.yaml#L84-#L92, right?

yes

vidarl:
Find my me, but I believe Larry @Crell  had some objections to that when I worked on redis settings for Platform.sh. Maybe due to increased resource usage ( a second redis )?

@Crell : may you comment ?

Crell:
A second Redis instance should work, with the caveat that you're then using more resources.  I believe we also discussed making the session Redis a persistent instance, again at the cost of the disk usage set aside for that.

I think we wrote a patch somewhere that includes an optional second Redis instance to just uncomment stuff for, didn't we?

bin/.platform/clear-cache.sh
```diff
@@ -0,0 +1,29 @@
...
+set -e
+
+. ../../.env
+
+function clearRedisCache
+{
+    REDISHOST=`php -r '$r=getenv("PLATFORM_RELATIONSHIPS");$r=json_decode(base64_decode($r), true);echo($r["rediscache"][0]["host"]);'`
+    REDISPORT=`php -r '$r=getenv("PLATFORM_RELATIONSHIPS");$r=json_decode(base64_decode($r), true);echo($r["rediscache"][0]["port"]);'`
+
+    echo "FLUSHALL" | redis-cli -h $REDISHOST -p $REDISPORT
```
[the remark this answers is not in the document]
andrerom: That said, seems we can remove the whole function in favor of `platform redis flushall` command.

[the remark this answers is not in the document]
Crell: The Platform CLI is not available by default, and takes a few steps to setup.  You're probably better off not relying on it directly in a default script as it would add several extra steps to everyone's setup process.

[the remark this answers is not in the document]
kmadejski: > This can be written redis-cli -h $REDISHOST -p $REDISPORT FLUSHALL to make it more readable

Improved in https://github.com/ezsystems/ezplatform/pull/288/commits/fc85471ff26a4d183e6c4e33be00aafcdf6f0d48.

kmadejski:
@andrerom As discussed, I've enabled Redis session storage by default in: https://github.com/ezsystems/ezplatform/pull/288/commits/2a192a172166dcae1f10fdb2459d198ca7c67f6a

.platform/services.yaml
```diff
@@ -23,8 +23,8 @@ rediscache:
 
 # If you wish to have a separate Redis instance for sessions, uncomment
 # this service and the corresponding relationship in .platform.app.yaml.
-#redissession:
-#    type: 'redis:3.2'
+redissession:
+    type: 'redis:3.2'
```
andrerom: 👍 But I think we prefer to have the one below as recommended defaults.

The persisted one, and for sure separated one (?)

Ok @vidarl and @Crell ?

.platform/services.yaml
```diff
@@ -28,6 +28,6 @@ rediscache:
 #
 # Alternatively if you have a requirement that sessions are persisted across server/redis restarts,
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
-#redissession:
-#    type: redis-persistent:3.2
-#    disk: 1024
+redissession:
+    type: redis-persistent:3.2
+    disk: 1024
```
Crell: If you're going to default to enabling a persistent redis for sessions then the comment blocks should be updated accordingly.  As is they're written for that to be opt-in.

andrerom:
@kmadejski Adding label "Needs JIRA Issue" here.

However what we really need is to better describe why we want to do this, if we actually by default clear all cache on _every_ deploy, we end up in cache warming issues as app will be slow to react to requests coming in when back up leading to timeouts if they where already waiting for response during deploy step _(according to reporter, still a bit unclear on the where and how this happens, aka what is causing the timeout)_. So will see if I can get some more info from the reporter on the needs here.

bin/.platform/clear-cache.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+# Due to EZP-28816: console cache:clear doesn't remove cache from Redis, so we need to do it manually.
+
+set -e
+
+. ../../.env
```
andrerom: why was this needed again?

Crell: I'm not sure, but I just realized where this file is.  Files in `.platform` are NOT available in the build phase, as the `.platform` directory controls services, not the application.  This file will need to live elsewhere.

Crell: Oh, wait, it's in bin/.platform.  Ignore me, I misread it.

andrerom: > That said, why the .platform instead of just platformsh

it should _(if we move forward with this)_ be renamed to avoid that very confusion :)

vidarl: I guess the line is there due to the use of `$SYMFONY_ENV` later on.

But I don't think it is correct to get value of `$SYMFONY_ENV` from `.env` as that file is meant for docker-compose

bin/.platform/clear-cache.sh
```diff
@@ -0,0 +1,29 @@
...
+
+    redis-cli -h $REDISHOST -p $REDISPORT FLUSHALL
+}
+
+# When deploying changes to existing cluster, clear all cache now that we have shared mounts available
+if [ "$SYMFONY_ENV" != "prod" ] ; then
+    # Clear class cache before we boot up symfony in case of interface changes on classes cached
+    rm -Rf ../../app/cache/$SYMFONY_ENV/*.*
+    php ../../app/console cache:clear
+elif [ -d "../../app/cache/prod/$PLATFORM_TREE_ID" ] ; then
```
vidarl: Should we check is `$PLATFORM_TREE_ID` and `$PLATFORM_RELATIONSHIPS` is actually set maybe, and abort if not?

.platform.app.yaml
```diff
@@ -24,7 +24,7 @@ relationships:
     rediscache: 'rediscache:redis'
     # If you wish to have a separate Redis instance for sessions, uncomment
```
vidarl: Comments needs to be altered if we enable redissession by default

bin/.platform/clear-cache.sh
andrerom: When will it not be set?

vidarl: if not executed on platform.sh.... so purely user mistake...

.platform.app.yaml
vidarl: The comment still says "If (...), uncomment this relationship (...)" even though it is not uncommented by default

.platform/services.yaml
```diff
@@ -23,11 +23,13 @@ rediscache:
 
 # If you wish to have a separate Redis instance for sessions, uncomment
 # this service and the corresponding relationship in .platform.app.yaml.
```
vidarl: I would recomend to delete the comments in line 24 and 25 entirely ( the corresponding relationship in .platform.app.yaml is anyway enabled by default now )

.platform/services.yaml
```diff
@@ -23,11 +23,13 @@ rediscache:
...
+# For persisted session storage use the `redis-persistent` storage provided below.
```
vidarl: Nitpick, but since persistent storage is enabled by default, I would move this section below the section where `redis-persistent` is enabled so that the user reads about that first

Next, I would change the wording into something like this:

```
-# For persisted session storage use the `redis-persistent` storage provided below.
+# If you don't need persistent sessions, you may use faster in-memory storage by replacing the redissession service definition above with this one:
```

bin/.platform/clear-cache.sh
andrerom: 1. `.env` does not contain $SYMFONY_ENV _(it was removed sometime last year afaik)_
2. `SYMFONY_ENV` is set in `.platform.app.yaml`'s evn section, so should be set afaik

kmadejski: @andrerom oh, right. Fixed it in https://github.com/ezsystems/ezplatform/pull/288/commits/4d619c8b2e79c14e5372b49e951f1f2d7539f547

.platform/services.yaml
```diff
@@ -21,15 +21,14 @@ mysqldb:
...
 
-# If you wish to have a separate Redis instance for sessions, uncomment
-# this service and the corresponding relationship in .platform.app.yaml.
-# For persisted session storage use the `redis-persistent` storage provided below.
-#redissession:
-#    type: 'redis:3.2'
-#
-
-# This configuration makes session storage persisted across server/redis restarts.
+# If you wish to have a separate Redis instance for sessions, uncomment this service and the corresponding relationship in .platform.app.yaml.
```
vidarl: This line still makes no sense as there is nothing below to uncomment !

.platform/services.yaml
```diff
@@ -21,13 +21,13 @@ mysqldb:
...
+# This configuration makes separated Redis session storage persisted across server/redis restarts.
+# Please, remember to have storage space to spare for this, and don't mind a bit slower instance type of redis.
+redissession:
+    type: redis-persistent:3.2
+    disk: 1024
+
+# If you don't need persistent sessions, you may use faster in-memory storage by replacing the redissession service definition above with this one:
 #redissession:
 #    type: 'redis:3.2'
 #
```
vidarl: Nitpick, this empty comment line can be removed

andrerom:
@Crell For the redis clearing part of this PR, we kind of need a way to flag/signify that this is needed for a deploy as we don't always want to do it given it can take notable toll on the performance of the system.

Main cases where we would need to clear this is on database resets _(restore, ..)_, and on upgrades where there is a chance cache structure might have changed.

Any input on existing conventions/concepts used for this in other projects?

bin/platformsh/clear-cache.sh
```diff
@@ -0,0 +1,25 @@
...
+    rm -Rf ../../app/cache/$SYMFONY_ENV/*.*
+    php ../../app/console cache:clear
+elif [ -d "../../app/cache/prod/$PLATFORM_TREE_ID" ] ; then
+    # Clear cache on re-deploy when the folder exits, move folder so post_deploy can cleanup
+    mv -f ../../app/cache/prod/$PLATFORM_TREE_ID ../../app/cache/prod/old_deploy
+fi
+
+clearRedisCache
+
+trap - EXIT
```
vidarl: The `set -e` and other traps are gone, so this line can be removed too

kmadejski: Thank you @vidarl for the review. Fixed all mentioned mistakes.

.platform.app.yaml
vidarl: This comment is still wrong

bin/platformsh/clear-cache.sh
```diff
@@ -0,0 +1,23 @@
...
+if [ "$SYMFONY_ENV" != "prod" ] ; then
+    # Clear class cache before we boot up symfony in case of interface changes on classes cached
+    rm -Rf ../../app/cache/$SYMFONY_ENV/*.*
+    php ../../app/console cache:clear
+elif [ -d "../../app/cache/prod/$PLATFORM_TREE_ID" ] ; then
+    # Clear cache on re-deploy when the folder exits, move folder so post_deploy can cleanup
+    mv -f ../../app/cache/prod/$PLATFORM_TREE_ID ../../app/cache/prod/old_deploy
+fi
+
+clearRedisCache
```
andrerom: Note: Open topic on how to only opt in to content related cache like redis when you want to avoid clearing all cache on every deploy _(not all deploys needs this performance hit)_.

Crell:
As discussed in Slack, I don't know of any existing conventions.  When a cache flush is "safe" is going to be very application specific.  If you have some way of tracking what doctrine migrations have already been run in order to see if there are any new ones to run, that is one possibility.  (New migrations probably mean need to clear the cache.)  I don't know EZ well enough to know what other triggers to check for.

bdunogier:
I have experienced issues with DB updates + persistence cache yesterday on p.sh. For what it's worth, running `php bin/console cache:pool:clear ezpublish.cache_pool` over SSH fixed them.

andrerom:
@Crell Returning to this, is it possible to detect a restore?

If so, that at least is a very clear case where we need to do this. And we can later expand to add more cases _(and with this merged partners can adapt to do this when they feel like it more applicable to their needs)_.

andrerom:
Not covering restore, but good input from @damz on handling schema changes
> For example, if you can compute a hash of the things that requires a full cache clear during build
> On deploy you compare that hash with a hash you store next to the cache
> And only flush the cache if the hash is different
> It can be as simple as "compute a hash of all the files that define my database schema" or something similar

So we could ship with something like that, but will need to allow partners to adapt the logic based on their needs.

Or if this gets too complex, we start clearing cache cache on every deploy and rather find a way to opt out when not needed.

andrerom:
I'm closing this as the solution here is rather to expose a symfony command to clear specific caches which is then able to clear only what is relevant _(avoiding to clear sessions for instance, or separate cache pool used for other things not needed to be cleared)_.

Details on how such a command can look like can be found here:
https://github.com/ezsystems/ezplatform-http-cache/pull/75

PR #289, opened 2018-05-03
Docker Redis Session Storage

raupie:
This conditional checks for two env variables that deal with redis storage configuration. If session storage is redis, use the configured redis server for session management.

### DIFF
--- a/app/config/env/generic.php
+++ b/app/config/env/generic.php
@@ -55,6 +55,13 @@ if (($pool = getenv('CACHE_POOL')) && file_exists(__DIR__ . "/../cache_pool/${po
     $loader->load($pool . '.yml');
 }
 
+// Session settings
+// If SESSION_STORAGE_HOST and SESSION_STORAGE_DSN env variable is set, use redis session storage
+if (($session_storage_host = getenv('SESSION_STORAGE_HOST')) && getenv('SESSION_STORAGE_DSN') && $session_storage_host == 'redis') {
+    $container->setParameter('ezplatform.session.handler_id', 'ezplatform.core.session.handler.native_redis');
+    $container->setParameter('session.save_path', getenv('SESSION_STORAGE_DSN'));
+}
+
 // Params that needs to be set at compile time and thus can't use Symfony's env()
 if ($purgeType = getenv('HTTPCACHE_PURGE_TYPE')) {
     $container->setParameter('purge_type', $purgeType);


raupie:
@andrerom any update on this? Will this make 2.2?

andrerom:
I'm kind of waiting for @vidarl input on this as he is the one that worked on redis sessions last.

app/config/env/generic.php
```diff
@@ -55,6 +55,13 @@
     $loader->load($pool . '.yml');
 }
 
+// Session settings
+// If SESSION_STORAGE_HOST and SESSION_STORAGE_DSN env variable is set, use redis session storage
+if (($session_storage_host = getenv('SESSION_STORAGE_HOST')) && getenv('SESSION_STORAGE_DSN') && $session_storage_host == 'redis') {
```
vidarl: I am not so keen on using value of $session_storage_host  to decide that session store is redis (ref. `&& $session_storage_host == 'redis'`)

I would think a third variable (`$SESSION_HANDLER`) or so would need to be added

raupie:
I'm following along with the documentation from ez and p.sh. When I set up the separate redis for sessions it failed. So the documentation here is incorrect. 

Resources:
https://docs.platform.sh/frameworks/ez.html
https://github.com/platformsh/platformsh-example-ezplatform/blob/demo/.platform/services.yaml

vidarl:
Let me get this right... you are trying to install master (not 1.13, 2,0 or 2.1) on Platform.sh with separate redis for sessions and that fails?

What happens, any error messages or anything?

It would be interesting if the [NativeSessionHandler.php](https://github.com/ezsystems/ezpublish-kernel/pull/2245/files#diff-bc067ed9ae749816c6f89c7653f29492) is actually used

raupie:
I originally mentioned this in slack. 

I installed a fresh eZ EE and my /admin on platform.sh is returning a `Fatal error: Uncaught RuntimeException: Failed to start the session in /app/ezplatform/vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/Session/Storage/NativeSessionStorage.php:146 Stack trace: #0` 

However now that I comment out my conditional from generic.php I can log into the admin and create content. I don't see any files in var/sessions when I'm in p.sh. So now I'm really freaked out. I did recently update my version of eZ.

andrerom:
The link @vidarl points to is a change you can find here: https://github.com/ezsystems/ezpublish-kernel/pull/2245

The issue this corresponds to is: [EZP-28800](https://jira.ez.no/browse/EZP-28800) Make it possible to store session in redis when using platform.sh

This change is part of 1.13.1 _(kernel: 6.13.1)_ and 2.1.0 _(kernel: 7.1.1)_, could that explain why it behaves different now?

raupie:
So I removed this from my code and re-deployed to p.sh and I'm no longer getting the error. No idea what fixed it.

andrerom:
Ok, thanks for confirming. Feel free to open issue/pr again if you experience anything.

Plopix:
I think we missed something here.

Now the default config is:

```yaml
framework:
    session:
        save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'

parameters:
    ezplatform.session.handler_id: session.handler.native_file
```
=> Meaning that by default it is ALWAYS file

If I want eZ to use the `php.ini` I have to change the default

```yaml
parameters:
    ezplatform.session.handler_id: ~
```
That works, it is fine, but that does not invalidate the logic of this PR.

eZ is looking to the ENV vars for `platform.sh` why not in the Docker context?

To me there is not reason to treat it differently and would avoid to touch anything we would have the same config locally and in prod using `ezplatform.core.session.handler.native_redis`

> I think the session storage should be configured in php.ini. Unfortunately that was not possible at Platform.sh

I agree @vidarl `php.ini` is better but more and more forgotten and making it globally ENV related would make more sense to me. And on Platform.sh it is possible it is just we don't know the HOST as it is dynamic... that could be the case with Docker too.

I think we need to counter part of the `platformsh.php` configuration in the `env.php` it would be consistent.

no? cc @andrerom @SylvainGuittard

Plopix:
@raupie can you reopen? thx!

vidarl:
Indeed, we have this in master : `ezplatform.session.handler_id: session.handler.native_file`

That does not sound right, and it is set to `~` in 1.13 : https://github.com/ezsystems/ezplatform/blob/1.13/app/config/default_parameters.yml#L4

@andrerom, Looks like a merge mistake ?

vidarl:
@raupie : I just tested redis session on Platform.sh and that seems to work as intended. 
My diff versus master:

```diff
diff --git a/.platform.app.yaml b/.platform.app.yaml
index 5b2aa87..c84f9fd 100644
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -24,7 +24,7 @@ relationships:
     rediscache: 'rediscache:redis'
     # If you wish to have a separate Redis instance for sessions, uncomment
     # this relationship and the corresponding service in .platform/services.yaml
-    #redissession: 'redissession:redis'
+    redissession: 'redissession:redis'
 
 variables:
     #php:
diff --git a/.platform/services.yaml b/.platform/services.yaml
index dfebfa6..468c3ba 100644
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -25,8 +25,8 @@ rediscache:
 
 # If you wish to have a separate Redis instance for sessions, uncomment
 # this service and the corresponding relationship in .platform.app.yaml.
-#redissession:
-#    type: 'redis:3.2'
+redissession:
+    type: 'redis:3.2'
 #
 # Alternatively if you have a requirement that sessions are persisted across server/redis restarts,
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
```

raupie:
soooo, my local dev is spitting out session files again in var/sessions/{dev,prod}. I'm trying to get the 2.2 update to work and in prod mode and having all sorts of issues. 

I can confirm that this PR fixes that. I put this back and when viewing the frontend and backend I'm not getting any session files.

raupie:
I've added 
```
- "SESSION_HANDLER_ID=ezplatform.core.session.handler.native_redis"
- "SESSION_SAVE_PATH=redis:6379"
```
to my docker-compose.yml

I can see entries in redis admin and in var/sessions/{dev,prod}

raupie:
Okay, I see in generic.php the check for SESSION_HANDLER_ID and SESSION_SAVE_PATH however in config.yml the framework.session.save_path: is set for `'%kernel.project_dir%/var/sessions/%kernel.environment%'` and not `'%session.save_path%'` so is this a bug?

raupie:
PR submitted. #307

PR #290, opened 2018-05-05
[Docker] Bump Varnish version to 6.0 for testing

andrerom:
~~Status: varnish-modules 0.14 is not uploaded to download section, and besides from the looks of it there needs to be a new release of it to get it working on Varnish 6.0.~~

~ref: https://github.com/varnish/varnish-modules/issues/113~

Varnish modules 0.15 is out and supports Varnish 6.0, so we can herby bump version _(assuming test pass)_.

### DIFF
--- a/doc/docker/Dockerfile-varnish
+++ b/doc/docker/Dockerfile-varnish
@@ -3,8 +3,8 @@ FROM debian:stretch-slim
 ENV VARNISH_MALLOC_SIZE="256M" \
     DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


andrerom:
@mnocon Yes it would, to avoid us introducing any breaks. Notice that this is not against master tough, it's against 1.13. ping if you need branch for master or 2.1 to test.

mnocon:
Ok, I've ported your changes to tmp branch in ezplatform-ee-demo:
https://github.com/ezsystems/ezplatform-ee-demo/pull/57
and made a PR in Studio relying on it:
https://github.com/ezsystems/StudioUIBundle/pull/905

mnocon:
It passed on 1.13 Studio, I don't think it's necessary to create more PRs to test - looks good to me.

PR #291, opened 2018-05-07
[Docker] Assuring empty cache in distribution image

jheba:
Environment variables were not taken into account due to pre-generated cache files.
Discovered while running SDC instances of 1.13 branch. Already fixed for 2.x branches (ref. #258)
Affected ezplatform and ezplatform ee

@vidarl and @andrerom for your acceptance


vidarl:
yes, that was a better way of explain it ;)

PR #293, opened 2018-05-13
[Platform.sh] Move cache to use /tmp instead of NFS to fix slowness on deploy and execution

andrerom:
~This is not solving issues with cache warmup taking a long time on NFS, however it makes sure it is done during deploy so incomming requests don't end up competing to create it.~

~So while this slows down deploy, in practice should make the app response quicker and more stable* once it start accepting requests again.~

Makes sure to warmup cache so requests coming in after deploy doesn't need to fight for it. Also does that to `/tmp` dir to make sure it can run fast as it needs to write many small files _(was taking 30-50s on NFS..)_. Afterwards we move the directory over to NFS mount.

The changes in AppKernel is backport from v2, we could also backport this for log directory, see: https://github.com/ezsystems/ezplatform/blob/master/app/AppKernel.php#L72-L93

\* _There have been reports on 503s happing on the first response being let in once site is back up, could be related all incomming requests needing to generate cache, but can also be related to custom stuff done on the given install._

Todo:
- [x] Set cache folder to /tmp during cache warmup and move files over in bulk once done
- [ ] Rebase and cleanup once varnish work has been merged

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -74,7 +74,7 @@ hooks:
             composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader
         fi
     # Deploy hook, access to services & done once (per cluster, not per node), only mounts are writable at this point
-    # Note: Http traffic is paused while this is running, so for prod code this should finish as fast as possible
+    # Note: Http traffic is paused while this is running, so for prod code this should finish as fast as possible, < 30s
     deploy: |
         set -e
         # Mainly relevant for eZ Platform demo usage, for own setup adapt this or remove and rely on migrations.
@@ -95,8 +95,9 @@ hooks:
                 mv -f app/cache/prod/$PLATFORM_TREE_ID app/cache/prod/old_deploy
             fi
             # Warmup cache so incomming requests does not need to end up doing that in paralell under load
-            # WARNING: ON NFS this can take quite a bit of time, from 20-50 sec depending of size of project
-            app/console cache:warmup
+            # We do this against /tmp and move it over afterwards in order to do this in under 10s instead of 30-50s
+            SYMFONY_TMP_DIR="/tmp" app/console cache:warmup; unset SYMFONY_TMP_DIR
+            mv /tmp/app/cache/prod/ app/cache/prod/$PLATFORM_TREE_ID
         fi
         # Example of additional deploy hooks if you use doctrine and/or kaliop migration bundle
         ##app/console doctrine:migrations:migrate --no-interaction --allow-no-migration
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -8,7 +8,7 @@
 # - PS: Platform.sh Professional
 
 # The name of this app. Must be unique within a project.
-name: app
+name: ez
 
 # The type of the application to build.
 type: php:7.2
@@ -39,8 +39,8 @@ variables:
         # For PE cluster, disable Symfony Proxy & instead use Fastly Bundle available on eZ Platform Enterprise
         #SYMFONY_HTTP_CACHE: 0
         # Place symfony cache on tmpfs to avoid performance issues of having it on NFS, re-deploy to clear cache reliably
-        # "app" comes from application name for platform.sh, so adapt it if you change the name for multi application setups
-        SYMFONY_CACHE_DIR: /tmp/app
+        # "ez" comes from application name for platform.sh, so adapt it if you change the name for multi application setups
+        SYMFONY_CACHE_DIR: /tmp/ez
         # PS: Given we just set cache directory to be non shared, we need to make sure http cache is shared (ignored when using Fastly)
         SYMFONY_HTTP_CACHE_DIR: app/cache/prod
         # Change this to dev if you want to run site & commands in dev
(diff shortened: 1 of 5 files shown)


andrerom:
FYI the `cache:warmup` takes 41 seconds on the given install _(large one)_, which in total takes 42-73 seconds depending on if migrations run.

So deploy time is still way to slow, and NFS is the main culprit. This just makes sure it is not done even slower, and potentially unstable, by several requests in parallel.

If we could configure local writable folder across Platform and Enterprise, _and_ had a way to be able to clear those mounts with platform cli command then that could be a solution to this.

.platform.app.yaml
```diff
@@ -85,10 +85,16 @@ hooks:
...
             rm -Rf app/cache/$SYMFONY_ENV/*.*
+            # Clear cache and warmup
             app/console cache:clear
-        elif [ -d "app/cache/prod/$PLATFORM_TREE_ID" ] ; then
-            # Clear cache on re-deploy when the folder exits, move folder so post_deploy can cleanup
-            mv -f app/cache/prod/$PLATFORM_TREE_ID app/cache/prod/old_deploy
+        else
+            # For prod we have a clean folder on each deploy, if re-deploying a sha we move & let post_deploy clean up
+            if [ -d "app/cache/prod/$PLATFORM_TREE_ID" ] ; then
+                mv -f app/cache/prod/$PLATFORM_TREE_ID app/cache/prod/old_deploy
```
Crell: A nice addition would be to add a post_deploy hook to rm old_deploy.  That way it doesn't block deploy but still clears out the old data.

andrerom: Indeed :) But look further down, we already do that, as it we delete all folders in prod which does not have same name as ´PLATFORM_TREE_ID´, this move is only needed for re-deploys.

Crell:
What exactly happens in cache warmup?  Are there any parts of it that could be done sans-DB, and so moved to the build hook?

andrerom:
@Crell It's not using database, so if all config was correctly setup (ref `$PLATFORM_RELATIONSHIPS`) we could have done it in build. This is why @antoligy (Alex) is asking for a `pre_deploy` hook, so we can do it there. In Symfony 4, probably around this time next year for FT I guess, would be able to warmup during build regardless of this.

@OriPekelman Suggested we could, if we have no soul, setup a ssh tunnel from build to prod to read in $PLATFORM_RELATIONSHIPS env for it. But besides the hack level of that, it would not take into account new settings on deploys as it will always read config from former deploy.

So request still stands for:
1. Getting rid of NFS usage for code cache, aka:
- A way to configure local writable mount across Professional and Enterprise
- At some point also a way to clear such mounts across several nodes via for instance platform cli command
2. `pre_deploy` _(with a big red warning in doc, warning people that prod is still running within same environment)_

_Setting local writable mount as prio 1 here, as this will benefit (eZ, @antoligy, and any other Symfony 2/3 users) also in regards to performance. That said, the root cause here is having to write cache to NFS, so one possible alternative could be to allow a per node step before deploy but after everything is setup, to write one last time to install dir before it's made read only._

andrerom:
Updated, added code to handle generating cache to /tmp during cache warmup.
The environment variable to set cache directory is back-ported from v2 so it's the same there _(but there also applicable for log directory)_.

.platform.app.yaml
```diff
@@ -93,8 +93,9 @@ hooks:
                 mv -f app/cache/prod/$PLATFORM_TREE_ID app/cache/prod/old_deploy
             fi
             # Warmup cache so incomming requests does not need to end up doing that in paralell under load
-            # WARNING: ON NFS this can take quite a bit of time, from 20-50 sec depending of size of project
-            app/console cache:warmup
+            # We do this against /tmp and move it over afterwards in order to do this in under 10s instead of 30-50s
+            SYMFONY_TMP_DIR="/tmp" app/console cache:warmup; unset SYMFONY_TMP_DIR
```
vidarl: It might be unsafe to create tmp files/directories in world writable locations without the use of [mktemp](https://www.gnu.org/software/autogen/mktemp.html)

However, no others should be able to write to /tmp so I guess it is fine to skip using mktemp in this case

andrerom: I guess we are pretty safe doing this here.

But if we where to use that in the future, how to use it for directly? I didn't get -d and -t to work together.

antoligy:
These changes seem to work okay, I have rolled them forward to an early environment and it seems okay.

antoligy:
That said the performance improvement isn't huge.

Crell:
@andrerom I am fairly certain that build having access to environment-dependent variables is never going to happen, even through a backdoor.  That would be a rather fundamental change to the entire Platform.sh architecture and eliminate the ability to reuse application slugs.

Reading the PLATFORM_RELATIONSHIPS array at runtime on each request is the only supported way to read that data.

andrerom:
@Crell Yup, aware of that. => Question for you to figure out with your team: Given we are talking per node cache here, do you or your team see any problems using /tmp for this? _(And continue to use NFS for logs and binary uploads and so on)._

antoligy:
I don't think that anything we can do here will get past the fact that even for 12-factor-applications, the ability to run some pre-requisite code prior to exposing an application available to the world is fairly essential.

Heroku's preboot is a great example of this, as are (the significantly more architecturally complex) rolling updates in Kubernetes using the readiness probe.
https://devcenter.heroku.com/articles/preboot
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-readiness-probes

A more Symfony specific example is how the Capistrano Symfony update task works.  The application cache is prewarmed _before_ the symlink is swapped and this happens in parallel to the application being made available.  Not as a blocker to it. https://github.com/capistrano/symfony/blob/master/lib/capistrano/tasks/deploy.rake

It feels a little bit naive to say that none of these architecturally distinct attempts to solve a common problem in release engineering, all of which would solve our actual problem, are not fit to be explored on the roadmap.

.platform.app.yaml
```diff
@@ -31,8 +34,12 @@ variables:
         # Example of setting php.ini config
         #"display_errors": "On"
     env:
-        # For enterprise cluster, disable Symfony Proxy & instead use Fastly Bundle available on eZ Platform Enterprise
+        # For PE cluster, disable Symfony Proxy & instead use Fastly Bundle available on eZ Platform Enterprise
         #SYMFONY_HTTP_CACHE: 0
+        # Place symfony cache on tmpfs to avoid performance issues of having it on NFS, re-deploy to clear cache reliably
+        SYMFONY_CACHE_DIR: /tmp/$USER
```
andrerom: Note: This probably don't work, but for PE we should somehow make sure platform.sh team adds something to keep document roots apart so they don't share cache.

andrerom:
Closing, this is already fixed in 1.13 and up recently after change to use Varnish for http cache and move var mounts to be local mounts once that was in place.

PR #298, opened 2018-06-17
[Docker] Move app Dockerfile into doc/docker like the rest of the files

andrerom:


andrerom:
MERGE NOTE: Needs to be done as a merge commit!!

PR #299, opened 2018-06-17
[Doc] Simplify installation doc further

andrerom:
Main simplifications:
- Simplify install doc so it fits within main README instead, so we can keep this in sync across repos a bit better _(ezplatform-demo v2 has both install file and instructions in readme for instance)_.

Small cleanup:
- Drop outdated ´GETTING_STARTED.md´, rather aim to have either link to instructions at end of install command, or default data in install pointing to _whats next_ on getting started developing on the system.


andrerom:
> They might miss that there is a repo for the demo. On the other hand, it's good because they won't start their projects using the demo :)

Added links to demo and ee in e1b77eb, better @SylvainGuittard ?

SylvainGuittard:
@andrerom Definitely better. Thanks :)

PR #301, opened 2018-06-21
EZP-29324: Upgrade to MariaDB 10.2 (+PHP 7.2) for eZ Platform Cloud

janit:
The MySQL column sizes limited installation to Platform.sh infrastructure with the previous setting of the database (MariaDB 10.1). This PR upgrades the MariaDB version to 10.2. Unrelated to this - It also bumps PHP to 7.2 by default (was 7.1) for improved performance and features.

NOTE: These changes should also be applied to the following repositories:

 - https://github.com/ezsystems/ezplatform-ee
 - https://github.com/ezsystems/ezplatform-ee-demo
 - https://github.com/ezsystems/ezplatform-demo

### DIFF
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -1,5 +1,6 @@
 mysqldb:
     type: mysql:10.2
+    # Version 10.1 might be a better option when running Legacy Bridge
     disk: 2048
     configuration:
         schemas:
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -1,6 +1,7 @@
 mysqldb:
     type: mysql:10.2
     # Version 10.1 might be a better option when running Legacy Bridge
+    # For more information see https://doc.ezplatform.com/en/2.1/getting_started/requirements_and_system_configuration/#supported-setups
     disk: 2048
     configuration:
         schemas:
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -1,5 +1,5 @@
 mysqldb:
-    type: mysql:10.1
+    type: mysql:10.2
     disk: 2048
     configuration:
         schemas:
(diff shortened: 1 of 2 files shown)


.platform/services.yaml
```diff
@@ -1,5 +1,5 @@
 mysqldb:
-    type: mysql:10.1
+    type: mysql:10.2
```
andrerom: maybe add a note here mentining it might be better to use 10.1 with legacy bridge setups, with link to requirements page which mentions it.

janit: I added a note on the version, but didn't find the requirements page.

andrerom: https://doc.ezplatform.com/en/2.1/getting_started/requirements_and_system_configuration/
```
* Note: Mysql 5.7 and MariaDB 10.2 changes how certain queries are parsed and is known to have issues with content attribute sorting queries in legacy because of that at the moment, MySQL 5.6 technically works but executes several hundred times slower on said queries. Because of this we overall recommend MariaDB 10.1 and 10.0, and don't offically support MySQL 5.6/5.7 and MariaDB 10.2 in use with legacy at the moment.
```

janit: Ok, I added the link. There is no anchor to that specific note, so you'll have to scroll down `¯\_(ツ)_/¯`

But another question: Should the PHP bump to 7.2 be considered regarding Legacy Bridge as well?

andrerom: > Another question: Should the PHP bump to 7.2 be considered regarding Legacy Bridge as well?

not heard any problems with 7.2 and legacy, have you?
at least leagcy bridge have tests running on 7.2 for v2 branch.

PR #303, opened 2018-06-29
[Composer] Remove conflict with Symfony 3.4.12

andrerom:
Issue has been solved in: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/tag/1.3.16

If travis agrees I guess we can merge this up.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
         "symfony/phpunit-bridge": "^3.4.11"
     },
     "conflict": {
-        "symfony/symfony": "3.4.9||3.4.12",
+        "symfony/symfony": "3.4.9",
         "doctrine/dbal": "2.7.0"
     },
     "scripts": {


andrerom:
Errors like:
```
eZ\Bundle\EzPublishRestBundle\Features\Context\RestContext::usePermissionsOfRole()
      eZ\Publish\Core\Base\Exceptions\UnauthorizedException: User does not have access to 'read' 'role' in vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/Repository/RoleService.php:784
```
So I think we hold back on allowing 3.4.12 until we know what this is and can solve it.

lserwatka:
We ship with 3.4.11. I don't want to risk.

emodric:
I'm also seing errors with `sudo` and 3.4.12, for example, `User does not have access to 'read' 'content' with: contentId '14''`

andrerom:
Closing we need 3.4.13 out for fix:
https://github.com/symfony/symfony/pull/27776

PR #307, opened 2018-07-13
Fix session save_path

raupie:
Currently the session save_path is set to use the filesystem by default. When trying to use redis the session save_path string would need to be modified. There is check for a env var in generic.php that sets the session.save_path variable however it isn't being used in the config.yml and continues to save to the filesystem instead of redis.


raupie:
@andrerom great idea. I will rename that.

andrerom:
@Plopix looks ok for you too?

Plopix:
@andrerom  did not test yet but it sounds ok. I will inject that new capabilities it offers into eZ Lauchpad asap. thx all

PR #308, opened 2018-07-18
Fix usage of session env variables and move some over to use env()

andrerom:
Some further cleanup on session env variables, and move some non handler based params to use `env()` and hence fix inline todo.

Besides cleanup, this fixes that platformsh.php was setting wrong session.save_path parameter.

This is followup to #307 (done in 2.2-dev) and #304
It also tries to clarify a bit the confusion in #306

review ping @alongosz, @raupie, @Plopix

### DIFF
--- a/app/config/default_parameters.yml
+++ b/app/config/default_parameters.yml
@@ -96,7 +96,7 @@ parameters:
     purge_type: local
 
     ## Session handler, by default set to file based in order to be able to use %session.save_path%
-    # evn: SESSION_HANDLER_ID
+    # env: SESSION_HANDLER_ID
     session.handler_id: session.handler.native_file
 
     # Admin siteaccess group name
--- a/app/config/default_parameters.yml
+++ b/app/config/default_parameters.yml
@@ -1,8 +1,6 @@
 # This file contains defaults for optional parameters, which you can override in parameters.yml
 parameters:
     locale_fallback: en
-    ezplatform.session.handler_id: session.handler.native_file
-    ezplatform.session.save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'
 
     # A secret key that's used to generate certain security-related tokens
     secret: '%env(SYMFONY_SECRET)%'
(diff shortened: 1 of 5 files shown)


app/config/default_parameters.yml
```diff
@@ -66,19 +72,32 @@ parameters:
...
+    # env: MAILER_TRANSPORT
     mailer_transport:  smtp
 
-    # Purge type used by HttpCache system ("local", "varnish"/"http", and on ee also "fastly")
+    ## Purge type used by HttpCache system ("local", "varnish"/"http", and on ee also "fastly")
+    # env: HTTPCACHE_PURGE_TYPE
     purge_type: local
 
+    ## Session handler, by default set to file based in order to be able to use %session.save_path%
+    # evn: SESSION_HANDLER_ID
```
raupie: type: # env

raupie:
Looks good. Everything is still working as expected. Using ezlaunchpad and still working.

app/config/default_parameters.yml
```diff
@@ -1,8 +1,6 @@
 # This file contains defaults for optional parameters, which you can override in parameters.yml
 parameters:
     locale_fallback: en
-    ezplatform.session.handler_id: session.handler.native_file
-    ezplatform.session.save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'
```
dew326: Isn't this a BC break? I had `ezplatform.session.save_path` overridden on my project and this change broke my installation.

andrerom: This was afaik introduced here and has not been made part of any release yet:
https://github.com/ezsystems/ezplatform/commit/fc0e356bca3c7dd03c813a40a976b359e5c0c31c


Or was it there somewhere else before?


However there is a  BC break here on `ezplatform.session.handler_id`, which has been around across several branches for quite a while. So the renaming here should maybe anyway be renamed back to also avoid your issue.

dew326: Maybe you are right and I did it between those two changes. If this wasn't released yet I think we are ok with the `ezplatform.session.save_path`.

andrerom: yeah, but I changed them to:
- `session.handler_id`
- `session.save_path`

to be somewhat more consistent with other params.

However change on first one is BC break as it's been around since 2.1.0 and 1.13.1.


So ok for you if I change it back to:
- `ezplatform.session.handler_id`
- `ezplatform.session.save_path`

?

andrerom: Fixed in 8b3d4e04818936937641c17dc3405751ac06acc1, looks ok to you? If so I'll merge up to the other repos

dew326: Looks ok, thanks.

PR #310, opened 2018-07-25
EZP-29455: Prepare regression setup with default configuration

mnocon:
JIRA: https://jira.ez.no/browse/EZP-29455

Added:
- notifications for Slack on broken builds.
- running PlatformUI tests on Varnish

Currently the build for 1.7 is failing because of Composer exceeding memory limits - I've increased it to 2GB (tested that 1GB is too low).

### DIFF
--- a/.env
+++ b/.env
@@ -4,6 +4,7 @@ COMPOSE_DIR=.
 
 # You'll need to adjust this for Windows and XDB Linux systems: https://getcomposer.org/doc/03-cli.md#composer-home
 COMPOSER_HOME=~/.composer
+COMPOSER_MEMORY_LIMIT=2G
 DATABASE_USER=ezp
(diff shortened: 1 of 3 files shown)


mnocon:
@andrerom there are some issues with current 1.7 build - it's exceeding composer memory limit and failing. Is the proposed way of setting it ok or do you have another suggestion?

andrerom:
> @andrerom there are some issues with current 1.7 build - it's exceeding composer memory limit and failing. Is the proposed way of setting it ok or do you have another suggestion?

@vidarl WDYT?

vidarl:
Using COMPOSER_MEMORY_LIMIT environment variable is fine by me

mnocon:
@lserwatka can we merge this one?

PR #313, opened 2018-08-01
[YML] Add default cache prefix in order to share cache across cluster 

andrerom:
Overrides the default cache namespace: https://github.com/symfony/symfony/blob/3.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/CachePoolPass.php#L33-L38

This makes sure cache is shared across a cluster no matter what the path of install is used or what Symfony Env you are using, hence making sure changes done on a dev env is affecting prod as well _(if those use different db's then inline doc here implies separate namespace should be used, or user should opt for default behavior instead)_

Open questions:
- Should this be master only maybe, as it affects cache logic for deployments and thus might be a bit intrusive to ship in a patch release

Todo:
- Doc, once approach is approved

-------
@emodric You'll see two first commits here with different approaches. In the end _(as of second commit)_ I think I agree with you, usage of cache seed prefix will:
1. affect other caches as well, so not good for mutli db installs as you tpyically _want_ to share symfony's internal app caches between the different db's
2. It will still use Symfony env as seed for namespace, meaning if some install is using several cache won't be properly cleared across envs. _(This is also the case when using file based regardless of this)_

Downside is cache service id is no longer automatically part of namespace, so multi db installs will need more adaption then before.

### DIFF
--- a/app/config/default_parameters.yml
+++ b/app/config/default_parameters.yml
@@ -36,8 +36,8 @@ parameters:
     # Cache DSN, see app/config/services/cache.yml for examples:
     cache_dsn: '%env(CACHE_DSN)%'
     
-    # Cache namespace prefix for use with Redis/Memcached
-    # For further info see: app/config/cache_pool/
+    # Cache namespace prefix for use with Redis/Memcached, 'ez' by default. 
+    # For further info incl alternatives for "blue/green deployment" and multi repo installs, see: app/config/cache_pool/*.yml
     cache_namespace: '%env(CACHE_NAMESPACE)%'
 
     # Settings for HttpCache
@@ -71,8 +71,8 @@ parameters:
     env(CACHE_POOL): "cache.app"
 
     env(CACHE_DSN): localhost
-    
-    env(CACHE_NAMESPACE): v1.mysite
+
+    env(CACHE_NAMESPACE): ez
 
     env(HTTPCACHE_PURGE_SERVER): "http://localhost:80"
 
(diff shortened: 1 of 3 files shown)


emodric:
I'm always for being explicit with configuration, so additional config for multi DB setups is something which is actually okay for me.

app/config/default_parameters.yml
```diff
@@ -67,6 +71,8 @@ parameters:
     env(CACHE_POOL): "cache.app"
 
     env(CACHE_DSN): localhost
+    
+    env(CACHE_NAMESPACE): v1.mysite
```
emodric: What about using an empty namespace by default? That is, an empty string (as opposed to `null`) ?

andrerom: I have no basis to answer that.
I did set it to `~` in first commit so it was using default logic out of the box, but not the same thing as empty string.

emodric: Using any string is fine, basically, as long as it is not `null`, since `null` falls back to default. Just asking because `v1.mysite` looks kinda weird :)

andrerom: yeah, that was an attempt at coming up with a default example.

kmadejski: As for me, empty string works well, but I can imagine that it might be confusing for others, therefore perhaps we can use for example `ez` here?

andrerom: @emodric WDYT?

emodric: `ez` is okay :+1:

PR #314, opened 2018-08-05
[SESSION] Configuration is not correct

Plopix:
This PR associated with:

Define that `php.ini` is the default configuration.

Then you can update that configuration through env. (current is `file` by default and that is not possible to override it (cause `null` value would NOT pass in the `if` in `generic.php`)

Also the kernel does not use `ezplatform.session.save_path` at all, then I am really confused.

It fixes also the save path for redis `tcp://`

> I did not really test live from a from scratch install

### DIFF
--- a/app/config/default_parameters.yml
+++ b/app/config/default_parameters.yml
@@ -43,9 +43,6 @@ parameters:
     # using EzSystemsPlatformHttpCacheBundle (default as of v1.12) which by design expires affected cache on changes
     httpcache_default_ttl: '%env(HTTPCACHE_DEFAULT_TTL)%'
 
-    # Session save path as used by symfony session handlers (eg. used for dsn with redis)
-    ezplatform.session.save_path: '%env(SESSION_SAVE_PATH)%'
-
     # Recommendation Bundle params
     ez_recommendation.default.yoochoose.customer_id: '%env(RECOMMENDATIONS_CUSTOMER_ID)%'
     ez_recommendation.default.yoochoose.license_key: '%env(RECOMMENDATIONS_LICENSE_KEY)%'
@@ -95,9 +92,12 @@ parameters:
     # env: HTTPCACHE_PURGE_TYPE
     purge_type: local
 
-    ## Session handler, by default set to file based (instead of ~) in order to be able to use %ezplatform.session.save_path%
-    # env: SESSION_HANDLER_ID
-    ezplatform.session.handler_id: session.handler.native_file
+    ## Session handler, by default relying on the php.ini
+    # env: SESSION_HANDLER_ID (to override it)
+    ezplatform.session.handler_id: ~
+    # Used when ezplatform.session.handler_id is not null
+    # env: SESSION_SAVE_PATH (to override it)
+    ezplatform.session.save_path: ~
 
     # Admin siteaccess group name
     admin_group_name: admin_group
--- a/app/config/env/platformsh.php
+++ b/app/config/env/platformsh.php
@@ -105,7 +105,7 @@ if (isset($relationships['redissession'])) {
         }
 
         $container->setParameter('ezplatform.session.handler_id', 'ezplatform.core.session.handler.native_redis');
-        $container->setParameter('ezplatform.session.save_path', sprintf('%s:%d', $endpoint['host'], $endpoint['port']));
+        $container->setParameter('ezplatform.session.save_path', sprintf('tcp://%s:%d', $endpoint['host'], $endpoint['port']));
     }
 } elseif (isset($relationships['rediscache'])) {
     foreach ($relationships['rediscache'] as $endpoint) {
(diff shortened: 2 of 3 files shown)


app/config/default_parameters.yml
```diff
@@ -95,9 +92,12 @@ parameters:
     # env: HTTPCACHE_PURGE_TYPE
     purge_type: local
 
-    ## Session handler, by default set to file based (instead of ~) in order to be able to use %ezplatform.session.save_path%
-    # env: SESSION_HANDLER_ID
-    ezplatform.session.handler_id: session.handler.native_file
```
andrerom: afaik this will break: https://github.com/symfony/symfony-standard/commit/070e53c45e435b23dc2c4b858e6d4f03ac796a05

andrerom:
> Also the kernel does not use ezplatform.session.save_path at all, then I am really confused.

kernel does not need to hard code usage of it, it's passed on to symfony here:https://github.com/ezsystems/ezplatform/blob/master/app/config/config.yml#L68-L69

app/config/default_parameters.yml
```diff
@@ -43,9 +43,6 @@ parameters:
     # using EzSystemsPlatformHttpCacheBundle (default as of v1.12) which by design expires affected cache on changes
     httpcache_default_ttl: '%env(HTTPCACHE_DEFAULT_TTL)%'
 
-    # Session save path as used by symfony session handlers (eg. used for dsn with redis)
-    ezplatform.session.save_path: '%env(SESSION_SAVE_PATH)%'
```
andrerom: why is this moved to generic.php?

app/config/default_parameters.yml
Plopix: then we have an issue because how would you override this config to "use the php.ini"?
In the link you mentioned they want to use the save_path setup in Symfony.

Here we want to let the php.ini handles the handler and the save_path.

It does not break to me, when both are `null` then php.ini will be used for both configuration.

app/config/default_parameters.yml
Plopix: to be consistent.

Plopix:
About your last comment, as mentioned here: https://github.com/ezsystems/ezpublish-kernel/pull/2410

The kernel *does NOT use* the `ezplatform.session.save_path` and it should be according to that:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml#L249

@andrerom, Am I completely mistaking here?

app/config/default_parameters.yml
andrerom: bBt that is in-consistent ;) Only params for handlers are done there now, as they can't be done at runtime.

andrerom:
> The kernel does NOT use the ezplatform.session.save_path and it should be according to that:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml#L249


The intention of this code is to get the param from Symfony framework config.

app/config/default_parameters.yml
andrerom: Could be a simple thing like this:
```
if (($value = getenv('SESSION_HANDLER_ID')) !== false) {
    $container->setParameter('ezplatform.session.handler_id', $value ?: null);
}
```

If all you need is to set it to null via env

Plopix:
Oh right, there is an automated injection of a parameter "session.save_path" by symfony: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L772
I did not know!

I closed the other PR, this one still has sense, right?

andrerom:
> I closed the other PR, this one still has sense, right?

Partly yes, but please get rid of the non bug related changes.

app/config/env/platformsh.php
```diff
@@ -105,7 +105,7 @@
         }
 
         $container->setParameter('ezplatform.session.handler_id', 'ezplatform.core.session.handler.native_redis');
-        $container->setParameter('ezplatform.session.save_path', sprintf('%s:%d', $endpoint['host'], $endpoint['port']));
+        $container->setParameter('ezplatform.session.save_path', sprintf('tcp://%s:%d', $endpoint['host'], $endpoint['port']));
```
vidarl: It works without 'tcp://' too right?
Cause it [won't work with memcached](http://php.net/manual/en/memcached.sessions.php) if we hardcode `tcp://`

andrerom: True, so maybe better to keep it as is then so we use same on SESSION_SAVE_PATH usage as on p.sh.

app/config/default_parameters.yml
andrerom: @Plopix Or:

```diff
diff --git a/app/config/env/generic.php b/app/config/env/generic.php
index 72ccf2be..7e4f69a7 100644
--- a/app/config/env/generic.php
+++ b/app/config/env/generic.php
@@ -69,5 +69,5 @@ if ($value = getenv('LOG_TYPE')) {
 }
 
 if ($value = getenv('SESSION_HANDLER_ID')) {
-    $container->setParameter('ezplatform.session.handler_id', $value);
+    $container->setParameter('ezplatform.session.handler_id', $value === '~' ? null : $value);
 }
```

PR #315, opened 2018-08-08
[Varnish] Configuration for Platform.sh

Plopix:
Enable Varnish by default on Platform.sh

Note: I still continue to think as mentioned here: https://github.com/ezsystems/ezplatform/pull/280

That SYMFONY_ENV set to `dev` on Platform.sh does not make any sense.

Unless proved wrong ;-)

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -25,14 +25,15 @@ relationships:
     # If you wish to have a separate Redis instance for sessions, uncomment
     # this relationship and the corresponding service in .platform/services.yaml
     #redissession: 'redissession:redis'
+    varnish: "varnish:http"
 
 variables:
     #php:
         # Example of setting php.ini config
         #"display_errors": "On"
     env:
-        # For enterprise cluster, disable Symfony Proxy & instead use Fastly Bundle available on eZ Platform Enterprise
-        #SYMFONY_HTTP_CACHE: 0
+        SYMFONY_HTTP_CACHE: 0
+        SYMFONY_TRUSTED_PROXIES: "127.0.0.1,localhost,varnish.internal"
         # Change this to dev if you want to run site & commands in dev
         SYMFONY_ENV: prod
         # Uncomment if you want to use DFS clustering:
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -33,3 +33,12 @@ rediscache:
 #redissession:
 #    type: redis-persistent:3.2
 #    disk: 1024
+
+varnish:
+    type: 'varnish:5.2'
+    relationships:
+        app: "app:http"
+    configuration:
+        vcl: !include
+            type: string
+            path: varnish.vcl.twig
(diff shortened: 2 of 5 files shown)


.platform.app.yaml
```diff
@@ -25,14 +25,15 @@ relationships:
...
 
 variables:
     #php:
         # Example of setting php.ini config
         #"display_errors": "On"
     env:
-        # For enterprise cluster, disable Symfony Proxy & instead use Fastly Bundle available on eZ Platform Enterprise
-        #SYMFONY_HTTP_CACHE: 0
+        SYMFONY_HTTP_CACHE: 0
+        SYMFONY_TRUSTED_PROXIES: "127.0.0.1,localhost,varnish.internal"
```
andrerom: host names don't really work or?

andrerom:
Looks good on first look, but Is this enabled on Platform.sh by default? If not we can not merge this.

.platform.app.yaml
Plopix: `SYMFONY_TRUSTED_PROXIES` according to Damien Platform.sh it is not required

.platform.app.yaml
```diff
@@ -25,14 +25,15 @@ relationships:
     # If you wish to have a separate Redis instance for sessions, uncomment
     # this relationship and the corresponding service in .platform/services.yaml
     #redissession: 'redissession:redis'
+    varnish: "varnish:http"
```
Plopix: remove that line to remove the circular reference

app/config/env/platformsh.php
```diff
@@ -117,3 +117,10 @@
         $container->setParameter('ezplatform.session.save_path', sprintf('%s:%d', $endpoint['host'], $endpoint['port']));
     }
 }
+
+if (isset($relationships['varnish'])) {
```
gauthiergarnier: If I understand correctly, we can't test this relationship since we are removing it to avoid circular reference issue.

Plopix: true, the purge server would be "public hostname" that the backend would reach using the token to purge

.platform/services.yaml
```diff
@@ -33,3 +33,12 @@ rediscache:
...
 #    disk: 1024
+
+varnish:
+    type: 'varnish:5.2'
+    relationships:
+        app: "app:http"
+    configuration:
+        vcl: !include
+            type: string
+            path: varnish.vcl.twig
```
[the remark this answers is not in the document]
andrerom: Maybe you should talk to @damianz5, template support was added at some point to be able to inject things. But as for why @Crell has not covered it in doc is more for him to reply on.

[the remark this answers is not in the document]
Crell: No.  The old config allows for Python Jinja2 templates only, and there was really only one thing you could do with it.  We're in the process of changing it for the actual release so that it's not templated at all, in any language, and we'll just auto-generate the part that would have been templated in the first place.

vidarl:
Closing in favour of https://github.com/ezsystems/ezplatform/pull/346

PR #317, opened 2018-08-09
EZP-29034: Cannot install ezplatform on 1.7 branch

glye:
> Fixes https://jira.ez.no/browse/EZP-29034

Since friendsofsymfony/http-cache version 1.4.5, eZ Platform 1.7 branch can't be installed or used unless one uses --no-warmup on cache clear commands, or clears caches with rm -rf. Class cache generated by the console conflicts with that from web.

This fixes it by:
- Cherry-picking https://github.com/ezsystems/ezplatform/commit/5fdf60c698760eda387286d5885ffe67ee32191f which ensures class cache is not used on PHP7, as it has no performance benefit there (Also: Disable bootstrap.php.cache on PHP 7 so we are in sync with how Symfony does it)
- Ensure that when cache is cleared by console command in PHP5, --no-warmup is enforced. https://github.com/ezsystems/ezplatform/commit/7e6b4716480d16e45d0ac4bd75b2b319cff42ff7 This is a filthy hack 💩 **UPDATE: The hack is now reverted and replaced by this cleaner approach** https://github.com/ezsystems/ezpublish-kernel/pull/2415

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -18,12 +18,12 @@ if (($useDebugging = getenv('SYMFONY_DEBUG')) === false || $useDebugging === '')
     $useDebugging = $environment === 'dev';
 }
 
-// Depending on SYMFONY_CLASSLOADER_FILE use custom class loader, otherwise use bootstrap cache, or autoload in debug
+// Depending on SYMFONY_CLASSLOADER_FILE use custom class loader, otherwise use bootstrap cache on PHP 5, or autoload in debug and PHP 7
 if ($loaderFile = getenv('SYMFONY_CLASSLOADER_FILE')) {
     require_once $loaderFile;
 } else {
     require_once __DIR__ . '/../app/autoload.php';
-    if (!$useDebugging) {
+    if (!$useDebugging && PHP_VERSION_ID < 70000) {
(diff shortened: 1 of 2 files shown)


andrerom:
You can also disable bootstrap.php.cache on PHP 7 so we are in sync with how Symfony does it: https://github.com/symfony/symfony-standard/blob/3.4/web/app.php#L6-L8

glye:
Update: Reverted dirty hack, replaced by compiler pass in kernel: https://github.com/ezsystems/ezpublish-kernel/pull/2415

web/app.php
```diff
@@ -18,12 +18,12 @@
     $useDebugging = $environment === 'dev';
 }
 
-// Depending on SYMFONY_CLASSLOADER_FILE use custom class loader, otherwise use bootstrap cache, or autoload in debug
+// Depending on SYMFONY_CLASSLOADER_FILE use custom class loader, otherwise use bootstrap cache on PHP 5, or autoload in debug and PHP 7
```
andrerom: `, otherwise use bootstrap cache on PHP 5 if not in debug mode`

PR #319, opened 2018-08-16
[Docker + p.sh] Switch to PHP 7.2 & change Dockerfiles to take version by ARG

andrerom:
PHP 7.2 should be rather stable now both on our side _(ezpublish-legacy just got patches as well)_, p.sh side and php side. So time to take advantage of the 5-7% speedup :)

As a minor bonus 7.2 image is also [38mb smaller](https://hub.docker.com/r/ezsystems/php/tags/) as it builds on Debian 9 _(:stretch-slim)_ instead of 8 _(:jessie)_.

This also adapts Dockerfiles to not hard code php image we use, in order to cleanup here and also get rid of things like https://github.com/ezsystems/docker-php/blob/master/bin/.travis/test.sh#L80-L86.

### DIFF
--- a/.env
+++ b/.env
@@ -10,8 +10,8 @@ DATABASE_PASSWORD=SetYourOwnPassword
 DATABASE_NAME=ezp
 
 ## Docker images (name and version)
-PHP_IMAGE=ezsystems/php:7.1-v1
-PHP_IMAGE_DEV=ezsystems/php:7.1-v1-dev
+PHP_IMAGE=ezsystems/php:7.2-v1
+PHP_IMAGE_DEV=ezsystems/php:7.2-v1-dev
 NGINX_IMAGE=nginx:stable
 MYSQL_IMAGE=healthcheck/mariadb
 SELENIUM_IMAGE=selenium/standalone-chrome-debug:3.4.0
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -8,7 +8,7 @@
 name: app
 
 # The type of the application to build.
-type: php:7.1
+type: php:7.2
 build:
     # "none" means we're running composer manually, see build hook
     flavor: "none"
--- a/doc/docker/Dockerfile-app
+++ b/doc/docker/Dockerfile-app
@@ -1,4 +1,5 @@
-FROM ezsystems/php:7.1-v1 as builder
+ARG PHP_IMAGE=ezsystems/php:7.2-v1
+FROM ${PHP_IMAGE} as builder
 
 # This is prod image (for dev use just mount your application as host volume into php image we extend here)
 ENV SYMFONY_ENV=prod
@@ -26,7 +27,7 @@ RUN rm -Rf app/logs/* app/cache/*/*
 RUN rm -rf web/css web/fonts web/js web/var
 
 
-FROM ezsystems/php:7.1-v1
+FROM ${PHP_IMAGE}
 
 # This is prod image (for dev use just mount your application as host volume into php image we extend here)
 ENV SYMFONY_ENV=prod
--- a/doc/docker/Dockerfile-nginx
+++ b/doc/docker/Dockerfile-nginx
@@ -1,4 +1,5 @@
-FROM ezsystems/php:7.1-v1 as web-build
+ARG PHP_IMAGE=ezsystems/php:7.2-v1
+FROM ${PHP_IMAGE} as web-build
 
 ENV SYMFONY_ENV=prod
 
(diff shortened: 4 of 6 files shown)


adamwojs:
Looks OK for me but the better choice for review will be @kmadejski / @mikadamczyk ;-)

mnocon:
Somehow, this change breaks our Travis setup. What I've discovered so far:

Travis builds now fail with:
```
  [RuntimeException]                                                           
  Failed to execute git clone --no-checkout '/root/.composer/cache/vcs/-var-w  
  ww-ezplatform-admin-ui/' '/var/www/vendor/ezsystems/ezplatform-admin-ui' --  
  dissociate --reference '/root/.composer/cache/vcs/-var-www-ezplatform-admin  
  -ui/' && cd '/var/www/vendor/ezsystems/ezplatform-admin-ui' && git remote s  
  et-url origin '/var/www/ezplatform-admin-ui' && git remote add composer '/v  
  ar/www/ezplatform-admin-ui'                                                  
                                                                               
  Cloning into '/var/www/vendor/ezsystems/ezplatform-admin-ui'...              
  fatal: reference repository '/root/.composer/cache/vcs/-var-www-ezplatform-  
  admin-ui/' is shallow                                                        
```

It happens only for non-meta repositories, which have to clone metarepo and add themselves as a dependancy. Metarepo builds are not affected.

The failure makes sense:
Travis by default clones to depth of 50 commits (https://docs.travis-ci.com/user/customizing-the-build/#git-clone-depth), 
we've even set this value to 30 for some repositories (Kernel: https://github.com/ezsystems/ezpublish-kernel/blob/master/.travis.yml#L94)

I've tried this scenario locally and it fails with the same message:
```
22208  cd test/
22209  git clone http://github.com/ezsystems/ezpublish-kernel.git --depth=1
22210  cd ezpublish-kernel/
22211  git log
22212  git status
22213  cd ..
22214  git clone --reference ~/Desktop/test/ezpublish-kernel/ http://github.com/ezsystems/ezpublish-kernel.git kernel2
```
(git version 2.17.0)

### Other information:

Everything started yesterday, about time when this PR was merged to 1.13, 2.2 and master (since then all tests started running on PHP 7.2)

1) AdminUI tests are failing (1.2 and master branches)
2) Kernel tests are mostly failing (7.2, master), but 6.13 is passing
3) Repository forms tests are passing on master
4) Page-builder tests are also failing (1.0 and master)

### My thoughts

1) I have no idea why it stopped working - or, to be more precise, why it used to work.

Both https://travis-ci.org/ezsystems/ezpublish-kernel/jobs/417219655 (passing) and https://travis-ci.org/ezsystems/ezpublish-kernel/jobs/418328012 (failing) use:
Composer: 1.7.2
Git: 2.15.1
Git clone depth: 30

I feel like it should've failed with the same error long time ago :( Unless there is bigger difference in the images except just for PHP version.

2) It's the PHP bump that causes it, what can be seen in PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/600

3) Repository-forms are passing on master, because they specify the previous PHP image directly: https://github.com/ezsystems/repository-forms/blob/master/.travis.yml#L21

I don't know why Kernel 6.13 passes.

### Possible solutions:

1) Revert this php docker image change until we know what's happening. Not a fan of this, but we should not tolerate broken CI setups (and this is probably the fastest way).

2) Set directly the PHP image used to 7.1 in all affected Travis setups. Not a fan of this (lots of PRs with changes that hopefully should be reverted), but also fixes CI.

3) Disable shallow cloning on all repositories and clear Travis cache. 
It works: https://github.com/ezsystems/ezplatform-page-builder/pull/158

BTW just clearing Travis cache is not enough, sadly.
It would allow us to benefit from PHP 7.2, but I'm a bit worried about the kernel builds (maybe the speed impact is not that big?)

4) Make an additional clone of the repository when setting up Travis.

So, instead of moving the currently cloned repo (by Travis) and reusing it in Composer, we could perform a full clone of the repository somewhere here: https://github.com/ezsystems/ezplatform/blob/master/bin/.travis/trusty/setup_ezplatform.sh#L64 and use that. This is the same as (3), but without specyfing it for all repositories which might be affected. It would require some `TRAVIS_PULL_REQUEST_SLUG` and `TRAVIS_BRANCH` magic, but should be possible.

@andrerom WDYT?

andrerom:
Hmm, in the past we called `git fetch --unshallow` before we used the repo (kernel, ui, ..) as source of the composer install. Is this still being done?

If not lets try to add it to the scripts setting things up, as it will avoid having to do full checkout for unit / integration tests, and also avoids us having specific logic around this in every repo.

mnocon:
Looks like it's not - I'll create a PR that adds it soon.

PR #321, opened 2018-08-22
[Travis] Perform full checkout to fix Travis setup

mnocon:
This is a followup to https://github.com/ezsystems/ezplatform/pull/319#issuecomment-414930052

I've tested the solution here: https://github.com/ezsystems/ezplatform-admin-ui/pull/604

First commit (without any changes in this PR) failed as expected.
Second commit (with changes from this PR) also failed (that was not expected).
After I've cleared the cache on Travis for AdminUI it passed.

So, it looks like in order to fix Travis we should merge this and then clear cache for all affected repositories (not a problem).

### DIFF
--- a/bin/.travis/trusty/setup_ezplatform.sh
+++ b/bin/.travis/trusty/setup_ezplatform.sh
@@ -64,9 +64,12 @@ if [[ -n "${DEPENDENCY_PACKAGE_NAME}" ]]; then
     BASE_PACKAGE_NAME=`basename ${DEPENDENCY_PACKAGE_NAME}`
     echo "> Move ${DEPENDENCY_PACKAGE_DIR} to ${EZPLATFORM_BUILD_DIR}/${BASE_PACKAGE_NAME}"
     mv ${DEPENDENCY_PACKAGE_DIR} ${EZPLATFORM_BUILD_DIR}/${BASE_PACKAGE_NAME}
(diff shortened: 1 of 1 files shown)


mnocon:
@andrerom do you think we can merge it, or should I request someone else to review?

andrerom:
I think we can ask also @alongosz to review, but as long as you have verified this fixes the issue as soon as we clear cache on the involved repos, then I think we are good for this change.

But I think we should also do what you proposed on the other repos.

As if, when we have this one kernel:
```
    - php: 7.1
      env: BEHAT_OPTS="--profile=rest --tags=~@broken --suite=fullJson" SYMFONY_ENV=behat
```
As you said we should add `PHP_IMAGE=ezsystems/php:7.1-v1 PHP_IMAGE_DEV=ezsystems/php:7.1-v1-dev`, as we should make sure both PHP on travis and the one inside containers is the same.

Besides that, I'm still a bit curious why Composer caches local repos, but as long as it works I guess that part is fine.

andrerom:
@mnocon merged to ezplatform:master and ezplatform-ee:2.2 + master.

mnocon:
Thanks! I've cleared the kernel cache and started a build: https://travis-ci.org/ezsystems/ezpublish-kernel/jobs/419161312

Fingers crossed 🤞

PR #323, opened 2018-09-06
[Travis] Improve PlatformUI stability, set COMPOSER_MEMORY_LIMIT

mnocon:
### Adding COMPOSER_MEMORY_LIMIT to app container: 

This is a followup to comment in https://github.com/ezsystems/ezplatform-demo/pull/92/files/b83035746dbf73caaa8d34edbab798814711ee15#diff-12e91abffee0d4e830c705c6d7258aa7R94

When developing tests across different repositories we're sometimes adding temporary Composer dependencies ([example of how it's done](https://gist.github.com/mnocon/c61372eba4f369f3e45cb9e962444fe5)). This operation performs composer operations directly on the `app` container, which does not use `COMPOSER_MEMORY_LIMIT` (it's only set for `install_app` - https://github.com/ezsystems/ezplatform/commit/add8d279f1d7b3b2666bb72935ff23c795bf812c#diff-12da65584b74987bec78f37c6b423dbfL23). For me it makes sense to set the value in `app` too.

### Rerun failed PlatformUI Behat tests

Sometimes the tests are failing, because there are concurrency issues when clearing cache after tests - all tests are passed, but because of issue that happened in `AfterScenario` phase build is marked as failed. I hope that using `travis_retry ... --rerun` will account for that and the issue won't occur again. This approach is used already in some StudioUI jobs, for example: https://github.com/ezsystems/StudioUIBundle/blob/1.7/.travis.yml#L73

~EDIT: it's marked as WIP, as I'm testing whether the failures are retried & reported correctly.~ It works as expected 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,12 @@ env:
     - SYMFONY_DEBUG=1
   # list of behat arguments to test
   matrix:
-    - TEST_CMD="bin/behat -vv --profile=rest --suite=fullJson --tags=~@broken" COMPOSE_FILE="doc/docker/base-dev.yml:doc/docker/varnish.yml:doc/docker/selenium.yml" WEB_HOST="varnish"
-    - TEST_CMD="bin/behat -vv --profile=rest --suite=fullXml --tags=~@broken"
-    - TEST_CMD="bin/behat -vv --profile=core --tags=~@broken"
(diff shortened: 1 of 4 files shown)


mnocon:
@andrerom as this is a follow-up to your comment - please have a look whether the changes (especially the Docker one) make sense - sometimes we're performing Composer operations in `app` directly, IMHO it's useful to set COMPOSER_MEMORY_LIMIT there.

mnocon:
@lserwatka can we merge this one?

PR #324, opened 2018-09-13
EZP-29608: eZ Platform Cloud certificate renewal

janit:
When deploying eZ Platform to eZ Platform Cloud (or plain Platform.sh), the system automatically deploys with SSL/TLS certificates from Let's Encrypt. By default there is no automatic renewal, which is done upon deployment to eZ Platform Cloud.

Platform.sh provides a solution to this: [Automatic certificate renewal](https://docs.platform.sh/configuration/routes/https.html#automatic-certificate-renewal). This is configured cron job section of the main configuration. The solution relies on periodical nightly deployments.

This PR includes a commented-out cronjob configuration blog. This is commented out, because otherwise this will enable automatic nightly deployments three times in a month, which might not be desired.


andrerom:
Seems like also @L0rdJ suggested something on this in #328 just now.

Any takes on what is preferred of these two takes?

L0rdJ:
You can not run automatic renewal without having Platform.sh CLI inside your instance.
That's why I would suggest #328 as it takes care of Platform.sh CLI (and it is not the case for #324)

janit:
I am fine with #328 as it seems to be more complete.

andrerom:
Closing in favor of #328

PR #326, opened 2018-09-18
Retting rid of "W: The minimum interval between cron runs is 5 minute…

SerheyDolgushev:
Getting rid of 
```
Validating configuration files
  W: The minimum interval between cron runs is 5 minutes, even if specified as less.
```

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -52,7 +52,7 @@ web:
             expires: 600
 
 # The size of the persistent disk of the application (in MB).
-disk: 2048
+disk: 3072
 
 # The mounts that will be performed when the package is deployed.
 mounts:
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -107,9 +107,11 @@ hooks:
 # see https://docs.platform.sh/configuration/app/cron.html#cron-jobs
 crons:
     frequent:
-        # https://docs.platform.sh/configuration/app/cron.html:
         # The minimum interval between cron runs is 5 minutes, even if specified as less.
-        spec: "*/5 * * * *"
+        # Except for PE. There crons can be run every minute.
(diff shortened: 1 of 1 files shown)


.platform.app.yaml
```diff
@@ -106,9 +106,10 @@ hooks:
 # The configuration of scheduled execution.
 # see https://docs.platform.sh/configuration/app/cron.html#cron-jobs
 crons:
-    minute:
-        # NOTE: Platform.sh PS does not execute every minute, so might sometimes miss jobs scheduled for a given time
-        spec: "* * * * *"
+    frequent:
+        # https://docs.platform.sh/configuration/app/cron.html:
+        # The minimum interval between cron runs is 5 minutes, even if specified as less.
+        spec: "*/5 * * * *"
```
andrerom: @vidarl On second thought, PE supports the previous value does it not?
So changing to every 5 minutes just risks defined jobs in cron command not running as it's not the right minute, also on PE. Right?

vidarl: Affirmative. On PE, crons may run every minute

vidarl: So, should we leave it as-is then ?  with "* * * * *", the cron will fire up every 5 minute on PS, and every minute on PE

andrerom: That was at least the original intention, we can add a more clear comment saying it can be changed to every 5 minutes on PS to avoid the warning.

SerheyDolgushev:
@vidarl @andrerom can you please review?

andrerom:
@SerheyDolgushev would you be able to rebase on  1.13?

andrerom:
Added it manually in https://github.com/ezsystems/ezplatform/commit/bf6b3cf5d789c6842f6d8edc3797f12326caeeae#diff-fe14053ff51cb358f8b709415b2a2d17

PR #327, opened 2018-09-18
Increase persistent disk size

SerheyDolgushev:
- [ ] Merge to 1.13

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -52,7 +52,7 @@ web:
             expires: 600
 
 # The size of the persistent disk of the application (in MB).
-disk: 2048
+disk: 3072
 
 # The mounts that will be performed when the package is deployed.
 mounts:
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -2,7 +2,7 @@ mysqldb:
     type: mysql:10.2
     # Version 10.1 might be a better option when running Legacy Bridge
     # For more information see https://doc.ezplatform.com/en/2.1/getting_started/requirements_and_system_configuration/#supported-setups
-    disk: 2048
+    disk: 1024
     configuration:
         schemas:
             - main
@@ -32,7 +32,7 @@ rediscache:
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
 #    type: redis-persistent:3.2
-#    disk: 1024
+#    disk: 512
 
 # If you wish to use solr, uncomment this service and the corresponding relationship in .platform.app.yaml.
 # Also, you need to generate the config using:


andrerom:
I'm ok with this, but any context here or any caveats to know about?

L0rdJ:
Platform.sh provides 5G storage space by default. And only 4G was used before in eZ Platform Cloud. I guess 1G was reserved for `redis`. But `redis` service is not using any disk storage, according to https://docs.platform.sh/configuration/services/redis.html:
> Data in an Ephemeral Redis instance is stored only in memory, and thus requires no disk space.

andrerom:
> But redis service is not using any disk storage, according to 

yes, but there has been talks of changing that for sessions: https://github.com/ezsystems/ezplatform/pull/288#issuecomment-385093683

L0rdJ:
By default `redis-persistent` service is disabled, and I am not sure we need to care about its storage.
If someone will decide to enable it, it is obvious that its resources (storage) need to be allocated. The same concept is applicable for solr: you need to allocate its storage before enabling it.

andrerom:
What I was saying is that we have been considering enabling redis-persistent, for sessions, by default, out of the box, in the config. To avoid some different issues

SerheyDolgushev:
I see that makes sense. But what about `mysqldb` service? Is there any reason why it uses `2G`? We are using `1G+` `mysqldb` services only on very big projects. Do you think we can resize it to `1G`?

andrerom:
For defaults I think that make sense. @Crell do you see any issues with that?

Crell:
Our default defaults are set to give users most of the available storage in the smallest plan size, plus some wiggle room to add Solr or similar.  There's nothing really magic about it.  The actual hard-floor on disk sizes is, I think, 128 MB.

So yeah, feel free to set the defaults for the eZ template to whatever makes sense for you, as long as they add up to less than 5 GB.

vidarl:
@Crell : What happends if you decrease from `disk: 2048` to `disk: 1024` while your actual disk usage is 1500 ? will deployment(git push) be refused? I kinda assume files won't be deleted at random...

Crell:
On any new project, decreasing the disk below the current usage will result in a 100% full disk of the size of the data (or rather, slightly more depending on what the exact numbers are).  No data loss, just a likely non-functional system.

NB: Some services (MariaDB in particular) respond very badly to their disk being unexpectedly full.

andrerom:
> So, how do we want this, really?

@vidarl For p.sh Pro Varnish is coming, so we would want to keep some space for that too as we will enable it by default.

@Crell Any words on when it comes _(officially, or at least reliably in terms of using it in out of the box config, so far it seems to have been in just beta in some regions)_, and how much space _(512 ok to start?)_ we need for it?

vidarl:
> For p.sh Pro Varnish is coming, so we would want to keep some space for that too as we will enable it by default.

@andrerom 
So, 512 for that ( , setting `disk : 2560` )?

andrerom:
> So, 512 for varnish ( setting disk : 2560 )?

For instance, if @Crell confirms

Crell:
Varnish is a disk-less service, as currently envisioned, like Redis.  Setting any disk on it at all would be rejected with an error.

andrerom:
> Varnish is a disk-less service, as currently envisioned, like Redis. Setting any disk on it at all would be rejected with an error.

ok, good to know

SerheyDolgushev:
Done: https://github.com/ezsystems/ezplatform/pull/327/commits/5c6208ef85c451689995d725cf8ba4af6d3648ac

andrerom:
@vidarl Is this ok to be cherry picked to 1.13 from your pov?

PR #328, opened 2018-09-18
Platform.sh crons

SerheyDolgushev:
- Installing Platform.sh CLI
- Daily Platform.sh snapshot
- Renew Platform.sh SSL certificates

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -41,9 +41,6 @@ variables:
         #PLATFORMSH_DFS_NFS_PATH: 'dfsdata'
         # Disable Platform.sh CLI automatic updates check (otherwise crons might stuck)
         PLATFORMSH_CLI_UPDATES_CHECK: 0
-        # API token (https://accounts.platform.sh/user/XXXXX/api-tokens)
-        # It would be more secure to set it as project variable in Platform.sh Web UI
-        PLATFORMSH_CLI_TOKEN: YOUR_TOKEN_GOES_HERE
 
 # The configuration of app when it is exposed to the web.
 web:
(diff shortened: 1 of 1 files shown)


andrerom:
Looks ok to me, but would like to hear what @vidarl and @Crell thinks here.

SerheyDolgushev:
Sorry, but for API token we are using sensitive project variable. It means that variable value is not available in Web UI/Paltform.sh CLI. The only place when it is possible to get the value of that variable is inside Platform.sh instance.
So it works pretty the same as `/app/.platformsh/.session/sess-cli-default/sess-cli-default.json`. And I we are proposing the most possible secure way to installing Platform.sh CLI inside Platform.sh instance.
If you want to increase security, you can create project-based Paltform.sh users and use their API tokens. So each API token will have access only to one project.
@Crell can you please provide some additional input?

Also `renew_ssl_certs` cron is now disabled on Enterprise projects.

.platform.app.yaml
```diff
@@ -39,6 +39,8 @@ variables:
         SYMFONY_ENV: prod
         # Uncomment if you want to use DFS clustering:
         #PLATFORMSH_DFS_NFS_PATH: 'dfsdata'
+        # Disable Platform.sh CLI automatic updates check (otherwise crons might stuck)
+        PLATFORMSH_CLI_UPDATES_CHECK: 0
```
Crell: Using `--yes` already disables the updates check, I believe.

Crell:
Installing the CLI on the application is a workaround, frankly.  I know we're working on built-in auto-snapshotting (no idea on the timeline at the moment), so eventually that will be unnecessary.  We want to do the same with Let's Encrypt, but there's some implementation challenges there (I don't know the details), which is why we haven't done it yet.

I'd recommend against building these into the repo.

andrerom:
> I'd recommend against building these into the repo.

ok. 
@SerheyDolgushev maybe we should rather go with approach like @janit proposed then, inline comments on how it can be done unofficially, and maybe also make people aware p.sh is working on a official feature to deal with this.

andrerom:
I'm with @vidarl  and -1 here I think, seems more to be something we could document instead _(ideally on platform.sh doc)_

PR #331, opened 2018-09-29
Platform.sh: install AWS CLI [only if AWS credentials are set]

SerheyDolgushev:
Default AWS region should be changed in `.aws/config`

### DIFF
--- a/.aws/config
+++ b/.aws/config
@@ -1,0 +1,3 @@
+[default]
+output = json
+region = us-east-1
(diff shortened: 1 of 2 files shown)


SerheyDolgushev:
@andrerom 
This change allows to install AWS CLI if credentials variables are set on Platform.sh. 
It should not have any affect on Azure, Google Cloud, etc.
But let's revert it If it seems redundant.

andrerom:
-1 from mys side, platform.sh is aiming to be agnostic to the service it runs on.

PR #332, opened 2018-10-19
Set solr_core from platform.sh relationships

raupie:
This fixes a bug for ezplatform platform.sh customers with enterprise subscriptions. Platform.sh creates solr_cores named after the environment and not from the platform.sh services.yml file. You need to set the solr_core variable dynamically from the solr relationship using the path entry.


vidarl:
Could you rebase this on 1.13 instead of master, please?

PR #334, opened 2018-10-24
Set solr_core variable from platform.sh relationships

raupie:
This fixes a bug for ezplatform platform.sh customers with enterprise subscriptions. Platform.sh creates solr_cores named after the environment and not from the platform.sh services.yml file. You need to set the solr_core variable dynamically from the solr relationship using the path entry.


raupie:
It would be convenient if the p.sh relationship for solr would include the core_name. Could p.sh do that?

Crell:
The correct approach is to always read the core name or DB name from the relationship.  It's an error to hard code that, pretty much always.  The "path" is the DB/core name (with `solr` prefixed to it).

In other words, this PR looks correct.

PR #335, opened 2018-10-25
Fixed bug in solr template for Platform.sh

vidarl:
Small bug in the solr template for Platform.sh

We need to point the `relationship` to the `endpoint` defined in [services.yml](https://github.com/ezsystems/ezplatform/blob/1.13/.platform/services.yaml#L50)

### DIFF
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -35,6 +35,8 @@ rediscache:
 # If you wish to use solr, uncomment this service and the corresponding relationship in .platform.app.yaml.
 # Also, you need to generate the config using:
 # vendor/ezsystems/ezplatform-solr-search-engine/bin/generate-solr-config.sh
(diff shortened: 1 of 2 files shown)


.platform/services.yaml
```diff
@@ -35,6 +35,8 @@ rediscache:
 # If you wish to use solr, uncomment this service and the corresponding relationship in .platform.app.yaml.
 # Also, you need to generate the config using:
 # vendor/ezsystems/ezplatform-solr-search-engine/bin/generate-solr-config.sh
+# Multi core setup is not supported on Platform.sh. Sharding does not work as the cores are unable to
```
andrerom: NOTE: Multi core setup is currently not supported on Platform.sh ...

vidarl: Ok, added "currently"

PR #338, opened 2018-11-13
[WIP] Increase Composer memory limit on Travis

mnocon:
The latest 1.7 builds fail with:
```
[33minstall_app_1  |[0m   - Installing ezsystems/ezpublish-kernel (6.7.x-dev 26402b8): DownloadingPHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 5966200 bytes) in phar:///usr/local/bin/composer/src/Composer/Util/RemoteFilesystem.php on line 490
[33minstall_app_1  |[0m 
[33minstall_app_1  |[0m Fatal error: Allowed memory size of 2147483648 bytes exhausted (tried to allocate 5966200 bytes) in phar:///usr/local/bin/composer/src/Composer/Util/RemoteFilesystem.php on line 490
[33minstall_app_1  |[0m 
[33minstall_app_1  |[0m Check https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors for more info on how to handle out of memory errors.Do not run Composer as root/super user! See https://getcomposer.org/root for details
[33minstall_app_1  |[0m Loading composer repositories with package information
[33minstall_app_1  |[0m Updating dependencies (including require-dev)
[33minstall_app_1  |[0m Package operations: 54 installs, 1 update, 0 removals
```

Examples:
https://travis-ci.org/ezsystems/ezplatform/builds/453686876
https://travis-ci.org/ezsystems/ezplatform/builds/454167429

Currently the limit is set to 2gb in https://github.com/ezsystems/ezplatform/blob/1.7/.env#L7

Testing whether increasing to 3gb solves this problem, but for me it's not the perfect solution - we can see that `composer install` requires more and more memory (it was increased to 2gb in July in https://github.com/ezsystems/ezplatform/pull/310), so maybe we should consider another strategy to stop this trend (not sure what causes it/how can we prevent it - can we set our dependencies in a different way to make Composer's job easier?)


andrerom:
One way to avoid it (temprary) is to reduce amount of packages composer checks against by bumping minimum requirements and narrowing the marching rules, see:
https://github.com/ezsystems/ezplatform/commit/f23bcef546c9c67f67047074f002a8a9abdfd05e#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

MarioBlazek:
One of the solutions would be to use `travis_wait` directive.

mnocon:
@andrerom changed approach as suggested - I've installed 1.7 branch (using PHP 5.6) and set the minimum required versions to what was installed. Setting Composer requirements is not something I do everyday, so careful review please 😉

If this is accepted - any suggestions how merge to ezplatform-ee should be handled? Should I create a PR with similar changes, including EE packages?

andrerom:
> If this is accepted - any suggestions how merge to ezplatform-ee should be handled? Should I create a PR with similar changes, including EE packages?

I can take care about the merging and handle that on merge up, it's a hell job so I'll not push that on someone else :P

composer.json
```diff
@@ -18,40 +18,40 @@
...
-        "incenteev/composer-parameter-handler": "~2.0",
-        "tedivm/stash-bundle": "~0.4",
-        "ezsystems/ezpublish-kernel": "~6.7.0@dev",
-        "ezsystems/repository-forms": "~1.5.0@dev",
+        "sensio/generator-bundle": "~2.5.3",
+        "incenteev/composer-parameter-handler": "~2.1.3",
+        "tedivm/stash-bundle": "~0.6.2",
+        "ezsystems/ezpublish-kernel": "~6.7.7@dev",
+        "ezsystems/repository-forms": "~1.5.5@dev",
         "ezsystems/ezplatform-solr-search-engine": "^1.1@dev",
```
andrerom: `^1.5` would be a safe change to make here

composer.json
```diff
@@ -18,40 +18,40 @@
...
-        "ezsystems/platform-ui-bundle": "~1.7.0@dev",
-        "ezsystems/ez-support-tools": "~0.1.0@dev",
-        "egulias/listeners-debug-command-bundle": "~1.9",
-        "white-october/pagerfanta-bundle": "1.0.*",
+        "ezsystems/platform-ui-bundle": "~1.7.8@dev",
+        "ezsystems/ez-support-tools": "~0.1.3@dev",
```
andrerom: you can bump this to `~0.2.1@dev` actually, we did a mistake and released 0.2 at some point forgetting poeple with `~0.1.x` won't get it. And only change is code formatting and a bug fix.

composer.json
```diff
@@ -18,40 +18,40 @@
...
+        "egulias/listeners-debug-command-bundle": "~1.9.1",
```
andrerom: This change means something else
`~1.9`: 1.x starting with 1.9.0
`~1.9.1`: 1.9.x starting with 1.9.1

You can change to `^1.9.1`, then it's same semantic meaning as before: 1.x starting with 1.9.1

composer.json
```diff
@@ -18,40 +18,40 @@
...
+        "white-october/pagerfanta-bundle": "~1.0.8",
         "hautelook/templated-uri-bundle": "~1.0 | ~2.0",
-        "doctrine/doctrine-bundle": "~1.3",
+        "doctrine/doctrine-bundle": "~1.9.1",
```
andrerom: same here `^1.9.1`

composer.json
```diff
@@ -18,40 +18,40 @@
...
         "symfony/expression-language": "~2.4",
```
andrerom: This can be `^2.8.47` afaik

composer.json
```diff
@@ -18,40 +18,40 @@
...
-        "sensio/framework-extra-bundle": "~3.0",
-        "ezsystems/content-on-the-fly-prototype": "~0.1.7",
+        "sensio/framework-extra-bundle": "~3.0.29",
```
andrerom: Doesn't make a difference here as there are no 3.1 releases and none planned, but semantically should have been `^3.0.29` if we wanted to retain the same meaning.

composer.json
```diff
@@ -18,40 +18,40 @@
...
-        "twig/extensions": "~1.0",
-        "symfony/assetic-bundle": "~2.3",
-        "symfony/swiftmailer-bundle": "~2.3",
-        "symfony/monolog-bundle": "~2.4",
+        "symfony/symfony": "~2.8.47",
+        "twig/extensions": "~1.5.2",
+        "symfony/assetic-bundle": "~2.8.2",
+        "symfony/swiftmailer-bundle": "~2.6.7",
+        "symfony/monolog-bundle": "~2.12.1",
         "sensio/distribution-bundle": "^3.0.36|^4.0.6|^5.0",
```
andrerom: `^3.0.36 | ^4.0.40 | ^5.0.22` from the other change on 1.13 and up

andrerom:
Review done, but control Question: Does it noticeably help on memory usage or not?

mnocon:
Changed as suggested. 

Before: it fails on my virtual machine that has total of 5gb ram (reaches 3598.1MB during composer install and fails with `proc_open(): fork failed - Cannot allocate memory`)
After: composer install on 1.7 branch reaches 2866.5MB memory maximum.

Also, thanks for handling the upstream merges 🎉 🥇

composer.json
mnocon: This one puzzles me, because it's not visible in `composer show` after installing 🤔

andrerom:
@lserwatka As we adjust almost all packages here, would you be ok if we enable package sorting as well to reduce conflicts in the future? _(packages will be sorted by name, and composer will do that for us on `composer require` calls)_

composer.json
```diff
@@ -18,40 +18,40 @@
...
+        "incenteev/composer-parameter-handler": "~2.1.3",
+        "tedivm/stash-bundle": "~0.6.2",
+        "ezsystems/ezpublish-kernel": "~6.7.7@dev",
+        "ezsystems/repository-forms": "~1.5.5@dev",
+        "ezsystems/ezplatform-solr-search-engine": "^1.5@dev",
+        "ezsystems/platform-ui-bundle": "~1.7.8@dev",
+        "ezsystems/ez-support-tools": "~0.2.1@dev",
+        "egulias/listeners-debug-command-bundle": "^1.9.1",
+        "white-october/pagerfanta-bundle": "~1.0.8",
         "hautelook/templated-uri-bundle": "~1.0 | ~2.0",
```
andrerom: `"~1.0 | ^2.1"`

andrerom:
> After: composer install on 1.7 branch reaches 2866.5MB memory maximum.

That is still a lot.. It feels like we are missing something here, as in that there is some / several indirect packages that we don't directly require in root package here, but which is required by the packages we ask for.

And that the version match for those indirect packages is huge and causes lots of versions to be checked against.

It's only a hypothesis, but would be good if we can find a way to try to get info from composer to indicate if this is the case or not.

mnocon:
How do we want to approach this? My main goal was to make Travis passing again, the fact that we've limited memory usage is a great "additional" win. 

I feel like a lot of time can be spent now on investigating indirect package requirements - on one hand it makes sense to make it all in one go (and do the merging dance only once), on the other hand this will take some time and the build will keep failing in the meantime (and I don't want anyone to start treating red builds as the norm). 

Would it make sense to extract this investigation part to another ticket and come back to it when the problem returns?

MarioBlazek:
@mnocon @andrerom maybe some optimisations should be done on ezpublish-kernel regarding version constraints.

andrerom:
@MarioBlazek We can, and should. But that is maybe more for master, otherwise we will need to have a BC / semantic versioning discussion with those thinking it would be a break to bump it on stable, or? :P

mnocon:
@andrerom what should I do now to move this forward?

andrerom:
Same here, 👍 
If someone wants to look into proposing narrowing down versions we match against on kernel as well that would also help _(To not include prior minor /patch versions not supported anymore on most packages. But no changes in regards to major versions unless otherwise discussed, or do only on master)_

Unless there are objections I plan to add package sorting after merge of this before merging up _(I'll do that in separate commit so Git manages to understand somewhat what is going on)_

andrerom:
In regards to kernel:

If someone wants to look into proposing narrowing down versions we match against on kernel as well that would also help _(To not include prior minor /patch versions not supported anymore on most packages. But no changes in regards to major versions unless otherwise discussed, or do only on master)_ but we can move forward on this one independently.

@MarioBlazek Side question on that, do NetGen still depend on Symfony 3 in Platform 1.x or can we drop that? _(we don't support it on 1.x and don't test it there so it would be one of the bigger things to "narrow down" in kernel)_

mnocon:
Thanks, I guess it can be merged then?

andrerom:
Yes, I'll handle that once I have a 1h windows ;)

emodric:
@andrerom As long as 3.4 is still allowed in `ezpublish-kernel`, we're okay with removing it here.

andrerom:
> @andrerom For ezpublish-kernel, keeping the latest 3.4 LTS is enough. We do have some projects which use 3.4 on 1.x (and I know of some others too) and it would be a shame to leave those projects without further bugfixes in 6.13.

ok, hope they plan to move to v2 and once you know they do give us a ping so we can drop this

PR #341, opened 2018-12-03
EZP-29852: Use security-checker instead of security-advisories

glye:
> https://jira.ez.no/browse/EZP-29852

Use [sensiolabs/security-checker](https://github.com/sensiolabs/security-checker/tree/v5.0.1) instead of [roave/security-advisories](https://github.com/Roave/SecurityAdvisories). security-advisories blocks installation in certain cases, leading people to remove it, thus losing protection in the cases where they could benefit from it. security-checker is less intrusive, showing a warning instead of blocking the install.

See Travis test output ("No packages have known vulnerabilities.") for how it looks.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
             "eZ\\Bundle\\EzPublishCoreBundle\\Composer\\ScriptHandler::dumpAssets",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
-            "vendor/sensiolabs/security-checker/security-checker security:check"
+            "@php bin/security-checker security:check"
         ],
         "post-update-cmd": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -59,15 +59,17 @@
...
         ],
         "post-update-cmd": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
             "eZ\\Bundle\\EzPublishCoreBundle\\Composer\\ScriptHandler::clearCache",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
             "eZ\\Bundle\\EzPublishCoreBundle\\Composer\\ScriptHandler::dumpAssets",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile"
+            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
+            "vendor/sensiolabs/security-checker/security-checker security:check"
```
andrerom: As this is a bin script we can instead do `@php bin/security-checker security:check` for instance

glye: @andrerom Right. The old composer (1.2.2 2016-11-03) on my debian doesn't understand the `@php`, but with a current one it works. If that's not ok, can we just say `bin/security-checker security:check`?
Anyway, I made the change.

PR #346, opened 2019-01-14
[Platform.sh] Added support for varnish and purge via token

vidarl:
Based on https://github.com/ezsystems/ezplatform/pull/315 which do not use token based purges

Depends on
 - ~https://github.com/ezsystems/ezplatform-http-cache/pull/74~
 - https://github.com/ezsystems/ezplatform-http-cache/pull/76

Due to this dependency, we should release new ezplatform-http-cache and require this new version in composer.json....

We cannot enabled token basen purge in docker by default due to :
 - https://github.com/ezsystems/docker-php/pull/34/files

*On merge*:
- Move forward on moving cache to use locale mount instead of shared, ref #293 _(needs simplifications to use locale mount)_
- Commit 1728baf31fceaf9aee237d0995fb8f8c9ff9eaff is only needed in 1.13


andrerom:
Can we drop the twig file?

vidarl:
> Can we drop the twig file?

Sure. fixed

web/app.php
```diff
@@ -67,7 +67,11 @@
 // If you are behind one or more trusted reverse proxies, you might want to set them in SYMFONY_TRUSTED_PROXIES environment
 // variable in order to get correct client IP
 if ($trustedProxies = getenv('SYMFONY_TRUSTED_PROXIES')) {
-    Request::setTrustedProxies(explode(',', $trustedProxies));
+    if ($trustedProxies === 'TRUST_ALL') {
```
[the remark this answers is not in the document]
gauthiergarnier: The example from the Symfony documentation looks a bit different, it's adding the local and remote IP in the TRUSTED_PROXIES: https://symfony.com/doc/2.8/deployment/proxies.html#but-what-if-the-ip-of-my-reverse-proxy-changes-constantly

vidarl:
Travis failures seems unrelated to this PR

web/app.php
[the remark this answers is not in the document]
andrerom: Not sure it's better, but in terms of "trust all" could be illegal netmask of `1.0.0.0/0` but.
But we don't specifically want to trust all IP's, we trust the immediate prox / remote, so maybe just rename to clarify that. `TRUST_REMOTE` or something?

web/app.php
```diff
@@ -67,7 +67,11 @@
...
+        Request::setTrustedProxies(array('127.0.0.1', $request->server->get('REMOTE_ADDR')));
```
andrerom: why did you need to set 127.0.0.1 again? Or is this from previous PR?
Afaik there where some issues in symfony and other places fixed by @kmadejski among others, but with latests symfony versions afaik this should normally not be needed.

kmadejski: That's correct. According to this PR: https://github.com/symfony/symfony/pull/26973 it should be no longer necessary to set `127.0.0.1` as a trusted proxy.

vidarl: Ok. I'll remove it.. I added 127.0.0.1 in the first place according to documentation : https://symfony.com/doc/current/deployment/proxies.html#but-what-if-the-ip-of-my-reverse-proxy-changes-constantly

web/app.php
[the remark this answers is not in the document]
vidarl: Okay, renamed to TRUST_REMOTE

vidarl:
Note that Commit 1728baf ("Fixed injecting of HTTPCACHE_VARNISH_INVALIDATE_TOKEN env variable in 1.13") is only needed in 1.13

.platform/services.yaml
```diff
@@ -51,3 +51,14 @@ rediscache:
 #        endpoints:
 #            collection1:
 #                core: collection1
+
+# Due to logic in app/config/env/platformsh.php, do not change the service name to something different than 'varnish'
+varnish:
+    type: 'varnish:5.2'
```
andrerom: Any reason why we don't opt for 6.0 (given we recommend using it) ?

vidarl: Varnish 6.0 on platform.sh does not support xkey at the moment

andrerom: ok, then we stay with this :)

web/app.php
```diff
@@ -67,7 +67,11 @@
 // If you are behind one or more trusted reverse proxies, you might want to set them in SYMFONY_TRUSTED_PROXIES environment
 // variable in order to get correct client IP
 if ($trustedProxies = getenv('SYMFONY_TRUSTED_PROXIES')) {
-    Request::setTrustedProxies(explode(',', $trustedProxies));
+    if ($trustedProxies === 'TRUST_REMOTE') {
+        Request::setTrustedProxies(array($request->server->get('REMOTE_ADDR')));
```
glye:
```suggestion
        Request::setTrustedProxies([$request->server->get('REMOTE_ADDR')]);
```
Nitpick: Old array syntax

app/config/env/platformsh.php
```diff
@@ -132,3 +134,28 @@
...
+    }
+}
+
+if ($route !== null) {
+    $container->setParameter('purge_type', 'varnish');
+    $container->setParameter('purge_server', rtrim($route, '/'));
+}
+
+// Setting default value for HTTPCACHE_VARNISH_INVALIDATE_TOKEN if it is not explicitly set
+if (!$varnishInvalidateToken = getenv('HTTPCACHE_VARNISH_INVALIDATE_TOKEN')) {
```
glye:
```suggestion
if (!getenv('HTTPCACHE_VARNISH_INVALIDATE_TOKEN')) {
```
No point in assigning to a variable you're not going to use, if I read this correctly.

PR #352, opened 2019-01-24
Increase shared memory size to avoid problems with chrome session

m-tyrala:
Egzample - [travis build](https://travis-ci.com/ezsystems/ezplatform-page-builder/jobs/172216914) - line 4381.

After verification on the occasion of the tests of [EZP-29419](https://jira.ez.no/browse/EZP-29419), it appears that this will help us.


m-tyrala:
@lserwatka ready to merge

PR #353, opened 2019-01-28
Increase shared memory to 1gb

mnocon:
Recently (in https://github.com/ezsystems/ezplatform/pull/352) the shared memory has been bumped to 512mb. There are fewer failures, but they still sometimes happen, for example:
https://travis-ci.com/ezsystems/ezplatform-form-builder/builds/98801861?utm_source=slack&utm_medium=notification

I believe increasing it to 1gb will make the tests stable, as right now the failures happen rarely.

(I've used format from https://docs.docker.com/compose/compose-file/#shm_size )


mnocon:

PR #355, opened 2019-01-28
Update composer.json

sunpietro:


lserwatka:
@micszo we need sanities on this PR. For entire Admin UI, Online Editor, Page Builder, etc.

PR #169, opened 2017-03-01
EZP-27060: Add platform.sh config

glye:
> https://jira.ez.no/browse/EZP-27060
> see also https://jira.ez.no/browse/EZS-1402
> Ready for review

Adds the necessary config files to run eZ Platform on platform.sh. This should also be added in the ezstudio and ezstudio-demo repos, see https://github.com/ezsystems/ezstudio/pull/43. Clustering not yet covered.

Note: The instructions in INSTALL.md are valid *after* this PR is merged. Before, in order to test it, you have to cherry pick the commits in this PR into the master branch of your fork, before attempting to make a platform.sh project out of it.

- [x] Add readme/installation doc in `doc/platformsh/`
- [x] ~~Switch to memcached once platform.sh has php-memcached 3.0.x available _(given it is what we officially recommend)_~~ Follow-up: https://jira.ez.no/browse/EZS-1413

PR #360, opened 2019-02-28
[CI/CD] Cleanup use of install command to use composer ezplatform-install everywhere

andrerom:
Changes all scripts (Docker, Platform.sh, Travis) to use `composer ezplatform-install` consistently to get rid of customizations per repo to specify install argument several places.

Needed to avoid further mess in https://github.com/ezsystems/ezcommerce-advanced/pull/6


.platform.app.yaml
```diff
@@ -81,8 +81,7 @@ hooks:
         set -e
         # Mainly relevant for eZ Platform demo usage, for own setup adapt this or remove and rely on migrations.
         if [ ! -f web/var/.platform.installed ]; then
-            . ./.env
-            php -d memory_limit=-1 bin/console ezplatform:install $INSTALL_EZ_INSTALL_TYPE
+            php -d memory_limit=-1 `which composer` ezplatform-install
```
andrerom: @vidarl just to check, do we really need to set `-d memory_limit=-1` when running installer?

andrerom: Edit, I see [blame of this](https://github.com/ezsystems/ezplatform/pull/169) is @glye

glye: Hm, I don't remember anymore if that was necessary or just adhering to tradition :)
Anyway, now we can set the php memory limit separately here in the same app file: https://docs.platform.sh/languages/php/ini.html
But uncertain if we can have separate settings for cli and sapi.

bin/.travis/prepare_ezpublish.sh
```diff
@@ -28,4 +28,4 @@ echo "> Run assetic dump for behat env"
 php app/console --env=behat --no-debug assetic:dump
 
 echo "> Installing ezplatform clean"
-php app/console --env=behat ezplatform:install clean
```
mnocon: Given that it still used app/console, meaning that it's not meant for v2 - can we remove this file?
If it is used, then the new approach will fail, as `ezplatform-install` was introduced in 2.2 or something like that (and won't be available anyway).

.platform.app.yaml
andrerom: Ok, seeing they have default of 128mb and the fact that this command will potentially triggering kaliop migrations on install..

... keeping this 😅 
Thanks for the input  @glye

bin/.travis/prepare_ezpublish.sh
andrerom: yes we can, most things in this folder was for travis `dist: precise`, so oldie stuff.

andrerom: done in 30dc6fc6c930204f9b3f531491e4182c2417ab53

PR #364, opened 2019-03-04
Added conflict to webpack-encore-bundle 1.2.0 and 1.2.1

dew326:
Until bug in webpack-encore-bundle is fixed we need avoid installing version 1.2 (https://github.com/symfony/webpack-encore-bundle/pull/51 or https://github.com/symfony/webpack-encore-bundle/pull/50)


lserwatka:
@dew326 could you merge it up?

PR #366, opened 2019-03-04
EZP-30208: Add ezsystems/ezplatform-graphql to meta repositories

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-30208

I opted for automatic installation of GraphiQL (nice UI for testing GraphQL queries) in `dev` environment.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "ezsystems/ezplatform-core": "^1.0@dev",
         "ezsystems/ezplatform-cron": "^2.0@dev",
         "ezsystems/ezplatform-design-engine": "^2.0@dev",
+        "ezsystems/ezplatform-graphql": "^1.0@dev",
         "ezsystems/ezplatform-http-cache": "~0.8@dev",
         "ezsystems/ezplatform-richtext": "^1.1@dev",
         "ezsystems/ezplatform-solr-search-engine": "^1.5.4@dev",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -34,6 +34,7 @@
         "ezsystems/ezplatform-core": "^1.0@dev",
         "ezsystems/ezplatform-cron": "^2.0@dev",
         "ezsystems/ezplatform-design-engine": "^2.0@dev",
+        "ezsystems/ezplatform-graphql": "^0.4.0@dev",
```
lserwatka: `^1.0@dev`

composer.json
```diff
@@ -89,7 +92,8 @@
             "yarn encore dev"
         ],
         "post-install-cmd": [
-            "@symfony-scripts"
+            "@symfony-scripts",
+            "@php bin/console ezplatform:graphql:generate-schema"
```
mnocon: Isn't it needed in `post-update-cmd` as well?
```
post-install-cmd: occurs after the install command has been executed with a lock file present.
```
it won't happen on stable branches, there is no lock there.

webhdx: I discussed it with @bdunogier. Since you only want to run this after making changes to the database (i.e. changing content types) I don't think there is a need to run it after updating packages. For people updating to 2.5 it should be mentioned in documentation that GraphQL schema has to be generated afterwards. cc @DominikaK

PR #369, opened 2019-03-11
EZP-30284: As a v2 User I want optimized cache for better performance

andrerom:
Story: https://jira.ez.no/browse/EZP-30284

Takes advantage of optimized File and Redis cache adapters provided by:
https://github.com/ezsystems/symfony-tools

Reduces cache lookups to 1/2 of what the plain symfony adapters does.
And as a bonus also gives more correct cache hit info with Symfony web debug toolbar.

Review note:
composer.json change is intentionally using `~1.0.0@dev`, see https://github.com/ezsystems/symfony-tools in regards to semantic versioning for Incubator features _(until accepted in Symfony 4 and they become plain backported features)_.


andrerom:
@kmadejski :
- Opted to not enable bundle from symfony-tools yet here, as it does not currently do anything. Should we for now remove bundle, or do you prefer we enable it here?
- In regards to `RedisSessionHandler` feature in symfony-tools I was thinking we can try to figure out how to enable that by default here as a follow up without blocking up when user does not have Redis

kmadejski:
> Opted to not enable bundle from symfony-tools yet here, as it does not currently do anything. Should we for now remove the bundle, or do you prefer we enable it here?

Sure, there is no reason for enabling bundle, at least not for now.  I think it's totally fine to skip it.

> In regards to RedisSessionHandler feature in symfony-tools I was thinking we can try to figure out how to enable that by default here as a follow up without blocking up when user does not have Redis

Can we introduce some sort of session storage handler factory in `symfony-tools` and let it decide which session storage handler should be used perhaps? In case of lack of Redis it could just fallback to standard filesystem session storage. I'll take a look on that as soon as possible.

andrerom:
> Can we introduce some sort of session storage handler factory in symfony-tools and let it decide which session storage handler should be used perhaps? In case of lack of Redis it could just fallback to standard filesystem session storage. I'll take a look on that as soon as possible.

ok _(and sounds lie then a case of use of the bundle :P)_

composer.json
```diff
@@ -40,9 +40,10 @@
         "ezsystems/ezplatform-richtext": "^1.1@dev",
         "ezsystems/ezplatform-solr-search-engine": "^1.6@dev",
         "ezsystems/ezplatform-standard-design": "^0.2@dev",
+        "ezsystems/ezplatform-user": "^1.0@dev",
         "ezsystems/ezpublish-kernel": "^7.5@dev",
         "ezsystems/repository-forms": "^2.5@dev",
-        "ezsystems/ezplatform-user": "^1.0@dev",
+        "ezsystems/symfony-tools": "~1.0.0@dev",
```
lserwatka: For consistency you should use `^1.0@dev` here. We change it as part of the release process.

andrerom: This is on purpose as this uses proposed features in Symfony which might change, see doc on the package:
https://github.com/ezsystems/symfony-tools#semantic-versioning-exception

composer.json
```diff
@@ -27,6 +27,7 @@
         "php": "^7.1",
         "doctrine/doctrine-bundle": "^1.9.1",
         "doctrine/orm": "^2.6.3",
+        "ezsystems/doctrine-dbal-schema": "^0.1@dev",
```
andrerom: @alongosz Moved this up here, as we use `"sort-packages": true,` now to simplify upgrades for users

composer.json
andrerom: NOTE: Risk here is that class name of the Adapters change.

But if we want to, we can create a deprecated alias class name extending the new one.

lserwatka: Let's leave it as it is. Better to be safe.

PR #370, opened 2019-03-13
EZP-30318: [DX] Simplified ezplatform.yml to avoid duplicated config

andrerom:
| Question           | Answer
| **JIRA issue** | [EZP-30318](https://jira.ez.no/browse/EZP-30318)
| **Resolves**     | <ul><li>[EZP-30305](https://jira.ez.no/browse/EZP-30305) (#373)</li><li>[EZP-29291](https://jira.ez.no/browse/EZP-29291)</li></ul>
| **Bug/Improvement**| yes
| **Target version** | `2.5 (master)`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | update for places where `admin_group` is referenced

v2 at some point ended up with lots of duplicate config due to the new admin siteaccess and it's group, however we can rather set `default` here to avoid that.

For demo this will mean we can simplify [from](https://github.com/ezsystems/ezplatform-demo/blob/master/app/config/ezplatform.yml#L42-L93) => to:
```yml
    system:
        default:
            # Cache pool service, needs to be different per repository (database) on multi repository install.
            cache_service_name: '%cache_pool%'
            # These reflect the current installers, complete installation before you change them. For changing var_dir
            # it is recommended to install clean, then change setting before you start adding binary content, otherwise you'll
            # need to manually modify your database data to reflect this to avoid exceptions.
            var_dir: var/site
            content:
                # As we by default enable EzSystemsPlatformHttpCacheBundle which is designed to expire all affected cache
                default_ttl: '%httpcache_default_ttl%'
            # HttpCache purge server(s) setting, eg Varnish, for when ezpublish.http_cache.purge_type is set to 'http'.
            http_cache:
                purge_servers: ['%purge_server%']

        admin_group:
            languages: [eng-GB, fre-FR, ger-DE, nor-NO]
            # For ee only
            page_builder:
                siteaccess_list: [site, fr, de, no]

        site_group:
            design: main

        fr:
            languages: [fre-FR, eng-GB]
        de:
            languages: [ger-DE, eng-GB]
        no:
            languages: [nor-NO, eng-GB]
```

Bonus: Makes people aware of possibility to set default config.

### DIFF
--- a/app/config/ezplatform.yml
+++ b/app/config/ezplatform.yml
@@ -48,7 +48,7 @@ ezpublish:
         #site_group:
         #    design: main
 
-        admin:
+        admin_group:
             # System languages. Note that by default, content, content types, and other data are in eng-GB locale,
             # so removing eng-GB from this list may lead to errors or content not being shown, unless you change
             # all eng-GB data to other locales first.
--- a/app/config/ezplatform.yml
+++ b/app/config/ezplatform.yml
@@ -26,7 +26,8 @@ ezpublish:
             URIElement: 1
 
     # System settings, grouped by default (global), siteaccess and/or siteaccess group
-    # TIP: For multi site install organize shared config in groups, with single site we can set shared config in default
+    # TIP: For multisite installations organize shared config into SiteAccess groups,
+    # for single-site you can place shared config under "default"
     system:
         default:
             # Cache pool service, needs to be different per repository (database) on multi repository install.
--- a/app/config/ezplatform.yml
+++ b/app/config/ezplatform.yml
@@ -29,7 +29,7 @@ ezpublish:
     # TIP: For multisite installations organize shared config into SiteAccess groups,
     #      for single-site you can place shared config under "default" and/or "global" for simplicity.
     system:
-        global:
+        default:
             # Cache pool service, needs to be different per repository (database) on multi repository install.
             cache_service_name: '%cache_pool%'
             # These reflect the current installers, complete installation before you change them. For changing var_dir
--- a/app/config/ezplatform.yml
+++ b/app/config/ezplatform.yml
@@ -25,11 +25,11 @@ ezpublish:
         match:
             URIElement: 1
 
-    # System settings, grouped by default (global), siteaccess and/or siteaccess group
+    # System settings, read in following order: `default`, <siteaccess-group>, <siteacces>, and `global`
     # TIP: For multisite installations organize shared config into SiteAccess groups,
-    # for single-site you can place shared config under "default"
+    #      for single-site you can place shared config under "default" and/or "global" for simplicity.
     system:
-        default:
+        global:
             # Cache pool service, needs to be different per repository (database) on multi repository install.
             cache_service_name: '%cache_pool%'
             # These reflect the current installers, complete installation before you change them. For changing var_dir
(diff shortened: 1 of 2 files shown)


app/config/ezplatform.yml
```diff
@@ -47,16 +48,8 @@ ezpublish:
                 purge_servers: ['%purge_server%']
                 varnish_invalidate_token: '%varnish_invalidate_token%'
 
-        # WARNING: Do not remove or rename this group.
```
andrerom: NOTE: it's not dangerous to remove this part, it's the part in `siteaccess:` which is important _(expand diff above)_.

lserwatka:
@mnocon this PR resolved problem with Bethat env. We should make QA on top of it and include this change in todays RC2.

andrerom:
👍 

@lserwatka Side: But could we also try to set aside time in the next cycle to strengthen SiteAccess/ConfigResolver system? There are too many strage issues in there now, and thye tend to buble up in support from time to time. 

We anyway need to change to quite a lot for 3.0 for Symfony 4 compatibility, so can maybe see if fixes for order rules / fallback system are applicable for backport too.

app/config/ezplatform.yml
```diff
@@ -25,9 +25,10 @@ ezpublish:
         match:
             URIElement: 1
 
-    # System settings, grouped by siteaccess and/or siteaccess group
+    # System settings, grouped by default (global), siteaccess and/or siteaccess group
+    # TIP: For multi site install organize shared config in groups, with single site we can set shared config in default
```
DominikaK:
```suggestion
    # TIP: For multisite installations organize shared config into SiteAccess groups,
    # for single-site you can place shared config under "default"
```

A suggestion to make the tip easier to read

mnocon:
@andrerom Travis is failing now, with screenshots like this one: https://res.cloudinary.com/ezplatformtravis/image/upload/v1552913021/screenshots/5c8f927d26b11656693384-vendor_ezsystems_ezplatform-admin-ui_features_contentcreation_feature_12_mociiu.png

lserwatka:
It uses wrong template.

andrerom:
Ok, let's see if this makes Travis happy going back to `default` _(but languages on group / SA)_. Using `global` seems to then trigger _another_ set of SA issues.

wizhippo:
> Ok, let's see if this makes Travis happy going back to `default` _(but languages on group / SA)_. Using `global` seems to then trigger _another_ set of SA issues 😞

What other problems?  Should they be addressed in another PR/issue?  Just curious as I recall other SA issues such as https://jira.ez.no/browse/EZP-28175, https://jira.ez.no/browse/EZP-29416 that one day I hope are addressed.  In a multi-site setup SA settings could be simplified too like you are doing here but because of some issues you end up having to duplicate settings.  Not the end of the world but DX wise inconvenient

Sorry if this is not relevant to this, it just made me think of the past issues.

andrerom:
> What other problems?

The language issue discovered above for instance: https://github.com/ezsystems/ezpublish-kernel/pull/2566
Which is related, but not completely solving your feature request: https://jira.ez.no/browse/EZP-28175, which we should definitely also do somehow.

> Should they be addressed in another PR/issue?

yes, we intend to iterate on this after this as well, so this is a followup actually it's covered by https://jira.ez.no/browse/EZP-29416

andrerom:
Merging to unblock rc2, additional rounds of testing will be done on the tag.

PR #371, opened 2019-03-15
[Platform.sh]Added support for injecting github token via environment variable

vidarl:
This PR makes it handy to let composer use a github token when communicating with github. A github token is needed if you are using private repos on github, or need to avoid githubs rate limit.

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -77,6 +77,13 @@ hooks:
     # Build hook, done before connected to services, disk is still writable here
     build: |
         set -e
+
+        # Inject GITHUB token if provided. This is handy if you have your own private repos, or need to avoid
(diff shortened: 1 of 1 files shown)


.platform.app.yaml
```diff
@@ -77,6 +77,13 @@ hooks:
     # Build hook, done before connected to services, disk is still writable here
     build: |
         set -e
+
+        # Inject GITHUB token if provided. This is handy if you have your own private repos, or need to avoid
+        # github's API rate limit
```
andrerom: Could maybe be shortend to something like:
`# Configure GITHUB token if set. For use if you have your own private repos, or to avoid API rate limits.`

vidarl: fixed

PR #374, opened 2019-03-18
[Platform.sh] Changed to 'old' solr config syntax

vidarl:
The syntax we use for solr config on platform.sh has been revoked by Platform, ref. https://github.com/platformsh/platformsh-docs/commit/f7f79d74223856f4d03a2b71954f1e865aef28ce#diff-3c2bd0d7625d58625747452f091e39af

So introducing the old "!archive" syntax. Without it, you'll simply won't be able to deploy with solr on Platform.sh

### DIFF
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -42,9 +42,7 @@ rediscache:
 #    disk: 512
 #    configuration:
 #        configsets:
-#            mainconfig: !include
-#                type: archive
-#                path: "configsets/solr6"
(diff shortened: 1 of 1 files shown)


andrerom:
As this fixes the syntax back to what it was some days ago, back to what it is in releases, I'm merging this.

blame: https://github.com/ezsystems/ezplatform/commit/3f21556889a665da036cbd62cc45b25e7e506256#diff-1ac4bf0767c87e430c34311c9ac36a56

PR #376, opened 2019-03-18
Update ezplatform-admin-ui-assets version for 2.4 in composer.json

micszo:
Q: Which is the better approach? "~3.2.0@dev" or "^3.1.0@dev"?

Discovered while testing https://github.com/ezsystems/ezplatform-admin-ui/pull/886.


lserwatka:
For stable branches and releases we are using first format.

lserwatka:
You can merge it up, but please make sure master uses `^4.1.0` tag

PR #380, opened 2019-03-28
[platform.sh] configure purge to local during the deploy phase

SylvainGuittard:
Thanks @andrerom for the fix.

jira ticket: https://jira.ez.no/browse/EZP-30370

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -120,10 +120,8 @@ hooks:
 
         # Example of additional deploy hooks if you use doctrine and/or kaliop migration bundle
         # During the deploy phase, varnish is not available. The purge type has to be configured to local
-        ##export HTTPCACHE_PURGE_TYPE="local"
         ##bin/console doctrine:migrations:migrate --no-interaction --allow-no-migration
         ##bin/console kaliop:migration:migrate --no-interaction --no-debug
-        ##unset HTTPCACHE_PURGE_TYPE
 
     # Post deploy hook, like deploy but after being deployed and live, for deploy tasks we can do asynchronously
     #post_deploy: |
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -106,8 +106,10 @@ hooks:
         set -e
         # Mainly relevant for eZ Platform demo usage, for own setup adapt this or remove and rely on migrations.
         if [ ! -f web/var/.platform.installed ]; then
+            export HTTPCACHE_PURGE_TYPE="local"
             php -d memory_limit=-1 `which composer` ezplatform-install
             touch web/var/.platform.installed
+            unset HTTPCACHE_PURGE_TYPE
         fi
 
         # Now that mounts are available, clear class/container cache. E.g. in case of interface changes on lazy proxies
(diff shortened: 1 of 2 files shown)


.platform.app.yaml
```diff
@@ -106,8 +106,10 @@ hooks:
         set -e
         # Mainly relevant for eZ Platform demo usage, for own setup adapt this or remove and rely on migrations.
         if [ ! -f web/var/.platform.installed ]; then
+            export HTTPCACHE_PURGE_TYPE="local"
             php -d memory_limit=-1 `which composer` ezplatform-install
             touch web/var/.platform.installed
+            unset HTTPCACHE_PURGE_TYPE
```
mnocon: If I set env `HTTPCACHE_PURGE_TYPE` to `fastly` when deploying to platform.sh - will it be lost at this stage?

andrerom: @vidarl Is this one meant to be set on deploy? 

@SylvainGuittard if so you'll need to backup the value and re-set it afterwards.

vidarl: I am a bit surprised that this approach even works, given that we first do:
https://github.com/ezsystems/ezplatform/blob/master/app/config/env/generic.php#L60
the later in the execution path do https://github.com/ezsystems/ezplatform/blob/master/app/config/env/platformsh.php#L152 

But if this works, original value must be reverted or this [cache clear](https://github.com/ezsystems/ezplatform/blob/master/.platform.app.yaml#L117 ) might create symfony cache where with varnish/fastly disabled
Okay, the `c:c` command is about clearing cache, but it will also warm parts of the symfony cache as a side-effect

vidarl: I fix should also be based on 1.13 maybe?

andrerom: This only works on new envs where local disk is empty, so for re-running anything with this you indeed need to clear container compile cache before and after.

But unsure if we should try to cover all edge cases here, proper solution is probably a way for installer to to tell eZ Platform to disable cache clearing and make sure to tell user he needs to run cache clear after install, maybe args to command to opt in / out of this.

vidarl: But at least revert to old value then, or varnish/fastly might be disabled after first-time install

andrerom: yes, and you might be right we need to clear cache before and after if @SylvainGuittard or others here expects to do install several times on same instance _(I know I need that when testing instances for instance)_.

andrerom: @vidarl Is it enough to skip `export`?

Or event just e.g `HTTPCACHE_PURGE_TYPE="local" && php -d memory_limit=-1 `which composer` ezplatform-install`

vidarl: No, but this should do the trick:

`HTTPCACHE_PURGE_TYPE=local php -d memory_limit=-1 `which composer` ezplatform-install`

andrerom: @vidarl Does it look ok now from your POV?


@SylvainGuittard I took the opportunity to propose some simplifications, given this is workaround for p.sh issue with Varnins we can skip spending lines on it in the example on kaliop migration bundle.

SylvainGuittard: Very good!
Thanks @andrerom

vidarl:
I gave my approval, but do not forget to rebase against 1.13 !

lserwatka:
@andrerom done

andrerom:
@SylvainGuittard / @damianz5 Does 1.13 also use Kaliop migration bundle? If not this is good as-si I think _(only going to 2.5+)_.

damianz5:
kaliop migration was added in >=2.0

andrerom:
Ok, then we are done here 🌮

SylvainGuittard:
Thanks everyone!

PR #381, opened 2019-03-28
Handle symlinks for itemToReplace in eZConfigManager.replace

SerheyDolgushev:
I'm using a symlink for `ezplatform-admin-ui` on my local. So `<EZ_ROOT>/vendor/ezsystems/ezplatform-admin-ui` links to `~/Projects/ez/bundles/ezplatform-admin-ui`.

Also I have local `ExtendedAdminUIBundle` bundle, and wanted to replace some items using `ez.config.manager.js`.
`src/ExtendedAdminUIBundle/Resources/encore/ez.config.manager.js`:
```
const path = require('path');

module.exports = (eZConfig, eZConfigManager) => {
    eZConfigManager.replace({
        eZConfig,
        entryName: 'ezplatform-admin-ui-alloyeditor-js',
        itemToReplace: path.resolve(__dirname, '../../../../vendor/ezsystems/ezplatform-admin-ui/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js'),
        newItem: path.resolve(__dirname, '../public/js/base-rich-text.js'),
    });
};
```

Without this fix I was unable to run `yarn encore dev` because of:
```
Error: Couldn't find item "<EZ_ROOT>/vendor/ezsystems/ezplatform-admin-ui/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js" in entry "ezplatform-admin-ui-alloyeditor-js".
```

It was happening because `ezplatform-admin-ui-alloyeditor-js` contains realpath, which is `~/Projects/ez/bundles/ezplatform-admin-ui/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js`.

### DIFF
--- a/ez.webpack.config.manager.js
+++ b/ez.webpack.config.manager.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const findItems = (eZConfig, entryName) => {
     const items = eZConfig.entry[entryName];
 
@@ -9,7 +10,7 @@ const findItems = (eZConfig, entryName) => {
 };
 const replace = ({ eZConfig, entryName, itemToReplace, newItem }) => {
     const items = findItems(eZConfig, entryName);
-    const indexToReplace = items.indexOf(itemToReplace);
+    const indexToReplace = items.indexOf(fs.realpathSync(itemToReplace));
 
     if (indexToReplace < 0) {
         throw new Error(`Couldn't find item "${itemToReplace}" in entry "${entryName}". Please check if there is a typo in the name.`);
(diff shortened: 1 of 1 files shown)


ez.webpack.config.manager.js
```diff
@@ -9,7 +10,7 @@ const findItems = (eZConfig, entryName) => {
 };
 const replace = ({ eZConfig, entryName, itemToReplace, newItem }) => {
     const items = findItems(eZConfig, entryName);
-    const indexToReplace = items.indexOf(itemToReplace);
+    const indexToReplace = items.indexOf(fs.realpathSync(itemToReplace));
```
dew326: Probably the same issue will be with remove method. Could you fix this also?

webhdx:
@SerheyDolgushev I added similar fix to `remove` method. We wanted to merge it right away :)

PR #384, opened 2019-04-08
Make tag command

glye:
Command for tagging the meta repo. Checks that various conditions and safeguards are met, and makes the tag if successful. Does not push the tag.

### DIFF
--- a/bin/.ci/make_tag.sh
+++ b/bin/.ci/make_tag.sh
@@ -10,7 +10,7 @@
 # - tag: v3.4.2
 # - composer args: arguments to pass to composer update (optional)
 
-UNSTAGED_CHANGES=`git st | grep 'Changes not staged for commit'`
+UNSTAGED_CHANGES=`git status | grep 'Changes not staged for commit'`
 if (( ${#UNSTAGED_CHANGES} > 0 )); then
   echo -e "\033[31m You have unstaged changes. Please commit or stash them. \033[0m"
   exit
--- a/bin/.ci/make_tag.sh
+++ b/bin/.ci/make_tag.sh
@@ -8,7 +8,7 @@
 #
 # Arguments:
 # - tag: v3.4.2
-# - composer args: arguments to pass to composer update
+# - composer args: arguments to pass to composer update (optional)
 
 set -e
 
--- a/bin/.ci/make_tag.sh
+++ b/bin/.ci/make_tag.sh
@@ -10,13 +10,13 @@
 # - tag: v3.4.2
 # - composer args: arguments to pass to composer update (optional)
 
-set -e
-
 UNSTAGED_CHANGES=`git st | grep 'Changes not staged for commit'`
-if [ ${#UNSTAGED_CHANGES} > 0 ]; then
+if (( ${#UNSTAGED_CHANGES} > 0 )); then
   echo -e "\033[31m You have unstaged changes. Please commit or stash them. \033[0m"
   exit
 fi
+
+set -e
 
 TAG=$1
 if [[ $TAG =~ ^v[0-9]+(\.[0-9]+){2,3}(-[a-z]+[0-9]*)?$ ]]; then
--- a/bin/.ci/make_tag.sh
+++ b/bin/.ci/make_tag.sh
@@ -13,10 +13,10 @@
 set -e
 
 TAG=$1
-if [[ $TAG =~ ^v[0-9]+(\.[0-9]+){2,3}$ ]]; then
+if [[ $TAG =~ ^v[0-9]+(\.[0-9]+){2,3}(-[a-z]+[0-9]*)?$ ]]; then
   echo -e "\033[36m Start work on making tag $TAG \033[0m"
 else
-  echo -e "\033[31m Tag argument did not look correct, should be v1.2.33 \033[0m"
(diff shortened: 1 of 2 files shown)


bin/.ci/make_tag.sh
```diff
@@ -0,0 +1,82 @@
...
+# Only aims to:
+# - prepare code for project use
+# - check some things to validate if we are ready for making tag
+# - make the tag
+#
+# Arguments:
+# - tag: v3.4.2
+# - composer args: arguments to pass to composer update (optional)
+
+UNSTAGED_CHANGES=`git st | grep 'Changes not staged for commit'`
```
andrerom: Alias detected, `git status` ? :)

bin/.ci/make_tag.sh
```diff
@@ -0,0 +1,82 @@
...
+  exit
+fi
+
+shift
+COMPOSER_ARGS="$@"
+CURRENT_BRANCH=`git branch | grep '*' | cut -d ' ' -f 2`
+
+# If we're in detached HEAD state
+if [[ $CURRENT_BRANCH == \(* ]]; then
+  CURRENT_BRANCH=`git branch | grep '*' | sed 's/.*detached at \([^)]\+\).*/\1/'`
```
andrerom: might need to double check if the output format has change between git versions. Think I have seen composer parsing this to and having to adjust for change that came in Git 2.x.

glye: Indeed: https://github.com/composer/composer/commit/313e6b914d351aea334224b9bc7403039ed57b41

glye: Fixed.

NB: I'm repeating the `CURRENT_BRANCH=${BASH_REMATCH[1]}` line on purpose, to allow different action/logging/output, if needed. Can be compacted I guess, but must be tested again for that.

bin/.ci/make_tag.sh
```diff
@@ -0,0 +1,92 @@
...
+git checkout -b "tmp_release_$TAG"
+
+echo -e "\033[36m Comment out *.lock files in .gitignore \033[0m"
+perl -pi -e 's/^(.*)\.lock$/#$1.lock/g' .gitignore
+
+minimumPHP=$(php -r '$hash = json_decode(file_get_contents("composer.json"), true); $php = str_replace(["^", "~"], "", $hash["require"]["php"]); echo explode("|", $php)[0];')
+echo -e "\033[36m Set minimum php version in composer.json (temporary to get vendor capable of working with it) to $minimumPHP \033[0m"
+composer config platform.php "$minimumPHP"
+
+# TODO: Check that ez packages (vendor whitelist?) don't use @dev
```
lserwatka: Shall we loop over `ezsystems/`? It is common mistake, that sometimes we left out `@dev` in composer.json

glye: Right, can be useful.

glye: Done in https://github.com/ezsystems/ezplatform/pull/384/commits/0a2952ae133d7d11cac905428dff8a391c70ac2c
(Sorry, that VM's clock is stuck in the past, will fix.)

glye:
Added variable for vendor prefix based on https://github.com/ezsystems/ez-support-tools/blob/a9feda20d71478ae8ebcbec676e398591c860b2f/SystemInfo/Collector/EzSystemInfoCollector.php#L89

andrerom:
@glye Wasn't this meant to be opened on 1.7 branch?

Otherwise we 1. won't fix this issue for 1.7/1.13 LTS, and 2. can't standardize on alway using this command to make meta tags.

bin/.ci/make_tag.sh
```diff
@@ -0,0 +1,105 @@
...
+  CURRENT_BRANCH=${BASH_REMATCH[1]}
+#* (no branch)
+else
+  echo -e "\033[31m Can't detect current branch/tag. \033[0m"
+  exit
+fi
+
+devPackageScanPhp='$hash = json_decode(file_get_contents("composer.json"), true); foreach($hash["require"] as $package => $version) { if(preg_match("'
+devPackageScanPhp+=$PACKAGE_WATCH_REGEX
+devPackageScanPhp+='", $package) === 1 && is_int(strpos($version, "@dev"))) { echo "  $package $version\n"; } }'
```
andrerom: Nit pick: I guess we can somehow use heredoc here to avoid the multiline string concating 
https://stackoverflow.com/questions/4937792/using-variables-inside-a-bash-heredoc

glye: Ok, I can try out this. Would be easier to read, esp. if the code grows later on.

glye: Done. We have to escape dollars in PHP variables (`\$var`) but other than that it's the same code.

glye: Would you like the same for the `minimumPHP` code?

andrerom: No that's ok, as it's only one line.

andrerom:
@vidarl / @lserwatka Any final remarks or approval?

lserwatka:
Looks good to me. Anyway 2.5.1 is unrealistic this week, still few important issues is under QA, so we push till next week.

PR #387, opened 2019-04-11
EZP-30451: Use VCL from ezplatform-http-cache in Travis testing

mnocon:
JIRA: https://jira.ez.no/browse/EZP-30451

As discussed previously (https://github.com/ezsystems/ezplatform/pull/349#issuecomment-459442840 , https://github.com/ezsystems/ezplatform-http-cache/pull/77#issuecomment-468615893) we should be using the VCL file from ezplatform-http-cache bundle when testing with Varnish.

This PR changes how our setup is installed on Travis:
before: start the containers and execute `composer install & composer ezplatform-install` inside them 
after: run `docker-compose -f doc/docker/install-dependencies.yml`, start the containers and install database.

`install-dependencies.yml` together with `install-database.yml` are reponsible for what `install.yml` used to do. Database install is moved after the containers are up, as this creates solr index.

I've decided NOT to remove the VCL files from this repository, I'll do it when master becomes future 3.0 (stable branches for 2.5 are created)

I've also tested that this still works for PRs in another repositories, where dependency is needed:
https://github.com/ezsystems/ezplatform-admin-ui/pull/973


.travis.yml
```diff
@@ -78,8 +78,6 @@ before_script:
   - echo "{\"github-oauth\":{\"github.com\":\"d0285ed5c8644f30547572ead2ed897431c1fc09\"}}" > auth.json
   # Setup eZ Platform inside docker container
   - /bin/bash ./bin/.travis/trusty/setup_ezplatform.sh "${COMPOSE_FILE}"
-  # Execute Symfony command if injected into test matrix
```
mnocon: I've deleted it because it's a part of install.yml: https://github.com/ezsystems/ezplatform/blob/master/doc/docker/install.yml#L39

andrerom:
Looks good 👍 

One thing probably missing here is handling of solr index, as we move back to two stage process where the first one is in charge of db install, use with solr will mean a empty index on boot.

Would it make sense to slim down "install", so it only is in charge of pulling in all code?
If so it would be more inline with platform.sh build stage.

mnocon:
It's back to WIP, as I'm trying to split install.yml into files - install-dependencies and install-database, so that people can get the current behaviour with them (and we can use install-dependencies only and perform DB install manually to take care of solr indexing).

mnocon:
@andrerom @vidarl @micszo @m-tyrala please have a look, as in an ideal world I'd like to finish this till the end of this week (I'm off next week).

.travis.yml
andrerom: is this till relevant or can we mark it as resolved?

mnocon: This comment is no longer valid, because it's the part of the (old) install.yml that requires database - as we're only installing Composer dependencies now we can't use this and we still need to use the old behaviour (SYMFONY_CMD in travis.yml)

andrerom:
@mnocon So.. Do we send to QA ? ;)

mnocon:
@andrerom I've tested that it installs correctly in both variants (`docker-compose -f doc/docker/install-dependencies.yml up --abort-on-container-exit` and `docker-compose -f doc/docker/install-dependencies.yml -f doc/docker/install-database.yml up --abort-on-container-exit`) on Travis. (Also tested with a PR in a dependant repository in https://github.com/ezsystems/ezplatform-admin-ui/pull/973 )

So I'd be in favour of merging it 😉 

One thing that worries me a bit: could there be any place (sales demo, something else) where we're using our Docker stack and we'll have to adjust for the change made here? I'm not aware of any, but as usual - it doesn't mean that such places do not exist.

mnocon:
I've went through the README again and the instructions are correct. So, I think we can merge this one.

mnocon:

PR #388, opened 2019-04-11
Enable support for GraphQL on eZ Platform Cloud / Platform.sh

janit:
eZ Platform v2.5 supports GraphQL by default. To enable support also on eZ Platform Cloud / Platform.sh the following changes are needed:

- Add mount: `app/config/graphql`
- Add automatic execution of schema generation: `php bin/console ezplatform:graphql:generate-schema`

Note that this will NOT enable the commonly used GraphiQL client in `prod`env mode. In fact, because of the ` --no-dev` flag used for composer, the GraphiQL package will not work even in `dev`env.


.platform.app.yaml
```diff
@@ -73,6 +73,10 @@ mounts:
     'web/var':
         source: local
         source_path: var
+    # GraphQL Schemas should be generated dynamically on deploy
+    'app/config/graphql':
```
vidarl: Honestly, I havn't looked at graphQL yet... but this folder is only written to by `ezplatform:graphql:generate-schema`, right?
if so, `ezplatform:graphql:generate-schema` should be moved to build hook, and this directory should not be a mount

vidarl: or are the other reasons for generating the schemas on deploy?

andrerom: It depends on data in database.

So strictly it should maybe be shared even, but as this is used for execution it's better if we stay with local disc and document it only gets updated on deploys.

vidarl: ok. then schema should be moved to var/......

bdunogier: Yes and no. The one that is always written to by the script is `app/config/graphql/ezplatform`. A project maintainer can add more files to `app/config/graphql`. So if one folder is a mount, it is the `graphql/ezplatform`, not `graphql`.

janit: I made the change that it is now only a suggested configuration that is commented by default (similar to Solr).

Dynamic schema creation is needed if you're working with the API only. There needs to be *some* way of generating schemas on the server, and the deploy is an easy way that at least it gets done on a deploy for sure. You could always log in via SSH and execute this as well, if you want to.

andrerom: > You could always log in via SSH and execute this as well, if you want to.

On PE cluster you'll need to login to all servers to generate on all the local mounts, but other then that yes.

vidarl: If I understand betrand correctly, `app/config/graphql/ezplatform` needs to be writeable?
so can `app/config/graphql/ezplatform` symlink to `var/graphql` ( or `var/graphql/ezplatform` )

I just don't fancy to put a writable mount below app/config ( or /app/ for that matter )

janit: > On PE cluster you'll need to login to all servers to generate on all the local mounts, but other then that yes.

True. Added a note on this.

janit: Do we need additional changes (removing gitignore + readme, etc.) or how will we proceed?

All in all, this is already progress once the PR gets indexed :)

andrerom:
@bdunogier Maybe you can take over on this? @janit is on mission, and me and @vidarl is on easter holliday.

.platform.app.yaml
bdunogier: Well, I wouldn't change the schema's location in general at this point for cloud, no. For on-premise cases, it is meant to be generated, committed and deployed. Having that directory in `.gitignore` is really a problem in that case.

Yes, the purely headless use-case would indeed require automatic & transparent updates live on the server. But for now, they're more of an edge case. What we could do maybe is make the root directory for the GraphQL schema configurable (it is built based on the root dir for now), and use that configuration directive for cloud ?

andrerom:
Thanks @bdunogier , from what I can gather then we should not adapt platform.sh config like this, but rather start to check in the schema file on `ezplatform-demo` and `ezplatform-ee-demo`, and make sure it is updated whenever there are content type changes in demo data (?)

/cc @damianz5 & @SylvainGuittard

bdunogier:
As for demo, @andrerom, if we consider it as a standard project (e.g. not the headless case where you never touch the config / templates), then yes, it is what we should do.

.platform.app.yaml
bdunogier: Regarding the schema dir and making it configurable, it would happen in this area: https://github.com/ezsystems/ezplatform-graphql/blob/v1.0.0/src/DependencyInjection/EzSystemsEzPlatformGraphQLExtension.php#L82. The path is based on `kernel.root_dir`, but we could add an option to change it to a totally different value, and use it for platform.sh. Would that work ?

bdunogier:
> we should not adapt platform.sh config like this, but rather start to check in the schema file on ezplatform-demo and ezplatform-ee-demo, and make sure it is updated whenever there are content type changes in demo data (?)

This would be a good option for the time being. There is a drawback though: it implies that we version the schema files, meaning that they would show up as modified when the script is executed. Alternatively, we could make that step part of the installer, and copy the schema files to the schema directory on install. As a bonus, we would be sure that the schema matches the install type, since it would come with it.

A short-term alternative is to make the schema files writable in `var`. It requires a few changes to the extension code, but it would work, I think.

The long-term solution is to make the GraphQL schema fully dynamic (e.g. updated whenever content types / config is modified). (Much) more complex.

PR #392, opened 2019-04-16
Rename default webpack.config.js to webpack.config.ezplatform.js

emodric:
Lots of our projects already have a `webpack.config.js` since we started to use Encore from the moment it was released. Upgrading to eZ Platform 2.5 (and future updates too), always generate git conflicts with existing `webpack.config.js`.

Because of that, I propose for eZ Platform to be a good citizen and not to takeover default config which is in most cases heavilly dependent on a project and its intricacies.

I also added two scripts to `package.json` which make it possible to build the assets with `yarn ezplatform:dev` and `yarn ezplatform:prod`.


Nattfarinn:
(`webpack.config.js` in a current form is already released in 2.5, keep in mind this is a BC break)

emodric:
I am aware that this is technically a BC break, but let me elaborate why I think it could be considered "not a BC break".

For new installs, obviously, nothing breaks.

When upgrading to lets say 2.5.1, you either have the file modified or don't. You also either upgrade by pulling with git from this repo, or upgrade "by hand" by comparing changes manually. So:

* file modified & upgrading via git - you will get conflicts that you need to resolve anyhow, so renaming the file is not an issue

* file modified & upgrading manually - you again need to touch the file to incorporate your changes to the updated file

* file not modified (or modified so no conflicts happen) & upgrading via git - no action needed, git will rename it for you

* file not modified & ugprading manually - this is the only case where you have to rename the file because it is renamed upstream, but since you need to compare the install anyhow and potentially touch other files, I don't see this as a problem.

Running `composer install` also will not break since the script in `bin/` now explicitly specifies which config to use.

Nattfarinn:
How different it is to what we have right now?

If you upgrade from `2.x` to `2.5.x` and you didn't use Encore/webpack before there's nothing to do.

If you did use Encore/webpack:
 * upgrading via git - you will get conflicts that you need to resolve
 * upgrading manually - you need to touch the file to incorporate your changes to the updated file

And all you have to do is copy & paste your own configuration to our `webpack.config.js` where `// Put your config here.` comment is.

Yet it keeps benefits of new extensibility points (extend Encore from bundle, override configuration and so on), you still only need one and only one command to compile assets (`yarn encore {env}`) and keep Symfony Encore experience for new installations (no need for external command wrappers like `composer ezplatform:dev`, not to mention not being able to use them with `--watch`).

lserwatka:
I agree with @Nattfarinn . You are doing this once when you upgrade.

emodric:
> If you upgrade from 2.x to 2.5.x and you didn't use Encore/webpack before there's nothing to do.

That *you have*. Not we, as in other developers who were using Encore for months now.

> And all you have to do is copy & paste your own configuration to our webpack.config.js where // Put your config here. comment is.

That here is the problem, there can be conflicts here between your webpack and project webpack config. Besides, we want to avoid constantly building eZ Platform Admin UI assets when developing a project.

> not to mention not being able to use them with --watch

Actually, they can be used. `yarn ezplatform:dev --watch` works fine.

Nattfarinn:
> Besides, we want to avoid constantly building eZ Platform Admin UI assets when developing a project.

Hmm. But we encourage to use our extensibility points and because of that we encourage you to compile it our way. It's not AdminUI related, but any bundle now can extend and overwrite assets.

But we do not force anyone to do that. In case of some tailored solution you can use `yarn encore dev --config=custom.webpack.config.js --watch` without forcing use of `--config` on everyone.

emodric:
> Hmm. But we encourage to use our extensibility points and because of that we encourage you to compile it our way. It's not AdminUI related, but any bundle now can extend and overwrite assets.

And that's fine. If someone wants to use your extensibility points, they can.

> But we do not force anyone to do that. In case of some tailored solution you can use yarn encore dev --config=custom.webpack.config.js --watch without forcing use of --config on everyone.

Sure, but I'd rather use `--config` for things which change rarely (like eZ Platform Admin UI assets), rather than for my daily workflow.

Nattfarinn:
I understand. But we have a feedback and people using it already seemed to enjoy it, so making a BC break with forcing them (and all current and future developers) to use different command on a daily basis doesn't sound fair.

There are ways how to avoid this, starting with one-time conflict resolve, or configuring different build process based on env in conflict resolved webpack.config.js, or having custom config in a first place. I'm sure there are other ways as well.

So while I understand it is not perfect for you and not very out-of-the-box in your workflow, I think considering there are ways to mitigate the issue to minimum, this PR doesn't seem like a generic solution in any way and will be cumbersome in a long term.

emodric:
I think that before rejecting this PR, community should have a say in it too. In forums, one community member already said he sees the value in configs being separated.

lserwatka:
We don’t have intention to maintain dedicated config or command. For consistency and what developers are used to in the Symfony ecosystem we will stick to current setup for Webpack Encore. Detailed explanation was given.

emodric:
That's too bad. Community would benefit from it  I'm sure.

lserwatka:
It would benefit more by solving this problem at its core. We should contribute to upsource, Webpack Encore, to solve fundamental lack of extensibility. It is very primitive tool at the moment. Not suitable for highly extensibly platform like ours.

emodric:
Webpack Encore is not the one responsible. It's a thin wrapper around Webpack. As for Webpack, it doesn't need extensibility. It's a framework, much as eZ Platform is. In my opinion, eZ Platform is the which should make sure not to antagonize its users by recognizing that, for example, someone other might already be using the file which eZ is trying to introduce and to at least try and play nice with existing projects.

PR #393, opened 2019-04-17
EZP-30479: Compile assets script is not working on some setups

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-30479
> Requires: https://github.com/ezsystems/ezplatform-core/pull/7

Replaces `bin/compile_assets.sh` with proper Script Handler that (hopefully) is multi platform.

PR #394, opened 2019-04-17
Move dumping javascript translations to Composer command for compiling assets

emodric:
Since compiling assets depends on javascript translations being present, this moves the dumping command from `composer.json` to Composer script.

This makes it possible to run the script by hand if needed (by running newly added script `composer ezplatform-compile-assets`), without risking translations not being present in compiled output, if e.g. web/assets folder was deleted by hand.

Depends on https://github.com/ezsystems/ezplatform-core/pull/8

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,7 +90,6 @@
             "eZ\\Bundle\\EzPublishCoreBundle\\Composer\\ScriptHandler::clearCache",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
-            "@php bin/console bazinga:js-translation:dump web/assets --merge-domains",
             "@php bin/console assetic:dump",
             "yarn install",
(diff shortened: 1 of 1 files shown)


emodric:
Obviously conflicts with #393, so I can update this PR when #393 is merged.

lserwatka:
@emodric this command does not belong into this context. We don't want to support execution both command while dumping assets. You should rather fix your build script and run dumping translation before generating assets. We will stick to two commands.

emodric:
There is no "my script". My point is that when you remove `web/assets` folder and want to regenerate assets manually by running `yarn encore prod`, you first need to dump the translations, or else admin UI will show translation placeholders instead of labels. This is not documented in any way and has bitten me several times in the last week.

If not by merging this, you should fix this by properly documenting the procedure then.

lserwatka:
Yes, I agree. We will update our documentation.

@ezsystems/documentation-team feels like we need to add more descriptive steps for generating assets when `web/assets`. Process must be clear about dumping translations first before generating assets by Wecpack Encore.

lserwatka:
Documentation updated via https://github.com/ezsystems/developer-documentation/pull/603

PR #395, opened 2019-04-17
Fix documentation about automatic behavior of SYMFONY_HTTP_CACHE

peterkeung:
I can see no code that conditionally sets SYMFONY_HTTP_CACHE based on SYMFONY_TRUSTED_PROXIES


benkmugo:
+1
I think the SYMFONY_HTTP_CACHE default behaviour line could be made clearer though

e.g. 
if empty + prod = enabled
if not set + prod = enabled
if set explictly 0|1 = disabled|enabled regardless of env

doc/apache2/vhost.template
```diff
@@ -56,8 +56,7 @@
     # Optional: Whether to use Symfony's builtin HTTP Caching Proxy.
     # Disable it if you are using an external reverse proxy (e.g. Varnish)
     # Possible values: 0, 1 or ""
-    # Defaults to disabled if SYMFONY_ENV is set to "dev" or SYMFONY_TRUSTED_PROXIES is set,
-    # and if this env value is omitted or empty
+    # Defaults to disabled if SYMFONY_ENV is set to "dev"
```
andrerom: Well the doc here was trying to explain this in `app.php`:
```
// Depending on the SYMFONY_HTTP_CACHE environment variable, tells whether the internal HTTP Cache mechanism is to be used.
// Recommendation is to use Varnish over this, for performance and being able to setup cluster if you need to.
// If not set, or "", it is auto activated if _not_ in "dev" environment.
if (($useHttpCache = getenv('SYMFONY_HTTP_CACHE')) === false || $useHttpCache === '') {
    $useHttpCache = $environment !== 'dev';
}
```

peterkeung: In reviewing what you just pasted, I think the suggested doc edit makes sense.

PR #396, opened 2019-04-25
EZP-30496: Enable source maps in the dev environment of eZ Platform

sunpietro:
https://jira.ez.no/browse/EZP-30496


andrerom:
@dew326 This is still relevant right? If so could you ask someone to do second review?

dew326:
After discussion with @tischsoic we agreed that we don't need this. It was an idea from Piotr to help him in the development. Also in the last commit, there is a BC break.
I will close this as obsolete.

PR #407, opened 2019-06-06
Update scssphp version, vendor&namespace

Cerdic:
Scssphp has moved https://github.com/scssphp/scssphp/issues/4


lserwatka:
Hi @Cerdic 

Thank you for your contribution. I think we should target this change for 2.5 branch. Could you change target branch?

Cerdic:
done! #408

PR #410, opened 2019-06-06
Use old Docker-Selenium container

mnocon:
All Travis build are currently failing, unable to login:

Example:
https://travis-ci.org/ezsystems/ezplatform/builds/542009341

This is caused by yesterday's release of Selenium Docker image, which upgraded Chrome version (see: https://github.com/SeleniumHQ/docker-selenium/issues/922 )

The workaround proposed in https://github.com/minkphp/MinkSelenium2Driver/issues/293#issuecomment-499458269 does not work for us, so the only option for me is keep using the latest working version until we figure this out.


mnocon:

lserwatka:
Could you merge it up and horizontally?

mnocon:
Yes, it's already merged to ezplatform and ezplatform-ee, now doing the demos

PR #412, opened 2019-06-11
EZP-30655: Package ezsystems/ezplatform-cron is missing from OSS

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-30655


webhdx:
Obviously Travis was failing because ezplatform-cron had not been merged yet. I restarted the build and it should confirm that's indeed working.

PR #413, opened 2019-06-11
EZP-30642: Added Travis dedicated error log

mnocon:
JIRA: https://jira.ez.no/browse/EZP-30642

Added a dedicated Travis log from which we will display errors on Travis (added in https://github.com/ezsystems/ezplatform-admin-ui/pull/1012/files)

### DIFF
--- a/app/config/config_behat.yml
+++ b/app/config/config_behat.yml
@@ -17,3 +17,10 @@ swiftmailer:
 
 assetic:
     use_controller: false
+
+monolog:
+    handlers:
+        travis:
+            type: stream
+            path: "%kernel.logs_dir%/travis_test.log"
+            level: error


app/config/config_behat.yml
```diff
@@ -17,3 +17,10 @@ swiftmailer:
 
 assetic:
     use_controller: false
+
+monolog:
+    handlers:
+        travis:
+            type: stream
+            path: "%kernel.logs_dir%/travis_test.log"
```
micszo: Would it be helpful to have a timestamp in the log file name? (if that's possible)

mnocon: I don't think it's necessary. For Travis there will be a new file for each job (so no old entries) and the entries themselves have the date specified next to them (see in AdminUI PR). This is how it will look like:
```
  │  Application errors:
    │  [2019-06-11 11:59:17] ERROR: <error message>
    │
```

PR #417, opened 2019-06-21
Add support for Chromium driver

mnocon:
JIRA: https://jira.ez.no/browse/EZP-30669

Related PR: https://github.com/ezsystems/docker-php/pull/39

This PR adds support for Chromium driver in our Docker setup and in Behat/Mink configuration. It will be used as a secondary driver, as Selenium driver does not support response header checking (and we want to use that for some tests in http-cache package).

Looks like Google does not provide any official Docker images with Chromium (unlike Selenium). I've used the image provided by the creator of Chrome integration extension for Mink and Behat.

I've also compared the execution time and potential cost of switching from Selenium to Chrome.

Results:
1) Tests are failing because of performance issues:
```
PHP Notice:  fwrite(): send of 57 bytes failed with errno=32 Broken pipe in /var/www/vendor/textalk/websocket/lib/Base.php on line 246
In Base.php line 249:
                                       
  [WebSocket\ConnectionException]      
  Could only write 0 out of 57 bytes.  
```

This is usually solved by increasing SHM size, as suggested here: https://gitlab.com/DMore/chrome-mink-driver/issues/50

But the Chrome in Docker is running with SHM usage disabled - https://gitlab.com/DMore/docker-chrome-headless/blob/7.1/files/supervisord.conf#L47 (meaning it uses tmp, see https://github.com/GoogleChrome/puppeteer/issues/1834 - it should be somehow improved)

2) Tests are failing because of drag and drop issues:
```
        Behat\Mink\Exception\DriverException: SyntaxError: missing ) after argument list in vendor/dmore/chrome-mink-driver/src/ChromeDriver.php:1000
        Stack trace:
        #0 vendor/dmore/chrome-mink-driver/src/ChromeDriver.php(976): DMore\ChromeDriver\ChromeDriver->evaluateScript('dragMock.dragSt...')
        #1 vendor/ezsystems/ezplatform-admin-ui/src/lib/Behat/Helper/UtilityContext.php(307): DMore\ChromeDriver\ChromeDriver->executeScript('dragMock.dragSt...')
```

Drag and drops are executed using a JS library (https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/Behat/lib/drag-mock.js) because Selenium does not support drag and drops between iframes (which is the case for FB and PB).

3) Problems with file uploading:
```
Fatal error: Call to undefined method DMore\ChromeDriver\ChromeDriver::getWebDriverSession() in vendor/ezsystems/ezplatform-admin-ui/src/lib/Behat/Helper/UtilityContext.php:292
```

I haven't investigated it deeply, but I suppose Chromium does have support for file uploads as well 😉 

4) Problems with Roles:
```
    And I go to "Test Role edited" "Role" page                                   # EzSystems\EzPlatformAdminUi\Behat\BusinessContext\AdministrationContext::iGoToListItem()
    When I start assigning users and groups to "Test Role edited" from role page # EzSystems\EzPlatformAdminUi\Behat\BusinessContext\RolesContext::iStartAssigningTo()
      Tag matching css ".ez-table-header [title^=Assign]" not found.
```

I suppose there is an issue with the selector, it shoudn't be hard to change it.

Given problems 1 and 2 there is no way to compare tests with PB and FB, and given the performance issues there is no way to run the test with fastest. So, here are the results for AdminUI:

| Name | Time | Link |
| ------------- |:-------------:| -----:|
| Chromium (without fastest) | 26m | https://travis-ci.com/ezsystems/ezplatform-page-builder/jobs/209895283 |
| Selenium (with fastest) 1 | 27m | https://travis-ci.org/ezsystems/ezplatform/jobs/547952346 |
| Selenium (with fastest) 2 | 32m | https://travis-ci.org/ezsystems/ezplatform/jobs/548461606 |
| Selenium (with fastest) 3 | 29m | https://travis-ci.org/ezsystems/ezplatform/jobs/547488160 |

Given the speed increase I think we should further investigate the issues, as there is a huge potential here 😉

IMHO the follow-ups should be about:
1) solving performance issues - this would allow us to use fastest, further speeding up the tests
2) solving the AdminUI issues - if that succeeds we might switch to Chromium on ezplatform, which would give us a time increase there
3) solving drag and drop issues - unblocking EE tests

This is not high-priority, just something we might want to do - the main goal (be able to use a driver with response header checking support) is achieved by this PR.


mnocon:
@andrerom @vidarl please have a look, as I don't want to introduce changes related to Docker without your approval 😉

mnocon:

lserwatka:
You can merge it to -ee

PR #419, opened 2019-06-25
EZP-30544: Migrate Http Cache Bundle to FOS Cache Bundle v2

ViniTou:
>Jira: https://jira.ez.no/browse/EZP-30544

## Desc 
Enable https://github.com/ezsystems/ezplatform-http-cache/pull/89.
I may have missed something from either configuration or something on sf4 side, so please check for those.
Not sure if we should also change default cache.pools to filesystem from array in ezplatform.yaml.
Also classes used in cache_pool configs (ie. cache.redis.yml) seems to not be valid.

As for merge, let's wait for above PR as version `1.0` of `ezplatform-http-cache` is non existing right now.


kmadejski:
> Not sure if we should also change default cache.pools to filesystem from array in ezplatform.yaml.

`cache.pool` has nothing to do with HTTP Cache.

> Also classes used in cache_pool configs (ie. cache.redis.yml) seems to not be valid.

What is wrong with them?

ViniTou:
well, they target  `Symfony\Component\Cache\Adapter\TagAware\FilesystemTagAwareAdapter` / `Symfony\Component\Cache\Adapter\TagAware\RedisTagAwareAdapter` and it seems they are now here `\Symfony\Component\Cache\Adapter\FilesystemTagAwareAdapter` and here `\Symfony\Component\Cache\Adapter\RedisTagAwareAdapter`

kmadejski:
Oh, right... These namespaces were correct when we used to use [ezsystems/symfony-tools](https://github.com/ezsystems/symfony-tools) which is no longer necessary on Symfony 4. Is it correct @andrerom?

andrerom:
So, this is off topic, but:
- NS was adjusted when PR was eventually accepted on Symfony 4.3
- 3.0 here needs to be updated for that, and once that is done you can revert change by @webhdx temporary adjusted default cache pool to the use non tag optimized one (cache.app afaik).
- ezsystems/symfony-tools should be updated for the new NS, with the old one as BC via class alias or deprecated class that extend the new one. In order to move the feature from "incubation" to "symfony 4.x backport".

PR #420, opened 2019-06-25
Change app.php > index.php and web > public in doc

glye:
https://github.com/ezsystems/ezplatform/commit/7750b01119c984972b2dd8a86bfedf20da86600b was not complete in doc, this completes the move from app.php to index.php, and from web/ to public/

For master only.

### DIFF
--- a/doc/apache2/vhost.template
+++ b/doc/apache2/vhost.template
@@ -19,7 +19,7 @@
     # Enabled for Dev environment
     #LogLevel debug
 
-    # "public" folder is what we expose to the world, all rewrite rules further down is relative to it.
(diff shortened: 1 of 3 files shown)


doc/apache2/vhost.template
```diff
@@ -19,7 +19,7 @@
     # Enabled for Dev environment
     #LogLevel debug
 
-    # "web" folder is what we expose to the world, all rewrite rules further down is relative to it.
+    # "public" folder is what we expose to the world, all rewrite rules further down is relative to it.
```
DominikaK:
```suggestion
    # "public" folder is what we expose to the world, all rewrite rules further down are relative to it.
```

PR #426, opened 2019-07-01
Enabled ezplatform-graphql

bdunogier:
Not much to add :-)


mnocon:
I'm getting following error in `prod` when running `composer run post-install-cmd`:
```
ocramius/package-versions: ...done generating version class
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!
!!   // Clearing the cache for the prod environment with debug
!!   // false
!!
!!
!!  In FileLoader.php line 166:
!!
!!    Bundle "OverblogGraphiQLBundle" does not exist or it is not enabled. Maybe
!!    you forgot to add it in the registerBundles() method of your App\Kernel.php
!!     file? in @OverblogGraphiQLBundle/Resources/config/routing.xml (which is be
!!    ing imported from "/home/vagrant/ee_demo/config/routes/ezplatform_graphql.y
!!    aml"). Make sure the "OverblogGraphiQLBundle" bundle is correctly registere
!!    d and loaded in the application kernel class. If the bundle is registered,
!!    make sure the bundle path "@OverblogGraphiQLBundle/Resources/config/routing
!!    .xml" is not empty.
!!
!!
!!  In Kernel.php line 231:
!!
!!    Bundle "OverblogGraphiQLBundle" does not exist or it is not enabled. Maybe
!!    you forgot to add it in the registerBundles() method of your App\Kernel.php
!!     file?
!!
!!
!!  cache:clear [--no-warmup] [--no-optional-warmers] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--siteaccess [SITEACCESS]] [--] <command>
!!
!!
Script @auto-scripts was called via post-install-cmd
```

bdunogier:
Thank you @mnocon. I've made some changes, it should be okay now.

PR #429, opened 2019-07-04
EZP-30723:Mailing does not work on Platform.sh

vidarl:
We don't set swiftmailer config correct on Platform.sh

**IMPORTANT**
When merging this to 2.4 and later, the following change is needed:
```patch
-if ($container->getParameter('mailer_host') === '127.0.0.1') {
+if ($container->resolveEnvPlaceholders($container->getParameter('mailer_host'), true) === '127.0.0.1') {

```


lserwatka:
You can merge it up.

PR #433, opened 2019-07-09
[Platform.sh] Varnish Comperssion

SerheyDolgushev:
Enable Gzip for a specified content type in Varnish

### DIFF
--- a/.platform/varnish.vcl
+++ b/.platform/varnish.vcl
@@ -130,6 +130,20 @@ sub vcl_backend_response {
 
     // Make Varnish keep all objects for up to 1 hour beyond their TTL, see vcl_hit for Request logic on this
     set beresp.grace = 1h;
+
+    // Compressing the content
+    if (beresp.http.Content-Type ~ "application/javascript"
+        || beresp.http.Content-Type ~ "application/json"
+        || beresp.http.Content-Type ~ "application/vnd.ms-fontobject"
+        || beresp.http.Content-Type ~ "application/x-font-ttf"
+        || beresp.http.Content-Type ~ "image/svg+xml"
+        || beresp.http.Content-Type ~ "text/css"
+        || beresp.http.Content-Type ~ "text/html"
+        || beresp.http.Content-Type ~ "text/plain"
+        || beresp.http.Content-Type ~ "text/xml"
+    ) {
+        set beresp.do_gzip = true;
+    }
 }
 
 // Handle purge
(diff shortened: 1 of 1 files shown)


.platform/varnish.vcl
```diff
@@ -130,6 +130,20 @@ sub vcl_backend_response {
...
+    // Compressing the content
+    if (beresp.http.Content-Type ~ "application/javascript"
+        || beresp.http.Content-Type ~ "application/json"
+        || beresp.http.Content-Type ~ "application/vnd.ms-fontobject"
+        || beresp.http.Content-Type ~ "application/x-font-ttf"
+        || beresp.http.Content-Type ~ "image/svg+xml"
+        || beresp.http.Content-Type ~ "text/css"
+        || beresp.http.Content-Type ~ "text/html"
+        || beresp.http.Content-Type ~ "text/plain"
+        || beresp.http.Content-Type ~ "text/xml"
```
adamwojs: Can we add "application/vnd.ez.api.*" as well?

.platform/varnish.vcl
```diff
@@ -130,6 +130,20 @@ sub vcl_backend_response {
...
     set beresp.grace = 1h;
+
+    // Compressing the content
+    if (beresp.http.Content-Type ~ "application/javascript"
+        || beresp.http.Content-Type ~ "application/json"
+        || beresp.http.Content-Type ~ "application/vnd.ms-fontobject"
+        || beresp.http.Content-Type ~ "application/x-font-ttf"
+        || beresp.http.Content-Type ~ "image/svg+xml"
+        || beresp.http.Content-Type ~ "text/css"
+        || beresp.http.Content-Type ~ "text/html"
```
andrerom: We’ll need to skip any of these for results containing secrets _(CSRF tokens, ...)_ to avoid being vulnerable to [BREACH](https://en.wikipedia.org/wiki/BREACH) attack.

In regards to CSRF, this goes for any kind of form, and the results can't be cached as it's per user. So in the context of Varnish AFAIK we should be able to just add a add a check here to make sure we don't gzip anything which is not cached. If this is wrong, then there is a bug in eZ Platform itself that needs to be fixed.

Open question is if there are other cases, or other secrets we send over the wire. /ping @glye

glye: None that I can think of, but important issue, good catch.

.platform/varnish.vcl
SerheyDolgushev: @adamwojs done

.platform/varnish.vcl
SerheyDolgushev: @and compersion was disabled for all HTML/XML responses.

andrerom:
👍  We should bring this change to httpcache package as well so VCL is in sync btw.

PR #437, opened 2019-07-23
Use relative symlinks for assets:install command

emodric:


andrerom:
Or do we want to re-add the config from earlier branches?
https://github.com/ezsystems/ezplatform/blob/2.5/composer.json#L125

emodric:
Does that still work?

If it does, it would be better, to prevent Symfony recipes overwriting the command in `auto-scripts`.

emodric:
Tried it, it doesn't work.

andrerom:
What about us asking them to re-add support for that?

But that said, I guess it won't help once we move to full symfony flex over meta repo.

emodric:
They actually removed it before due to it not working on some setups: https://github.com/symfony/recipes/pull/406

> But that said, I guess it won't help once we move to full symfony flex over meta repo.

Not sure what you mean?

emodric:
I'm not sure what would be the best course of action here. This patch has an effect of default version of `assets:install` being readded to `auto-scripts` as described in https://github.com/symfony/recipes/pull/406. It would require `symfony.lock` to be present to prevent it from happening (hence my other PR)

PR #438, opened 2019-07-30
Add recipes created by initial composer install

emodric:
Symfony recomends storing the `symfony.lock` file in the repo for your projects.

To reduce friction for new installs, this adds the recipes installed by the initial `composer install` as well as adding `symfony.lock` to the repo so users don't have to touch `.gitignore` when they need to store the lock file.


andrerom:
hmm, but do we want it in _our_ repo?

Atm we don't keep locks in product repo, but when making tags we swap this so projects do keep them: https://github.com/ezsystems/ezplatform/blob/master/bin/.ci/make_tag.sh#L86-L87

andrerom:
Reminds me that we should consider adding .gitattribute for certain things also: https://medium.com/@pablorsk/be-a-git-ninja-the-gitattributes-file-e58c07c9e915

emodric:
Hm, that does solve the `.gitignore` issue, but it still would cause readding the default `assets:install` command from the recipe if we would merge https://github.com/ezsystems/ezplatform/pull/437.

Related issue on `symfony/recipes`: https://github.com/symfony/recipes/issues/631

Also, Sylius does keep `symfony.lock` in their repo: https://github.com/sylius/sylius

PR #439, opened 2019-07-31
EZP-30717: Added default pagination limit for admin_group

pawbuj:
This is one of 2 PR's that moves the default value of `pagination.user_setting_limit` for `admin_group` from bundle defaults to meta repos. This is done to improve developers experience and avoid misunderstandings described here :  https://jira.ez.no/browse/EZP-30717 .

orginal ticket : https://jira.ez.no/browse/EZP-30553

### DIFF
--- a/config/packages/ezplatform.yaml
+++ b/config/packages/ezplatform.yaml
@@ -74,6 +74,9 @@ ezpublish:
 
         site:
             languages: [eng-GB]
+        admin_group:
+            pagination:
+                user_settings_limit: 10
 
     url_alias:
         slug_converter:


emodric:
For what it's worth, I don't think it's a good idea to chunk every piece of config in global `ezplatform.yml` so users can blindly change it without understanding config resolver scopes.

Issue described in https://jira.ez.no/browse/EZP-30717 is weird as it is clearly "working as intended". You can just as well open one issue for every piece of configuration available in admin UI or other scope specific configs.

pawbuj:
> For what it's worth, I don't think it's a good idea to chunk every piece of config in global `ezplatform.yml` so users can blindly change it without understanding config resolver scopes.
> 
> Issue described in https://jira.ez.no/browse/EZP-30717 is weird as it is clearly "working as intended". You can just as well open one issue for every piece of configuration available in admin UI or other scope specific configs.

The config of ez platform has cascade nature, when someone adds siteaccess-specific configuration to the bundle, the cascade nature of config is broken. It is very confusing for new ez developers (as me). It also against KISS.

emodric:
> The config of ez platform has cascade nature, when someone adds siteaccess-specific configuration to the bundle, the cascade nature of config is broken.

So don't add siteaccess specific config to default admin UI config :)

> It is very confusing for new ez developers (as me). It also against KISS.

I'm not saying it shouldn't be simple. I'm saying that it's not a good idea to chunk every config to a global `ezplatform.yml`. Sooner or later you will have a mess in the file.

pawbuj:
After internal discussion, we decided to do it a different way

PR #442, opened 2019-08-05
EZP-30816: [3.0] Support for Platform.sh 

damianz5:
https://jira.ez.no/browse/EZP-30816

note: composer.lock is required for platform.sh on build phase to avoid out of memory/resources and will be removed before merge

requires: https://github.com/ezsystems/ezplatform/pull/450


config/packages/cache_pool/cache.memcached.yaml
```diff
@@ -2,6 +2,9 @@
 #
 # For further reading on setup with eZ Platform and Memcached:
 # https://doc.ezplatform.com/en/latest/guide/persistence_cache/#memcached
+parameters:
+    cache_namespace: ezp-memcached
```
kmadejski: Maybe simply `ezp` would be enough as it is for Redis? The same question applies to: https://github.com/ezsystems/ezplatform/pull/442/files#diff-87571c11819ea311a4eadf3a9b92ea33R5

.platform.app.yaml
```diff
@@ -107,13 +109,13 @@ hooks:
         # Mainly relevant for eZ Platform demo usage, for own setup adapt this or remove and rely on migrations.
         if [ ! -f public/var/.platform.installed ]; then
             # To workaround issues with p.sh Varnish we clear container cache & temporary set Symfony Proxy
-            rm -Rf var/cache/$SYMFONY_ENV/*.*
-            HTTPCACHE_PURGE_TYPE="local" php -d memory_limit=-1 `which composer` ezplatform-install
```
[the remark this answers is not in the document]
damianz5: no, ezplatform-install is running installer and @auto-scripts (on 2.5 it was only the installer), but for deploy phase it is to late for dumping the assets or generate other files, as the filesystem is read only.

currently the `composer install` command is executed at build stage (which will also execute auto-scripts) and the ezp installer is executed at deploy (as the mysql is present during deploy)

I will try to find a better solution for this - to keep the same file for all repos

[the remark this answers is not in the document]
kmadejski: Ok, understand, thanks for the explanation. I think it can stay as you proposed, not a big deal IMO.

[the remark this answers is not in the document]
damianz5: @webhdx can we remove this line? https://github.com/ezsystems/ezplatform/pull/442/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L121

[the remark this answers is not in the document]
damianz5: Changed in composer.json https://github.com/ezsystems/ezplatform/pull/442/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L121

config/packages/000-platformsh.php
```diff
@@ -0,0 +1,142 @@
...
+use Symfony\Component\DependencyInjection\Loader;
+
+// Run for all hooks, incl build step
+if (getenv('PLATFORM_PROJECT_ENTROPY')) {
+    // Disable PHPStormPass as we don't have write access & it's not localhost
+    $container->setParameter('ezdesign.phpstorm.enabled', false);
+}
+
+// Will not be executed on build step
+$relationships = getenv('PLATFORM_RELATIONSHIPS');
```
andrerom: Some of the things here can probably be omitted in favour of https://github.com/platformsh/symfonyflex-bridge

Some of the remaining things here will need to be:
1. Symfony things => contributed to the bridge 
2. eZ things, we either need extension point in the bridge, or an eZ bridge extending the symfony flex bridge

andrerom: Benefit: We move our config to only use ENV variables, no funky php logic. And mapping between platform.sh config and EVN is done by bridge, before Symfony starts up _(it hooks into autoload logic to do it)_

damianz5: the bridge is already used is this PR for db access, but it's not supporting redis, memcached, redissession, solr or dfs (https://github.com/platformsh/symfonyflex-bridge/blob/master/platformsh-flex-env.php) and it is covered by this file.

andrerom: Sorry, missed this. Would have been easier if it was possible to see diff from 2.5 here.
Would it be possible to first do commit to revert the removal, then commit to adjust it for master?

Will make it easier to see, and might help on merges on updates in the future.

damianz5: updated

PR #444, opened 2019-08-08
[Platform.sh] Revise bundled config & inline doc

andrerom:
Goals, most trying to improve DX:
- Define clear hints on what should be done on Platform.sh Enterprise Dedicated Cluster, for partner and for Platform.sh staff as well
- Update config to cover more common needs inline
    - _e.g. redis cache clear (NB: varnish/fastly to be covered later as followup)_
- Simplify:  use COMPOSER_AUTH  and ~change to composer flavour~ _(Skipped for now as we need to install another version of node.js in build step)_
- Simplify: Align Enterprise and Open Source doc to cover all in Open source doc to avoid EE doc getting out of sync, possible now thanks to change to suggest generic COMPOSER_AUTH use for any auth needs

Target 2.5LTS and up for simplicity, we could consider bringing this to 1.13LTS as well if we anticipate further changes there (and hence conflicts due to these changes).

*Review guidelines*:
- Please use Github "*Suggest" feature if think something should be different 
- Beware the doc/platformsh/* was written long ago by someone else, and has mostly not been maintained, so I'm trying to brush it up and simplify it a bit. Ideally it should be greatly simplified, and better synced with https://docs.platform.sh/frameworks/ez.html#ez-platform and https://doc.ez.no. _But that is more for someone in eZ + platform.sh product team to deal with and not within the "improvement" scope I'm aiming for here (IMO)_

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -47,7 +47,7 @@ variables:
         # If you change to "dev" remove "--no-dev" from `composer install` command.
         SYMFONY_ENV: prod
         # Uncomment if you want to use DFS clustering:
-        # NOTE: Recommended on PE Dedicated cluster setup. Required if using legacy-bridge.
+        # NOTE: Recommended on PE Dedicated cluster setup. Required if using legacy-bridge on PE dedicated cluster setup.
         #PLATFORMSH_DFS_NFS_PATH: 'dfsdata'
 
 # The configuration of app when it is exposed to the web.
@@ -92,7 +92,7 @@ mounts:
 #    'src/AppBundle/MigrationVersions/References':
 #        source: local
 #        source_path: MigrationVersionsReferences
-# Uncomment if you want to use DFS clustering, required if using legacy-bridge.
+# Uncomment if you want to use DFS clustering, required if using legacy-bridge on PE dedicated cluster setup.
 #    'dfsdata':
 #        # Platform.sh Staff: This MUST be shared on cluster, all others SHOULD be local for performance reasons
 #        source: local
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -94,7 +94,8 @@ mounts:
 #        source_path: MigrationVersionsReferences
 # Uncomment if you want to use DFS clustering, required if using legacy-bridge.
 #    'dfsdata':
-#        source: shared
+#        # Platform.sh Staff: This MUST be shared on cluster, all others SHOULD be local for performance reasons
+#        source: local
 #        source_path: dfsdata
 
 # The hooks that will be performed when the package is deployed.
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -1,8 +1,9 @@
 # This file describes an application. You can have multiple applications
 # in the same project.
 
-# Please see doc/platformsh/README.md and doc/platformsh/INSTALL.md
-# NB: Clustered eZ Platform setups are not tested and are likely to have issues.
+# Please see doc/platformsh/README.md and doc/platformsh/INSTALL.md for eZ specific getting started instructions.
+# Full documentation: https://docs.platform.sh
+# Requirements: https://doc.ezplatform.com/en/2.5/getting_started/requirements/#ez-platform-cloud-requirements-and-setup
 
 # The name of this app. Must be unique within a project.
 name: app
@@ -13,8 +14,10 @@ dependencies:
 
 # The type of the application to build.
 type: php:7.3
+
 build:
     # "none" means we're running composer manually, see build hook
+    # We currently need to do this because of the need to install newer version of Node.js
     flavor: "none"
 
 # The relationships of the application with services or other applications.
@@ -26,7 +29,7 @@ relationships:
     # Uncomment if you want to store dfs tables in a separate database:
     #dfs_database: 'mysqldb:dfs'
     rediscache: 'rediscache:redis'
-    # If you wish to have a separate Redis instance for sessions, uncomment
+    # [Recommended] To have an isolated & persisted Redis instance for sessions, uncomment
     # this relationship and the corresponding service in .platform/services.yaml
     #redissession: 'redissession:redis'
     # If you wish to use solr, uncomment this relationship and the corresponding service in .platform/services.yaml
@@ -40,9 +43,11 @@ variables:
         # We disable Symfony Proxy, as we rather use Varnish
         SYMFONY_HTTP_CACHE: 0
         SYMFONY_TRUSTED_PROXIES: "TRUST_REMOTE"
-        # Change this if you use a different env then "prod", if you change this to "dev" remove "--no-dev" in build step
+        # Change this if you use a different env then "prod"
+        # If you change to "dev" remove "--no-dev" from `composer install` command.
         SYMFONY_ENV: prod
         # Uncomment if you want to use DFS clustering:
+        # NOTE: Recommended on PE Dedicated cluster setup. Required if using legacy-bridge.
         #PLATFORMSH_DFS_NFS_PATH: 'dfsdata'
 
 # The configuration of app when it is exposed to the web.
@@ -65,6 +70,8 @@ disk: 3072
 
 # The mounts that will be performed when the package is deployed.
 mounts:
+# PE Cluster Note: By default will set all to shared, so if moving to PE dedicated cluster you should ask platform.sh
+# Support to make sure at least cache + logs is local, while you can let web/var be shared if you prefer that over DFS.
     'var/cache':
         source: local
         source_path: cache
@@ -85,7 +92,7 @@ mounts:
 #    'src/AppBundle/MigrationVersions/References':
 #        source: local
 #        source_path: MigrationVersionsReferences
-# Uncomment if you want to use DFS clustering, WARNING: Needs to be shared, contact Platform.sh support to set this up
+# Uncomment if you want to use DFS clustering, required if using legacy-bridge.
 #    'dfsdata':
 #        source: shared
 #        source_path: dfsdata
@@ -96,6 +103,7 @@ hooks:
     build: |
         set -e
 
+        # Install newer version of Node.js, as current default (v6LTS) does not play well with WebpackEncore
         unset NPM_CONFIG_PREFIX
         curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | dash
         export NVM_DIR="$HOME/.nvm"
@@ -103,13 +111,15 @@ hooks:
         nvm current
         nvm install 10.15.3
 
-        # Configure GITHUB token if set. For use if you have own private repos, or to avoid API rate limits.
-        if [ -n "$GITHUB_TOKEN" ]; then
-            composer config github-oauth.github.com $GITHUB_TOKEN
+        if ["$COMPOSER_AUTH" == ""]; then
+            echo "TIP: If you need to authenticate against Github/Gitlab/updates.ez.no, use COMPOSER_AUTH env variable"
+            echo "See: https://docs.platform.sh/tutorials/composer-auth.html#set-the-envcomposerauth-project-variable"
         fi
 
-        rm web/app_dev.php
         composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader
+
+        # Generic cleanup to remove files we don't want to be deployed to prod (you don't need this for "dev" ENV)
+        rm web/app_dev.php
 
     # Deploy hook, access to services & done once (per cluster, not per node), only mounts are writable at this point
     # Note: Http traffic is paused while this is running, so for prod code this should finish as fast as possible, < 30s
@@ -127,15 +137,23 @@ hooks:
             touch web/var/.platform.installed
         fi
 
-        # Now that mounts are available, clear class/container cache. E.g. in case of interface changes on lazy proxies
-        rm -Rf var/cache/$SYMFONY_ENV/*.*
+        # Now that mounts are available, clear cache on mount.
+        # NB! Skip on PE Cluster setup using e.g. "if [$PLATFORM_BRANCH" != 'production']; then" & get p.sh to enable this on internal per node "pre_start" hook
+        ./bin/platformsh/pre_start-per_node-clear_cache.sh
 
-        # With that done, make sure to clear + warmup all remaining caches to take proper env variables into account
-        bin/console cache:clear
+        # If you also need to clear redis cache on every deploy, you can either use this command or redis-cli
+        # Normally this should only be needed if cached data structures changes (upgrades), or you change data via sql (e.g. restore backup)
+        ##php bin/console cache:pool:clear cache.redis
 
         # Example of additional deploy hooks if you use doctrine and/or kaliop migration bundle
-        ##bin/console doctrine:migrations:migrate --no-interaction --allow-no-migration
-        ##bin/console kaliop:migration:migrate --no-interaction --no-debug
+        ##php bin/console doctrine:migrations:migrate --no-interaction --allow-no-migration
+        ##php bin/console kaliop:migration:migrate --no-interaction --no-debug
+
+        # When using Solr, there are two cases where you'll need to rebuild indexes:
+        # - When Solr / search configuration change
+        # - On database import/restore
+        # So in development it might be convenient to rebuild indexes, slowing down deploy time
+        ##php bin/console ezplatform:reindex --processes=auto
 
     # Post deploy hook, like deploy but after being deployed and live, for deploy tasks we can do asynchronously
     #post_deploy: |
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -17,7 +17,7 @@ type: php:7.3
 
 build:
     # "none" means we're running composer manually, see build hook
-    # We currently need to do this because of the need to install newer version of Node.js
+    # We currently need to do this to install newer version of Node.js
     flavor: "none"
 
 # The relationships of the application with services or other applications.
@@ -29,7 +29,7 @@ relationships:
     # Uncomment if you want to store dfs tables in a separate database:
     #dfs_database: 'mysqldb:dfs'
     rediscache: 'rediscache:redis'
-    # [Recommended] To have an isolated & persisted Redis instance for sessions, uncomment
+    # [Recommended] To have an isolated and persisted Redis instance for sessions, uncomment
     # this relationship and the corresponding service in .platform/services.yaml
     #redissession: 'redissession:redis'
     # If you wish to use solr, uncomment this relationship and the corresponding service in .platform/services.yaml
@@ -43,11 +43,11 @@ variables:
         # We disable Symfony Proxy, as we rather use Varnish
         SYMFONY_HTTP_CACHE: 0
         SYMFONY_TRUSTED_PROXIES: "TRUST_REMOTE"
-        # Change this if you use a different env then "prod"
-        # If you change to "dev" remove "--no-dev" from `composer install` command.
+        # Change this if you use a different env than "prod"
+        # If you change to "dev" remove "--no-dev" from the `composer install` command.
         SYMFONY_ENV: prod
         # Uncomment if you want to use DFS clustering:
-        # NOTE: Recommended on PE Dedicated cluster setup. Required if using legacy-bridge on PE dedicated cluster setup.
+        # NOTE: Recommended on PE Dedicated cluster setup. Required if using Legacy Bridge on PE dedicated cluster setup.
         #PLATFORMSH_DFS_NFS_PATH: 'dfsdata'
 
 # The configuration of app when it is exposed to the web.
@@ -71,7 +71,7 @@ disk: 3072
 # The mounts that will be performed when the package is deployed.
 mounts:
 # PE Cluster Note: By default will set all to shared, so if moving to PE dedicated cluster you should ask platform.sh
-# Support to make sure at least cache + logs is local, while you can let web/var be shared if you prefer that over DFS.
+# Support to make sure at least cache + logs are local, while you can let web/var be shared if you prefer that over DFS.
     'var/cache':
         source: local
         source_path: cache
@@ -92,7 +92,7 @@ mounts:
 #    'src/AppBundle/MigrationVersions/References':
 #        source: local
 #        source_path: MigrationVersionsReferences
-# Uncomment if you want to use DFS clustering, required if using legacy-bridge on PE dedicated cluster setup.
+# Uncomment if you want to use DFS clustering, required if using Legacy Bridge on PE dedicated cluster setup.
 #    'dfsdata':
 #        # Platform.sh Staff: This MUST be shared on cluster, all others SHOULD be local for performance reasons
 #        source: local
@@ -139,10 +139,10 @@ hooks:
         fi
 
         # Now that mounts are available, clear cache on mount.
-        # NB! Skip on PE Cluster setup using e.g. "if [$PLATFORM_BRANCH" != 'production']; then" & get p.sh to enable this on internal per node "pre_start" hook
+        # Note: Skip on PE Cluster setup using e.g. "if [$PLATFORM_BRANCH" != 'production']; then" & get p.sh to enable this on internal per node "pre_start" hook
         ./bin/platformsh/pre_start-per_node-clear_cache.sh
 
-        # If you also need to clear redis cache on every deploy, you can either use this command or redis-cli
+        # If you also need to clear Redis cache on every deploy, you can either use this command or redis-cli
         # Normally this should only be needed if cached data structures changes (upgrades), or you change data via sql (e.g. restore backup)
         ##php bin/console cache:pool:clear cache.redis
 
@@ -151,7 +151,7 @@ hooks:
         ##php bin/console kaliop:migration:migrate --no-interaction --no-debug
 
         # When using Solr, there are two cases where you'll need to rebuild indexes:
-        # - When Solr / search configuration change
+        # - When Solr / search configuration changes
         # - On database import/restore
         # So in development it might be convenient to rebuild indexes, slowing down deploy time
         ##php bin/console ezplatform:reindex --processes=auto
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -1,10 +1,10 @@
-# Default settings in order to setup eZ Platform demo install on Platform.sh dev instances
+# Default settings in order to set up eZ Platform demo install on Platform.sh dev instances
 #
-# NB: Like on own servers, make sure to tune Redis/Solr/Varnish/Mysql memory/disk size for your installation to avoid issues.
+# Note: Like on own servers, make sure to tune Redis/Solr/Varnish/MySQL memory/disk size for your installation to avoid issues.
 #     Reach out to platform.sh support to get help on this and insight into your disk/memory usage.
 
 mysqldb:
-    # NB! If you change this also change DATABASE_VERSION, as used by "doctrine.dbal.server_version" in config.yml.
+    # Note: If you change this also change DATABASE_VERSION, as used by "doctrine.dbal.server_version" in config.yml.
     type: mysql:10.2
     # Version 10.1 might be a better option when running Legacy Bridge
     # For more information see https://doc.ezplatform.com/en/2.1/getting_started/requirements_and_system_configuration/#supported-setups
@@ -28,7 +28,7 @@ mysqldb:
 # For use by Symfony Cache (used by eZ Platform SPI Persistence Cache)
 rediscache:
     type: 'redis:5.0'
-    # For cache you might need to increase the size of your plan if your install sizeable amount of content.
+    # For cache you might need to increase the size of your plan if your installation has a sizeable amount of content.
     # Check with platform.sh staff if in doubt on this, and if it would make sense to configure larger redis size here.
     # size: L
     configuration:
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -48,8 +48,8 @@ rediscache:
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
 #    type: redis-persistent:5.0
- # Disk size should be bigger than redis'  "maxmemory" setting due to https://redis.io/topics/persistence#log-rewriting.
- # The memory given to redis is dependend on your plan and "size: ". Adjust "disk: " accordingly
+ # Disk size should be bigger than Redis'  "maxmemory" setting due to https://redis.io/topics/persistence#log-rewriting.
+ # The memory given to Redis depends on your plan and "size: ". Adjust "disk: " accordingly.
  #    disk: 512
 #    configuration:
 #        maxmemory_policy: allkeys-lru
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -29,7 +29,7 @@ mysqldb:
 rediscache:
     type: 'redis:5.0'
     configuration:
-        # For cache eZ Platform 2.5+ RedisTagAwareAdapter requries one of the 'volatile-lru' eviction policies
+        # For cache eZ Platform 2.5+ RedisTagAwareAdapter requries one of the 'volatile-*' eviction policies
         # https://docs.platform.sh/configuration/services/redis.html#eviction-policy
         # https://doc.ezplatform.com/en/2.5/getting_started/requirements/
         maxmemory_policy: volatile-lru
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -28,6 +28,9 @@ mysqldb:
 # For use by Symfony Cache (used by eZ Platform SPI Persistence Cache)
 rediscache:
     type: 'redis:5.0'
+    # For cache you might need to increase the size of your plan if your install sizeable amount of content.
+    # Check with platform.sh staff if in doubt on this, and if it would make sense to configure larger redis size here.
+    # size: L
     configuration:
         # For cache eZ Platform 2.5+ RedisTagAwareAdapter requries one of the 'volatile-*' eviction policies
         # https://docs.platform.sh/configuration/services/redis.html#eviction-policy
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -48,7 +48,9 @@ rediscache:
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
 #    type: redis-persistent:5.0
-#    disk: 512
+ # Disk size should be bigger than redis'  "maxmemory" setting due to https://redis.io/topics/persistence#log-rewriting.
+ # The memory given to redis is dependend on your plan and "size: ". Adjust "disk: " accordingly
+ #    disk: 512
 #    configuration:
 #        maxmemory_policy: allkeys-lru
 
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -48,9 +48,9 @@ rediscache:
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
 #    type: redis-persistent:5.0
- # Disk size should be bigger than Redis'  "maxmemory" setting due to https://redis.io/topics/persistence#log-rewriting.
- # The memory given to Redis depends on your plan and "size: ". Adjust "disk: " accordingly.
- #    disk: 512
+# Disk size should be bigger than Redis'  "maxmemory" setting due to https://redis.io/topics/persistence#log-rewriting.
+# The memory given to Redis depends on your plan and "size: ". Adjust "disk: " accordingly.
+#    disk: 512
 #    configuration:
 #        maxmemory_policy: allkeys-lru
 
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -1,4 +1,10 @@
+# Default settings in order to setup eZ Platform demo install on Platform.sh dev instances
+#
+# NB: Like on own servers, make sure to tune Redis/Solr/Varnish/Mysql memory/disk size for your installation to avoid issues.
+#     Reach out to platform.sh support to get help on this and insight into your disk/memory usage.
+
 mysqldb:
+    # NB! If you change this also change DATABASE_VERSION, as used by "doctrine.dbal.server_version" in config.yml.
     type: mysql:10.2
     # Version 10.1 might be a better option when running Legacy Bridge
     # For more information see https://doc.ezplatform.com/en/2.1/getting_started/requirements_and_system_configuration/#supported-setups
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -42,7 +42,7 @@ rediscache:
 #redissession:
 #    type: 'redis:5.0'
 #    configuration:
-#        maxmemory_policy: allkeys-ttl
+#        maxmemory_policy: allkeys-lru
 #
 # Alternatively if you have a requirement that sessions are persisted across server/redis restarts,
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
@@ -50,7 +50,7 @@ rediscache:
 #    type: redis-persistent:5.0
 #    disk: 512
 #    configuration:
-#        maxmemory_policy: allkeys-ttl
+#        maxmemory_policy: allkeys-lru
 
 # If you wish to use solr, uncomment this service and the corresponding relationship in .platform.app.yaml.
 # Also, you need to generate the config using:
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -27,18 +27,27 @@ mysqldb:
 
 # For use by Symfony Cache (used by eZ Platform SPI Persistence Cache)
 rediscache:
-    type: 'redis:3.2'
+    type: 'redis:5.0'
+    configuration:
+        # For cache eZ Platform 2.5+ RedisTagAwareAdapter requries one of the 'volatile-lru' eviction policies
+        # https://docs.platform.sh/configuration/services/redis.html#eviction-policy
+        # https://doc.ezplatform.com/en/2.5/getting_started/requirements/
+        maxmemory_policy: volatile-lru
 
 # If you wish to have a separate Redis instance for sessions, uncomment
 # this service and the corresponding relationship in .platform.app.yaml.
 #redissession:
-#    type: 'redis:3.2'
+#    type: 'redis:5.0'
+#    configuration:
+#        maxmemory_policy: allkeys-ttl
 #
 # Alternatively if you have a requirement that sessions are persisted across server/redis restarts,
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
-#    type: redis-persistent:3.2
+#    type: redis-persistent:5.0
 #    disk: 512
+#    configuration:
+#        maxmemory_policy: allkeys-ttl
 
 # If you wish to use solr, uncomment this service and the corresponding relationship in .platform.app.yaml.
 # Also, you need to generate the config using:
--- a/bin/platformsh/pre_start-per_node-clear_cache.sh
+++ b/bin/platformsh/pre_start-per_node-clear_cache.sh
@@ -9,6 +9,6 @@ echo "removing var/cache/${SYMFONY_ENV}/*.* to avoid Symfony container issues on
 rm -Rf var/cache/${SYMFONY_ENV}/*.*
 #date
 echo "clearing application cache"
-bin/console cache:clear
+php bin/console cache:clear
 #date
 echo "done executing pre_start cache clear"
--- a/bin/platformsh/pre_start-per_node-clear_cache.sh
+++ b/bin/platformsh/pre_start-per_node-clear_cache.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
-# This script gets run as part of the .platform.app.yaml deploy step
-# On PE Cluster (usually just production) this should be setup by P.sh team as part of pre_start event
+# This script is run as part of the .platform.app.yaml deployment step
+# On PE Cluster (usually just production) this should be setup by platform.sh team as part of pre_start event
 
 set -e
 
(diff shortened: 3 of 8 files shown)


.platform.app.yaml
```diff
@@ -115,13 +115,22 @@ hooks:
...
 
+        # If you also need to clear redis cache on every deploy, you can either use this command or redis-cli
+        # Normally this should only be needed if cached data structures changes (upgrades), or you change data via sql (e.g. restore backup)
+        ##bin/console cache:pool:clear cache.redis
+
         # Example of additional deploy hooks if you use doctrine and/or kaliop migration bundle
         ##bin/console doctrine:migrations:migrate --no-interaction --allow-no-migration
         ##bin/console kaliop:migration:migrate --no-interaction --no-debug
 
+      # P.sh TODO: Explain Pre_Start here? (assuming it runs after deploy) And what should be in it in commented out form, probably:
```
kmadejski: This is definitely required. I had an issue regarding that on a cluster, it turned out that on `deploy` stage cache is cleared **only** for the one node. We ended up with a rebuild container on one node, but on other nodes, we were getting exceptions. 

__Note__: P.sh support created a bash snippet which takes care of clearing caches on all nodes. The script should be executed on `pre_start` hook. I think it's a good idea to add it to our repository (commented out for non-cluster setups) as well or we can even switch to using it for all (assuming that `pre_start` hook  is available for non-cluster setups as well) and remove cache clearing stuff from `deploy` hook then. 

```
# This script gets run as part of the .platform.app.yaml
# on production this would get executed just before the deploy hook.

set -e

date
echo "removing var/cache/${SYMFONY_ENV}/*.*"
rm -Rf var/cache/${SYMFONY_ENV}/*.*
date
echo "clearing application cache"
bin/console cache:clear
date
echo "done executing pre_start hook"
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+If you are on a Platform.sh Enterprise Dedicated Cluster setup, typically named PE-6 / PE-12 / (...), you'll need to do some adjustment on the platform.sh config and you'll
+have dedicated on-boarding where topics around using eZ Platform on this cluster will be gone true before you can deploy to it.
+
+If you want to read up a bit, look in the bundled platform.sh config there are a lot of config commented out by default, some of these are explicitly for Dedicated Cluster setup.
+
+As a minimum it's recommended to:
+- Enable eZ Clustered DFS setup _(A **requirement** if you use eZ Publish legacy or legacy bridge)_
+- Setup own persisted redis service for sessions
+- As always, tune memory size for Redis and Solr to make sure it can handle your traffic and data needs
+
```
kmadejski: It's rather on P.sh side but in my opinion worth to mention here as well, that `var/cache` and `var/logs` should be mounted as `local`. This is currently not doable by ourselves/config files and needs to be handled by P.sh because cluster doesn't really take into account mounts configuration from `.platform.app.yml` file apparently. Anyway, good to have a reminder here for both us and them.

.platform.app.yaml
devicezero: `pre_start` only exists on clusters.

It can replace the parts in the build hooks e.g. via
```
if test "$PLATFORM_BRANCH" != 'production'; then
    ./psh-pre_start.sh
fi
```
(means it only runs on non-production branches (production being on a cluster)

andrerom: Ok, so we should maybe place it in `.platform/pre_start.sh` for instance then, or ´bin/.platform.pre_start.sh´, and call that from deploy like you show above. WDYT @devicezero ?

devicezero: the `.platform` folder is not available on the server, so it would need to bin in `bin/` or elsewhere

andrerom: Tried to propose something for this here:
https://github.com/ezsystems/ezplatform/pull/444/commits/f04c4afaa0ff9b52fc77259c7e7598a1d65ba82b

As usual having a hard time writing generic config that will work seamlessly across PAAS/Grid and Cluster.... So details are in comment and it's "PAAS/Grid Frist"

.platform.app.yaml
```diff
@@ -26,7 +29,7 @@ relationships:
     # Uncomment if you want to store dfs tables in a separate database:
     #dfs_database: 'mysqldb:dfs'
     rediscache: 'rediscache:redis'
-    # If you wish to have a separate Redis instance for sessions, uncomment
+    # [Recommended] To have a isolated & persisted Redis instance for sessions, uncomment
```
michalmilc90:
```suggestion
    # [Recommended] To have an isolated & persisted Redis instance for sessions, uncomment
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
 # Install eZ Platform on Platform.sh
 
-> **Beta**: Instructions and Tools *(Platform.sh configuration files, scripts, ...)* described on this page are currently in Beta for community testing & contribution, and may change without notice.
-
 ## Installation using eZ Platform template
 This is the simplest approach, but may be less up to date with current development than the other, more manual approach.
 
+### Using eZ Platform Cloud
+With a eZ Platform Cloud subscription you have access to create both open source, Enterprise and in the future Commerce using templates:
```
michalmilc90:
```suggestion
With eZ Platform Cloud subscription, you are able to create an instance of both open source and Enterprise using templates. In the future it also allow you to create an instance of Commerce version.

To use eZ Platform Cloud:
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
-1. Fork [eZ Platform](https://github.com/ezsystems/ezplatform/) and clone your fork locally.
-1. Add the platform remote of your project, and push your branch. The Platform.sh setup wizard tells you the command to use. Example:  
-   `git remote add platform my_project@git.eu.platform.sh:my_project.git`  
-   1. Optional: Set the SYMFONY_ENV environment variable to 'prod' or 'dev':  
-      `export SYMFONY_ENV=dev`  
-      If you don't do this, the local build will default to 'prod'.  
-      **NB:** For this to affect remote builds, it must also be set as a Platform.sh project variable, see below.
-1. Push your branch. The Platform.sh setup wizard tells you the command to use. Example:  
+1. Login or create an account at [Platform.sh](https://platform.sh) _(or login to [cloud.ezplatform.com](https://cloud.ezplatform.com/) if you have an subscription)_
+2. Unless this has already been done for you, create a new project, using the "Import your existing code" option. Follow the setup wizard, but halt at the end, before clicking "Finish".
```
michalmilc90:
```suggestion
2. Unless this has already been done for you, create a new project by using the **Import your existing code** option. Follow the setup wizard, but halt at the end, before clicking **Finish**.
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+3. Fork [eZ Platform](https://github.com/ezsystems/ezplatform/) _(or [eZ Platform Enterprise](https://github.com/ezsystems/ezplatform-ee/)/[eZ Commerce](https://github.com/ezsystems/ezcommerce/) if you have a subscription)_ and clone your fork locally.
+4. Add the platform remote of your project, and push your branch. The Platform.sh setup wizard tells you the command to use. Example:
```
michalmilc90:
```suggestion
4. Add the platform remote of your project, and push your branch. The Platform.sh setup wizard provides the command to use. Example:
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+   `git remote add platform my_project@git.eu.platform.sh:my_project.git`
+5. [Optional] Authentication against Github/Bitbucket/Gitlab/updates.ez.no for locale development:
+   If you have private packages from own git repositories or use eZ Platform Enterprise, you can use Composer's
+   [auth.json](https://getcomposer.org/doc/articles/http-basic-authentication.md) file or [COMPOSER_AUTH](https://getcomposer.org/doc/03-cli.md#composer-auth) environment variable for this.
+   Here's an example of `COMPOSER_AUTH` usage to authenticate for eZ Enterprise packages:
+   `export COMPOSER_AUTH='{"http-basic":{"updates.ez.no":{"username":"network-id","password":"token-key"}}}'`
+6. [Optional] Set the `env:symfony_env` or `env:COMPOSER_AUTH` project variables
```
michalmilc90:
```suggestion
6. [Optional] Set the `env:symfony_env` or `env:COMPOSER_AUTH` project variables by performing the following steps:
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+   1. Install the Platform.sh CLI according to https://docs.platform.sh/gettingstarted/own-code/cli-install.html
+   2. Run `platform`
+      Run `platform get <your project id>`
+   3. Authentication against Github/Bitbucket/Gitlab/updates.ez.no
+       For example set the project variables for your eZ Network installation ID and token:
+      `platform project:variable:create env:COMPOSER_AUTH '{"http-basic":{"updates.ez.no":{"username":"network-id","password":"token-key"}}}' --no-visible-runtime --sensitive true`
+   4. If you have the need to debug things remotely, set the SYMFONY_ENV environment variable to 'dev':
```
michalmilc90:
```suggestion
   4. If you have the need to debug things remotely, set the `SYMFONY_ENV` environment variable to 'dev':
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+      `platform project:variable:set env:SYMONY_ENV dev`
```
michalmilc90:
```suggestion
      `platform project:variable:set env:SYMONY_ENV dev`.
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+7. Push your branch. The Platform.sh setup wizard tells you the command to use. Example:
```
michalmilc90:
```suggestion
7. Push your branch. The Platform.sh setup wizard provides the command to use. Example:
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+
+Yes, like all configuration (YML/VCL/..) bundled with the eZ Platform installation, all config in the "root project" is yours, and for you to customize for your needs.
+The bundled config is a recommended safe default that you can start from, just make sure to make it possible to merge in future config changes when you later need to upgrade.
+
+## Using Enterprise Dedicated Cluster
+
+If you are on a Platform.sh Enterprise Dedicated Cluster setup, typically named PE-6 / PE-12 / (...), you'll need to do some adjustment on the platform.sh config and you'll
+have dedicated on-boarding where topics around using eZ Platform on this cluster will be gone true before you can deploy to it.
+
+If you want to read up a bit, look in the bundled platform.sh config there are a lot of config commented out by default, some of these are explicitly for Dedicated Cluster setup.
```
michalmilc90:
```suggestion
If you want to read up a bit, look in the bundled platform.sh config. There are a lot of config commented out by default, some of these are explicitly for Dedicated Cluster setup.
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+
+NOTE: On hardware side we recommend at a minimum a PE-6++ ("ME-6") instance with additional memory available for Redis and Solr needs.
+
+## Downgrading services
+
+Note that if you downgrade certain services like mysql, this can cause your deploy to hang. Contact platform.sh support if this happens and they will downgrade it for you.
+
+## Re-install
+
+If you use default config, where database is installed on first deploy and thus have installed eZ Platform or the Enterprise Edition on this Platform.sh instance before, you may need to remove the web/var/.platform.installed file to ensure the installation is performed in the deploy stage.
```
michalmilc90:
```suggestion
If you use default config, where database is installed on first deploy and thus have installed eZ Platform or the Enterprise Edition on this Platform.sh instance before, you may need to remove the `web/var/.platform.installed` file to ensure the installation is performed in the deploy stage.
```

.platform.app.yaml
```diff
@@ -13,9 +14,11 @@ dependencies:
 
 # The type of the application to build.
 type: php:7.3
-build:
-    # "none" means we're running composer manually, see build hook
-    flavor: "none"
+
+# "composer" is default on PHP, it means packages will be installed before build hook is executed, alternative is "none"
```
bdunogier: `"composer" is default on PHP. Packages will be...`

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,81 @@
...
+
+The symptom for this is when, in the backend, you go to Content -> Form Manager and get an error message, or when the "My Drafts Scheduled for Future Publication" and "All Drafts Scheduled for Future Publication" sections on "My Dashboard" will not load.
```
michalmilc90:
```suggestion
The symptom for this is when, in the Back Office, you go to **Content** -> **Form** Manager and get an error message or when the "My Drafts Scheduled for Future Publication" and "All Drafts Scheduled for Future Publication" sections on "My Dashboard" do not load.
```

.platform.app.yaml
```diff
@@ -13,9 +14,11 @@ dependencies:
...
+# NB! To authentication to Github/Gitlab/updates.ez.no use COMPOSER_AUTH as described in doc/platformsh/INSTALL.md
```
bdunogier: "To authentication to..." -> "To authenticate against" ?

.platform.app.yaml
bdunogier: "an isolated"

.platform.app.yaml
```diff
@@ -43,6 +46,7 @@ variables:
         # Change this if you use a different env then "prod", if you change this to "dev" remove "--no-dev" in build step
         SYMFONY_ENV: prod
         # Uncomment if you want to use DFS clustering:
+        # NOTE: Recommended on PE Dedicated cluster setup, especially if using legacy-bridge where it's a requirement.
```
bdunogier: "Recommended on PE Dedicated cluster setup. Required if using legacy-bridge."

.platform.app.yaml
```diff
@@ -61,6 +65,8 @@ disk: 3072
 
 # The mounts that will be performed when the package is deployed.
 mounts:
+# PE Cluster Note: If you plan to move to PE cluster you should ask p-sh support to make sure at least
```
bdunogier: The phrase feels complex. Could it be simplified / splitted ?

.platform.app.yaml
```diff
@@ -112,11 +112,13 @@ hooks:
             touch web/var/.platform.installed
         fi
 
-        # Now that mounts are available, clear class/container cache. E.g. in case of interface changes on lazy proxies
-        rm -Rf var/cache/$SYMFONY_ENV/*.*
+        # Now that mounts are available, clear cache on mount.
+        # NB! Skip on PE Cluster setup using e.g. "if [$PLATFORM_BRANCH" != 'production']; then" & get p.sh to endable this on internal per node "pre_start" hook
```
bdunogier: "endable" -> enable ?

.platform.app.yaml
andrerom: please use suggest feature @bdunogier ;)

bdunogier: Ooops, I will next time, I promise :)

.platform.app.yaml
andrerom:
```suggestion
# PE Cluster Note:When/if moving to PE dedicated cluster you should ask platfrom.sh Support to make sure at least
```

.platform.app.yaml
andrerom: Did it for you this time, but please remember next time, it's seriously a waste of both reviewers and contributors time to suggest changes without actually directly suggesting the diff

Especially now that we have the means to do so.

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,83 @@
...
+Note that if you downgrade certain services like mysql, this can cause your deploy to hang. Contact platform.sh support if this happens and they will downgrade it for you.
+
+## Re-install
+
+If you use default config, where database is installed on first deploy and thus have installed eZ Platform or the Enterprise Edition on this Platform.sh instance before, you may need to remove the `web/var/.platform.installed` file to ensure the installation is performed in the deploy stage.
+
+The symptom for this is when, in the Back Office, you go to **Content** -> **Form** Manager and get an error message or when the "My Drafts Scheduled for Future Publication" and "All Drafts Scheduled for Future Publication" sections on "My Dashboard" do not load.
+
+To do this, go to the Platform.sh web interface -> "Access site" and copy the "SSH access" command. Then, run the SSH command from a terminal, and:
+`rm web/var/.platform.installed`
```
crevillo: worth to mention that you may need to do this per environment? also probably quicker  using `platform ssh` from platform cli?

andrerom: How would you suggest it? 😏

doc/platformsh/INSTALL.md
andrerom: Tried to fix this in 0fd5d00

andrerom:
⚠️ NB! Change to use composer build flavor is incompatible with the change in #450* ⚠️ 


_\* As composer flavour will run composer install before build runs, meaning the node.js version fix won't work anymore._

doc/platformsh/INSTALL.md
crevillo: how about... `To do this, execute platform ssh -e [env] in a terminal, being env the one where you want to perform the reinstall. Then, run rm web/var/.platform.installed`

.platform.app.yaml
```diff
@@ -112,11 +113,13 @@ hooks:
...
-        rm -Rf var/cache/$SYMFONY_ENV/*.*
+        # Now that mounts are available, clear cache on mount.
+        # NB! Skip on PE Cluster setup using e.g. "if [$PLATFORM_BRANCH" != 'production']; then" & get p.sh to enable this on internal per node "pre_start" hook
+        ./bin/platformsh/pre_start-per_node-clear_cache.sh
 
-        # With that done, make sure to clear + warmup all remaining caches to take proper env variables into account
-        bin/console cache:clear
+        # If you also need to clear redis cache on every deploy, you can either use this command or redis-cli
+        # Normally this should only be needed if cached data structures changes (upgrades), or you change data via sql (e.g. restore backup)
+        ##bin/console cache:pool:clear cache.redis
```
reithor: Note regarding solr indexes:
```
# When using solr, in most cases rebuilding indexes on each deploy is not neccessary.
# Especially during early stages of a project -  when configuration is often changed
# - rebuilding indexes automatically is an option to keep search results accurate
## ./bin/console ezplatform:reindex --processes=auto

```

.platform.app.yaml
```diff
@@ -43,6 +47,7 @@ variables:
         # Change this if you use a different env then "prod", if you change this to "dev" remove "--no-dev" in build step
```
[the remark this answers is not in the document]
andrerom: @reithor Would you be able to use the suggest feature here on the comments?

reithor:
```suggestion
        # Change this if you use a different env then "prod", if you change this to "dev" remove "--no-dev"
        # and comment out 'rm web/app_dev.php' in build step
```

.platform.app.yaml
reithor:
```suggestion
        # When using solr, in most cases rebuilding indexes on each deploy is not neccessary.
        # Especially during early stages of a project -  when configuration is often changed
        # - rebuilding indexes automatically is an option to keep search results accurate
        ##bin/console ezplatform:reindex --processes=auto
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,89 @@
...
```
reithor:
```suggestion
`rm web/var/.platform.installed`

Having platform.sh CLI installed, you could alternatively execute platform ssh -e [env] 'rm web/var/.platform.installed', being env the environment where you want to perform the reinstall.
```

.platform.app.yaml
andrerom: @reithor Actually you don't need to comment out `rm web/app_dev.php`, we handle all dev stuff in `app.php`, while `app_dev.php` is just convince to now have to set ENV, which you anyway will have to do here.

```suggestion
       # Change this if you use a different env then "prod", if you change this to "dev" remove "--no-dev" from `composer install` command.
```

ok? _(assumes we go back to build flavour:none and reinstate own call to composer install, otherwise this section should be removed actually)_

.platform.app.yaml
andrerom: What about:

```suggestion

        # When using solr, there are two cases where you'll need to rebuild indexes:
        # - When configuration is often changed
        # - On database import/restore
        # So in development it might be convenient to rebuild indexes, slowing down deploy time
        ##bin/console ezplatform:reindex --processes=auto

```

doc/platformsh/INSTALL.md
andrerom: Sounds ok to me, @crevillo wdyt?

doc/platformsh/INSTALL.md
andrerom: I'll do this one locally, as it affects line above too.

doc/platformsh/INSTALL.md
crevillo: yeah, ok for me too :).

.platform.app.yaml
crevillo: hi @reithor , that's weird. i've just tried in our p.sh instance and no need to add php there
```
web@mzz63aip7jyjzyt6hihw53vcle:~/ezplatform$ bin/console ezplatform:reindex --processes=auto
Re-indexing started for search engine: eZ Platform Legacy (SQL) Search Engine

Purging index...
Re-Creating index for 567 items across 12 iteration(s), using 7 parallel child processes:
 12/12 [============================] 100%
```

(note: if you see ezplatform in our prompt is because we're using launchpad here, but our .platform.app.yml is almost the same as the one provide with eZ 2.5)

andrerom: @reithor They probably have installed from one of our downloads with all in one, we've had issues with the zip / tar archives not retaining rights on archives a few times.

One solution to that is to add to permissions steps in doc also add command for making bin/* executable as part of the instructions, and in that case the change (if you get one) should be checked into git.

reithor: > hi @reithor , that's weird. i've just tried in our p.sh instance and no need to add php there
> 

By default (when using git clone/composer) `console` should be executable - so there  is no need to add php.
It is more a fail-safe for the case when `console` is not executable- for whatever reason

But I now feel, that checking pemissions in bin/ - and when necessary fixing missing executable option - is the better option.

andrerom:
Fixed, rebased and ready for final review @devicezero, @Crell, @kmadejski, @vidarl, @reithor, @michalmilc90 ;)

.platform.app.yaml
```diff
@@ -85,7 +92,7 @@ mounts:
 #    'src/AppBundle/MigrationVersions/References':
 #        source: local
 #        source_path: MigrationVersionsReferences
-# Uncomment if you want to use DFS clustering, WARNING: Needs to be shared, contact Platform.sh support to set this up
+# Uncomment if you want to use DFS clustering, required if using legacy-bridge.
 #    'dfsdata':
 #        source: shared
```
alustenberg:
```suggestion
#        source: local
```

`source: shared` will be rejected as invalid configuration, should be `local` for integration environments.   dedicated enterprise clusters have their mounts managed elsewhere, which is where 'shared' (glusterfs) mounts come into play.

andrerom: DFS Cluster is for the most part pointless for single server, so in lack of a better way this was intended to tell p.sh support staff that this __mount__ must be shared. While cache and log __strongly should__ be local.

andrerom: @alustenberg What about:
```suggestion
#        # Platform.sh Staff: This MUST be shared on cluster, all others SHOULD be local for performance reasons
#        source: local
```

.platform/services.yaml
```diff
@@ -21,18 +27,27 @@ mysqldb:
 
 # For use by Symfony Cache (used by eZ Platform SPI Persistence Cache)
 rediscache:
-    type: 'redis:3.2'
+    type: 'redis:5.0'
+    configuration:
+        # For cache eZ Platform 2.5+ RedisTagAwareAdapter requries one of the 'volatile-lru' eviction policies
+        # https://docs.platform.sh/configuration/services/redis.html#eviction-policy
+        # https://doc.ezplatform.com/en/2.5/getting_started/requirements/
+        maxmemory_policy: volatile-lru
```
andrerom: Note: This can be changed to `volatile-lfu` which is better match for us, once Platform.sh config supports it _(ETA start rollout mid week 39 or 40)_

.platform/services.yaml
```diff
@@ -21,18 +27,27 @@ mysqldb:
 
 # For use by Symfony Cache (used by eZ Platform SPI Persistence Cache)
 rediscache:
-    type: 'redis:3.2'
+    type: 'redis:5.0'
+    configuration:
+        # For cache eZ Platform 2.5+ RedisTagAwareAdapter requries one of the 'volatile-lru' eviction policies
```
andrerom:
```suggestion
        # For cache eZ Platform 2.5+ RedisTagAwareAdapter requries one of the 'volatile-*' eviction policies
```

.platform/services.yaml
```diff
@@ -21,18 +27,30 @@ mysqldb:
...
+#        maxmemory_policy: allkeys-ttl
 #
 # Alternatively if you have a requirement that sessions are persisted across server/redis restarts,
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
-#    type: redis-persistent:3.2
+#    type: redis-persistent:5.0
 #    disk: 512
+#    configuration:
+#        maxmemory_policy: allkeys-ttl
```
ralt: that's not a valid policy, is it?

bin/platformsh/pre_start-per_node-clear_cache.sh
```diff
@@ -0,0 +1,14 @@
...
+# On PE Cluster (usually just production) this should be setup by P.sh team as part of pre_start event
+
+set -e
+
+#date
+echo "removing var/cache/${SYMFONY_ENV}/*.* to avoid Symfony container issues on interface changes"
+rm -Rf var/cache/${SYMFONY_ENV}/*.*
+#date
+echo "clearing application cache"
+bin/console cache:clear
```
ralt: you use `php bin/console` in some places but not here, is it normal?

.platform/services.yaml
```diff
@@ -1,4 +1,10 @@
+# Default settings in order to setup eZ Platform demo install on Platform.sh dev instances
+#
+# NB: Like on own servers, make sure to tune Redis/Solr/Varnish/Mysql memory/disk size for your installation to avoid issues.
+#     Reach out to platform.sh support to get help on this and insight into your disk/memory usage.
+
 mysqldb:
+    # NB! If you change this also change DATABASE_VERSION, as used by "doctrine.dbal.server_version" in config.yml.
     type: mysql:10.2
```
ralt: May I recommend `mariadb:10.4`? :-)

.platform.app.yaml
```diff
@@ -40,9 +43,11 @@ variables:
         # We disable Symfony Proxy, as we rather use Varnish
         SYMFONY_HTTP_CACHE: 0
         SYMFONY_TRUSTED_PROXIES: "TRUST_REMOTE"
-        # Change this if you use a different env then "prod", if you change this to "dev" remove "--no-dev" in build step
+        # Change this if you use a different env then "prod"
+        # If you change to "dev" remove "--no-dev" from `composer install` command.
         SYMFONY_ENV: prod
         # Uncomment if you want to use DFS clustering:
+        # NOTE: Recommended on PE Dedicated cluster setup. Required if using legacy-bridge.
```
vidarl:
```suggestion
        # NOTE: Recommended on PE Dedicated cluster setup. Required if using legacy-bridge on PE dedicated cluster setup.
```

.platform.app.yaml
```diff
@@ -85,9 +92,10 @@ mounts:
 #    'src/AppBundle/MigrationVersions/References':
 #        source: local
 #        source_path: MigrationVersionsReferences
-# Uncomment if you want to use DFS clustering, WARNING: Needs to be shared, contact Platform.sh support to set this up
+# Uncomment if you want to use DFS clustering, required if using legacy-bridge.
```
vidarl:
```suggestion
# Uncomment if you want to use DFS clustering, required if using legacy-bridge on PE dedicated cluster setup.
```

.platform/services.yaml
andrerom: not yet, we haven't tested that. But we can do it in master without needing to test it for sure.

but then moving between branches is a pain as certain services needs intervention on mysql downgrades... :P Or has that been fixed now?

.platform/services.yaml
andrerom: Indeed, wishful thinking there :P

```suggestion
#        maxmemory_policy: allkeys-lru
```

.platform/services.yaml
```diff
@@ -21,18 +27,30 @@ mysqldb:
...
+        # https://doc.ezplatform.com/en/2.5/getting_started/requirements/
+        maxmemory_policy: volatile-lru
 
 # If you wish to have a separate Redis instance for sessions, uncomment
 # this service and the corresponding relationship in .platform.app.yaml.
 #redissession:
-#    type: 'redis:3.2'
+#    type: 'redis:5.0'
+#    configuration:
+#        maxmemory_policy: allkeys-ttl
```
andrerom:
```suggestion
#        maxmemory_policy: allkeys-lru
```

.platform/services.yaml
```diff
@@ -21,18 +27,30 @@ mysqldb:
...
 #
 # Alternatively if you have a requirement that sessions are persisted across server/redis restarts,
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
-#    type: redis-persistent:3.2
+#    type: redis-persistent:5.0
 #    disk: 512
```
vidarl: We need more insight on size of redis..
I believe `disk:` should actually be bigger than redis' `maxmemory` setting due to https://redis.io/topics/persistence#log-rewriting.
And the `maxmemory` of redis is dependend on plan and `size: ` AFAIK...

bin/platformsh/pre_start-per_node-clear_cache.sh
andrerom: It doesn't matter as `bin/console` has `#!/usr/bin/env php`, but let's try to be consistent nevertheless 👍 :

```suggestion
php bin/console cache:clear
```

.platform/services.yaml
andrerom: I think you are right, but I'm also a bit lost what we should suggest here.

@vidarl Can you check with @ralt for instance and suggest something here?

.platform.app.yaml
```diff
@@ -40,9 +43,11 @@ variables:
         # We disable Symfony Proxy, as we rather use Varnish
         SYMFONY_HTTP_CACHE: 0
         SYMFONY_TRUSTED_PROXIES: "TRUST_REMOTE"
-        # Change this if you use a different env then "prod", if you change this to "dev" remove "--no-dev" in build step
+        # Change this if you use a different env then "prod"
```
DominikaK:
```suggestion
        # Change this if you use a different env than "prod"
```

.platform.app.yaml
```diff
@@ -40,9 +43,11 @@ variables:
...
+        # If you change to "dev" remove "--no-dev" from `composer install` command.
```
DominikaK:
```suggestion
        # If you change to "dev" remove "--no-dev" from the `composer install` command.
```

.platform.app.yaml
```diff
@@ -40,9 +43,11 @@ variables:
...
         SYMFONY_ENV: prod
         # Uncomment if you want to use DFS clustering:
+        # NOTE: Recommended on PE Dedicated cluster setup. Required if using legacy-bridge on PE dedicated cluster setup.
```
DominikaK:
```suggestion
        # NOTE: Recommended on PE Dedicated cluster setup. Required if using Legacy Bridge on PE dedicated cluster setup.
```

.platform.app.yaml
```diff
@@ -65,6 +70,8 @@ disk: 3072
 
 # The mounts that will be performed when the package is deployed.
 mounts:
+# PE Cluster Note: By default will set all to shared, so if moving to PE dedicated cluster you should ask platform.sh
+# Support to make sure at least cache + logs is local, while you can let web/var be shared if you prefer that over DFS.
```
DominikaK:
```suggestion
# Support to make sure at least cache + logs are local, while you can let web/var be shared if you prefer that over DFS.
```

.platform.app.yaml
```diff
@@ -85,9 +92,10 @@ mounts:
 #    'src/AppBundle/MigrationVersions/References':
 #        source: local
 #        source_path: MigrationVersionsReferences
-# Uncomment if you want to use DFS clustering, WARNING: Needs to be shared, contact Platform.sh support to set this up
+# Uncomment if you want to use DFS clustering, required if using legacy-bridge on PE dedicated cluster setup.
```
DominikaK:
```suggestion
# Uncomment if you want to use DFS clustering, required if using Legacy Bridge on PE dedicated cluster setup.
```

.platform.app.yaml
```diff
@@ -127,15 +138,23 @@ hooks:
             touch web/var/.platform.installed
         fi
 
-        # Now that mounts are available, clear class/container cache. E.g. in case of interface changes on lazy proxies
-        rm -Rf var/cache/$SYMFONY_ENV/*.*
+        # Now that mounts are available, clear cache on mount.
+        # NB! Skip on PE Cluster setup using e.g. "if [$PLATFORM_BRANCH" != 'production']; then" & get p.sh to enable this on internal per node "pre_start" hook
```
DominikaK:
```suggestion
        # Note: Skip on PE Cluster setup using e.g. "if [$PLATFORM_BRANCH" != 'production']; then" & get p.sh to enable this on internal per node "pre_start" hook
```

.platform.app.yaml
```diff
@@ -127,15 +138,23 @@ hooks:
...
+        ./bin/platformsh/pre_start-per_node-clear_cache.sh
 
-        # With that done, make sure to clear + warmup all remaining caches to take proper env variables into account
-        bin/console cache:clear
+        # If you also need to clear redis cache on every deploy, you can either use this command or redis-cli
```
DominikaK:
```suggestion
        # If you also need to clear Redis cache on every deploy, you can either use this command or redis-cli
```

.platform.app.yaml
```diff
@@ -127,15 +138,23 @@ hooks:
...
+        ##php bin/console cache:pool:clear cache.redis
 
         # Example of additional deploy hooks if you use doctrine and/or kaliop migration bundle
-        ##bin/console doctrine:migrations:migrate --no-interaction --allow-no-migration
-        ##bin/console kaliop:migration:migrate --no-interaction --no-debug
+        ##php bin/console doctrine:migrations:migrate --no-interaction --allow-no-migration
+        ##php bin/console kaliop:migration:migrate --no-interaction --no-debug
+
+        # When using Solr, there are two cases where you'll need to rebuild indexes:
+        # - When Solr / search configuration change
```
DominikaK:
```suggestion
        # - When Solr / search configuration changes
```

.platform/services.yaml
```diff
@@ -1,4 +1,10 @@
+# Default settings in order to setup eZ Platform demo install on Platform.sh dev instances
```
DominikaK:
```suggestion
# Default settings in order to set up eZ Platform demo install on Platform.sh dev instances
```

.platform/services.yaml
```diff
@@ -1,4 +1,10 @@
...
+#
+# NB: Like on own servers, make sure to tune Redis/Solr/Varnish/Mysql memory/disk size for your installation to avoid issues.
```
DominikaK:
```suggestion
# Note: Like on own servers, make sure to tune Redis/Solr/Varnish/MySQL memory/disk size for your installation to avoid issues.
```

.platform/services.yaml
```diff
@@ -1,4 +1,10 @@
...
+#     Reach out to platform.sh support to get help on this and insight into your disk/memory usage.
+
 mysqldb:
+    # NB! If you change this also change DATABASE_VERSION, as used by "doctrine.dbal.server_version" in config.yml.
```
DominikaK:
```suggestion
    # Note: If you change this also change DATABASE_VERSION, as used by "doctrine.dbal.server_version" in config.yml.
```

.platform/services.yaml
```diff
@@ -21,18 +27,30 @@ mysqldb:
 
 # For use by Symfony Cache (used by eZ Platform SPI Persistence Cache)
 rediscache:
-    type: 'redis:3.2'
+    type: 'redis:5.0'
+    # For cache you might need to increase the size of your plan if your install sizeable amount of content.
```
DominikaK:
```suggestion
    # For cache you might need to increase the size of your plan if your installation has a sizeable amount of content.
```

bin/platformsh/pre_start-per_node-clear_cache.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# This script gets run as part of the .platform.app.yaml deploy step
```
DominikaK:
```suggestion
# This script is run as part of the .platform.app.yaml deployment step
```

bin/platformsh/pre_start-per_node-clear_cache.sh
```diff
@@ -0,0 +1,14 @@
...
+# On PE Cluster (usually just production) this should be setup by P.sh team as part of pre_start event
```
DominikaK:
```suggestion
# On PE Cluster (usually just production) this should be setup by platform.sh team as part of pre_start event
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,92 @@
...
-> **Beta**: Instructions and Tools *(Platform.sh configuration files, scripts, ...)* described on this page are currently in Beta for community testing & contribution, and may change without notice.
-
 ## Installation using eZ Platform template
 This is the simplest approach, but may be less up to date with current development than the other, more manual approach.
 
+### Using eZ Platform Cloud
+With eZ Platform Cloud subscription, you are able to create an instance of both open source and Enterprise using templates. In the future it also allow you to create an instance of Commerce version.
+
+To use eZ Platform Cloud:
+1. Login at [cloud.ezplatform.com](https://cloud.ezplatform.com/)
```
DominikaK:
```suggestion
1. Log in to [cloud.ezplatform.com](https://cloud.ezplatform.com/)
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,92 @@
 # Install eZ Platform on Platform.sh
 
-> **Beta**: Instructions and Tools *(Platform.sh configuration files, scripts, ...)* described on this page are currently in Beta for community testing & contribution, and may change without notice.
-
 ## Installation using eZ Platform template
 This is the simplest approach, but may be less up to date with current development than the other, more manual approach.
 
+### Using eZ Platform Cloud
+With eZ Platform Cloud subscription, you are able to create an instance of both open source and Enterprise using templates. In the future it also allow you to create an instance of Commerce version.
```
DominikaK:
```suggestion
With eZ Platform Cloud subscription, you are able to create an instance of both open source and Enterprise using templates. In the future it will also enable you to create an instance of eZ Commerce.
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,92 @@
...
+
+To use eZ Platform Cloud:
+1. Login at [cloud.ezplatform.com](https://cloud.ezplatform.com/)
+2. Click to Add a project, during setup wizard you'll get the choice to pick a flavour of eZ Platform.
```
DominikaK:
```suggestion
2. Click Add a project, during setup wizard you'll get the choice to pick a flavor of eZ Platform.
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,92 @@
...
+3. Complete the setup wizard, and your eZ Platform site will be created.
+
+### Using vanilla platform.sh
+Using an platform.sh account, you can create an instance of open source version of eZ Platform from template:
```
DominikaK:
```suggestion
Using a platform.sh account, you can create an instance of open source version of eZ Platform from template:
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,92 @@
...
-1. Create a Platform.sh project, using the "Import your existing code" option. Follow the setup wizard, but halt at the end, before clicking "Finish".
-1. Fork [eZ Platform](https://github.com/ezsystems/ezplatform/) and clone your fork locally.
-1. Add the platform remote of your project, and push your branch. The Platform.sh setup wizard tells you the command to use. Example:  
-   `git remote add platform my_project@git.eu.platform.sh:my_project.git`  
-   1. Optional: Set the SYMFONY_ENV environment variable to 'prod' or 'dev':  
-      `export SYMFONY_ENV=dev`  
-      If you don't do this, the local build will default to 'prod'.  
-      **NB:** For this to affect remote builds, it must also be set as a Platform.sh project variable, see below.
-1. Push your branch. The Platform.sh setup wizard tells you the command to use. Example:  
+1. Login or create an account at [Platform.sh](https://platform.sh) _(or login to [cloud.ezplatform.com](https://cloud.ezplatform.com/) if you have an subscription)_
```
DominikaK:
```suggestion
1. Log in or create an account at [Platform.sh](https://platform.sh) _(or login to [cloud.ezplatform.com](https://cloud.ezplatform.com/) if you have an subscription)_
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,92 @@
...
+2. Unless this has already been done for you, create a new project by using the **Import your existing code** option. Follow the setup wizard, but halt at the end, before clicking **Finish**.
+3. Fork [eZ Platform](https://github.com/ezsystems/ezplatform/) _(or [eZ Platform Enterprise](https://github.com/ezsystems/ezplatform-ee/)/[eZ Commerce](https://github.com/ezsystems/ezcommerce/) if you have a subscription)_ and clone your fork locally.
+4. Add the platform remote of your project, and push your branch. The Platform.sh setup wizard provides the command to use. Example:
+   `git remote add platform my_project@git.eu.platform.sh:my_project.git`
+5. [Optional] Authentication against Github/Bitbucket/Gitlab/updates.ez.no for locale development:
+   If you have private packages from own git repositories or use eZ Platform Enterprise, you can use Composer's
```
DominikaK:
```suggestion
   If you have private packages from your own git repositories or use eZ Platform Enterprise, you can use Composer's
```

doc/platformsh/INSTALL.md
```diff
@@ -1,48 +1,92 @@
...
+
+Yes, like all configuration (YML/VCL/..) bundled with the eZ Platform installation, all config in the "root project" is yours, and for you to customize for your needs.
+The bundled config is a recommended safe default that you can start from, just make sure to make it possible to merge in future config changes when you later need to upgrade.
+
+## Using Enterprise Dedicated Cluster
+
+If you are on a Platform.sh Enterprise Dedicated Cluster setup, typically named PE-6 / PE-12 / (...), you'll need to do some adjustment on the platform.sh config and you'll
+have dedicated on-boarding where topics around using eZ Platform on this cluster will be gone true before you can deploy to it.
+
+If you want to read up a bit, look in the bundled platform.sh config. There are a lot of config commented out by default, some of these are explicitly for Dedicated Cluster setup.
```
DominikaK:
```suggestion
If you want to read up a bit, look in the bundled platform.sh config. There is a lot of config commented out by default, some of these are explicitly for Dedicated Cluster setup.
```

.platform/services.yaml
```diff
@@ -21,18 +27,30 @@ mysqldb:
...
+#    type: 'redis:5.0'
+#    configuration:
+#        maxmemory_policy: allkeys-lru
 #
 # Alternatively if you have a requirement that sessions are persisted across server/redis restarts,
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
-#    type: redis-persistent:3.2
+#    type: redis-persistent:5.0
 #    disk: 512
```
vidarl:
```suggestion
 # Disk size should be bigger than redis'  `maxmemory` setting due to https://redis.io/topics/persistence#log-rewriting.
 # The memory given to redis is dependend on your plan and `size: `. Addjust `disk: ` accordingly
 #    disk: 512
```

.platform.app.yaml
alustenberg: LGTM!

.platform/services.yaml
andrerom: @vidarl updated

.platform/services.yaml
```diff
@@ -21,18 +27,32 @@ mysqldb:
...
+#    configuration:
+#        maxmemory_policy: allkeys-lru
 #
 # Alternatively if you have a requirement that sessions are persisted across server/redis restarts,
 # have storage space to spare for this, and don't mind a bit slower instance type of redis
 #redissession:
-#    type: redis-persistent:3.2
-#    disk: 512
+#    type: redis-persistent:5.0
+ # Disk size should be bigger than redis'  "maxmemory" setting due to https://redis.io/topics/persistence#log-rewriting.
```
DominikaK:
```suggestion
 # Disk size should be bigger than Redis'  "maxmemory" setting due to https://redis.io/topics/persistence#log-rewriting.
```

.platform/services.yaml
```diff
@@ -21,18 +27,32 @@ mysqldb:
...
+ # The memory given to redis is dependend on your plan and "size: ". Adjust "disk: " accordingly
```
DominikaK:
```suggestion
 # The memory given to Redis depends on your plan and "size: ". Adjust "disk: " accordingly.
```

.platform/services.yaml
andrerom: Oh, and FYI doing so in master failed due to logic in `platformsh/symfonyflex-bridge`: https://github.com/platformsh/symfonyflex-bridge/issues/29

PR #450, opened 2019-08-22
EZP-30915: [platform.sh]Fix for node version

damianz5:
[platform.sh]Fix for node version

https://jira.ez.no/browse/EZP-30915

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -92,6 +92,13 @@ hooks:
     build: |
         set -e
 
+        unset NPM_CONFIG_PREFIX
+        curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | dash
+        export NVM_DIR="$HOME/.nvm"
+        [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+        nvm current
+        nvm install 10.15.3
+
         # Configure GITHUB token if set. For use if you have own private repos, or to avoid API rate limits.
         if [ -n "$GITHUB_TOKEN" ]; then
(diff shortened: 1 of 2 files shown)


.platform.app.yaml
```diff
@@ -92,6 +92,13 @@ hooks:
     build: |
         set -e
 
+        unset NPM_CONFIG_PREFIX
+        curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | dash
+        export NVM_DIR="$HOME/.nvm"
+        [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+        nvm current
+        nvm install 12.6.0
```
SylvainGuittard: Second review. We recommend `Node.js 10.15.3 LTS` 
https://doc.ezplatform.com/en/latest/getting_started/requirements/#recommended-setups

.environment
```diff
@@ -0,0 +1,7 @@
+# This is necessary for nvm to work.
```
andrerom: Anyway this can be placed in p.sh config/directory? Given it is p.sh specific.

damianz5:
Updated - only necessary changes have been made

andrerom:
@SylvainGuittard He moved it to build / deploy.

@damianz5 With the updated webpack encore, does it still now support the newer node version which is used by default? (its v12 right?)

damianz5:
default node is 'node -v v6.17.1'.
we can omit .environment because its node/yarn is needed only build/deploy stage - no need to set the version for app itself

damianz5:
ping @andrerom final review please, currently it contains only changes from https://docs.platform.sh/languages/nodejs/nvm.html and version bump

PR #454, opened 2019-09-06
Added support for COMPOSER_AUTH

jheba:
ref. https://getcomposer.org/doc/03-cli.md#composer-auth

I found it convenient for automated builds when you need to use different creds for updates.ez.no depending on what product images you build (for example eZC pulls TTL packages while eZP and eZPEE BUL packages).


jheba:
@andrerom please also give feedback

andrerom:
Cherry picked in https://github.com/ezsystems/ezplatform/commit/183229be0e85edcecefee70409d03a6ec59bbd3c

PR #455, opened 2019-09-17
Docker on 3.x

vidarl:
According to this [doc](https://symfony.com/doc/current/configuration.html#configuration-based-on-environment-variables) : `The existing env vars are never overwritten by the values defined in .env, so you can combine both.` That seems to be only partially true..

So in .env we have :

```
DATABASE_URL=${DATABASE_PLATFORM}://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_NAME}
```

Variables like `DATABASE_NAME` and `DATABASE_PORT` are also set in the `.env` file..
However, it does not work to then change `DATABASE_NAME` via environment variable. It must be changed in `.env` too, because `DATABASE_URL` is defined there. Or, as I do here, redefine `DATABASE_URL` instead

~note : https://github.com/ezsystems/ezplatform/pull/464 should be merged first, and I'll rebase this PR...´~

### DIFF
--- a/.env
+++ b/.env
@@ -6,6 +6,9 @@ COMPOSE_DIR=.
 COMPOSER_HOME=~/.composer
 COMPOSER_MEMORY_LIMIT=4G
 
+# Feel free to change project name to something meaningful (preferably unique per project)
+COMPOSE_PROJECT_NAME=ezplatform
+
 ## Docker images (name and version)
 PHP_IMAGE=ezsystems/php:7.3-v1
 PHP_IMAGE_DEV=ezsystems/php:7.3-v1-dev
--- a/doc/docker/install_script.sh
+++ b/doc/docker/install_script.sh
@@ -12,7 +12,6 @@ if [ "$s" != "0" ]; then
 fi
 mkdir -p public/var
 
-
 if [ "${INSTALL_DATABASE}" == "1" ]; then 
     export DATABASE_URL=${DATABASE_PLATFORM}://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_NAME}
 
--- a/doc/docker/install_script.sh
+++ b/doc/docker/install_script.sh
@@ -12,7 +12,10 @@ if [ "$s" != "0" ]; then
 fi
 mkdir -p public/var
 
+
 if [ "${INSTALL_DATABASE}" == "1" ]; then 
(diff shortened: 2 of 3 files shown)


.env
```diff
@@ -87,3 +87,7 @@ MAILER_URL=null://localhost
 ###> nelmio/cors-bundle ###
 CORS_ALLOW_ORIGIN=^https?://localhost(:[0-9]+)?$
 ###< nelmio/cors-bundle ###
+
```
wizhippo: I would move this to the top with the reset of the `COMPOSE_*` variables.

I would also almost recommend defaulting it to ezplatform.  As it stands now by default it ends up docker which is a bit confusing.

doc/docker/install_script.sh
```diff
@@ -6,6 +6,8 @@
 for i in $(seq 1 3); do composer install --no-progress --no-interaction --prefer-dist --no-suggest --optimize-autoloader && s=0 && break || s=$? && sleep 1; done; (exit $s);
 mkdir -p public/var;
 
+export DATABASE_URL=${DATABASE_PLATFORM}://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_NAME}
```
wizhippo: Nitpick missing `;`

.env
vidarl: good idea.. Fixed

doc/docker/install_script.sh
vidarl: Actually, let's remove all that superfluous `; `, https://github.com/ezsystems/ezplatform/pull/464

.env
```diff
@@ -6,6 +6,9 @@ COMPOSE_DIR=.
 COMPOSER_HOME=~/.composer
 COMPOSER_MEMORY_LIMIT=4G
 
+#Feel free to change project name to something meaningful
```
andrerom:
```suggestion
# Feel free to change project name to something meaningful (preferably unique per project)
```

doc/docker/install_script.sh
```diff
@@ -12,6 +12,8 @@ if [ "$s" != "0" ]; then
 fi
 mkdir -p public/var
 
+export DATABASE_URL=${DATABASE_PLATFORM}://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_NAME}
```
andrerom: This can be moved inside `if` right?

vidarl: yeah. fixed

andrerom:
@vidarl / @damianz5 This is now on all relevant repos (ee, demo, ee-demo).

PR #456, opened 2019-09-18
[Varnish] Use cookie vmod to filter cookies

SerheyDolgushev:
This PR provides a much simpler way to filter the cookies on Varnish side

### DIFF
--- a/.platform/varnish.vcl
+++ b/.platform/varnish.vcl
@@ -9,6 +9,7 @@
 //vcl 4.1;
 //import std;
 import xkey;
+import cookie;
 
 // For customizing your backend and acl rules see parameters.vcl
 // Includes not available on Platform.sh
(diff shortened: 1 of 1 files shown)


.platform/varnish.vcl
```diff
@@ -60,14 +61,10 @@ sub vcl_recv {
...
     // Remove all cookies besides Session ID, as JS tracker cookies and so will make the responses effectively un-cached
     if (req.http.cookie) {
-        set req.http.cookie = ";" + req.http.cookie;
-        set req.http.cookie = regsuball(req.http.cookie, "; +", ";");
-        set req.http.cookie = regsuball(req.http.cookie, ";(eZSESSID[^=]*)=", "; \1=");
-        set req.http.cookie = regsuball(req.http.cookie, ";[^ ][^;]*", "");
-        set req.http.cookie = regsuball(req.http.cookie, "^[; ]+|[; ]+$", "");
-
+        cookie.parse(req.http.cookie);
+        cookie.filter_except("eZSESSID");
```
andrerom: Does this also cover `eZSESSIDSiteAccessName` cases?

SerheyDolgushev: @andrerom nope. You need to whitelist separately cookie for each siteaccess. We can add a comment regarding this. Or would you prefer to close this PR instead?

andrerom: I think I prefer to close this as long as the vmod does not support regex/pattern/prefix based cookie whitelist.

PR #458, opened 2019-09-23
Added graphql schema generation after database install

mnocon:
After https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/200 GraphQl is always used in the product and the schema must be generated before AdminUI can be used.


mnocon:
@ezsystems/documentation-team something for you to vertify, but I think we need to update our installation procedure to mention this step

mnocon:
The fix for tests is https://github.com/ezsystems/ezplatform-graphql/pull/55, but you can see in the log that the scheme is correctly generated

mnocon:

PR #460, opened 2019-09-27
GraphQL schema generator in installer + p.sh directory

damianz5:
Added graphql schema generator for installer and p.sh shared directory for graphql schema

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -80,6 +80,9 @@ mounts:
     'public/var':
         source: local
         source_path: var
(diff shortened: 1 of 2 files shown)


.platform.app.yaml
```diff
@@ -80,6 +80,9 @@ mounts:
     'public/var':
         source: local
         source_path: var
+    'config/graphql/types':
```
SylvainGuittard: Maybe we can have something like that:
```suggestion
    'config/graphql/types/ezplatform':
```

PR #468, opened 2019-10-03
[docker] Increase composer timeout limit on install assets

wizhippo:
Installing assets always fails for me as composer hits it's time limit.

Adding `COMPOSER_PROCESS_TIMEOUT` solves this issue


wizhippo:
working now without

PR #470, opened 2019-10-14
[Docker] Allowed to set PHP memory limit through env

mnocon:
Required by https://github.com/ezsystems/ezplatform-page-builder/pull/406 to fix repository-forms on EE demo tests, which are failing because of memory issues.


mnocon:
@andrerom please have a look, as this is what we've discussed in https://github.com/ezsystems/docker-php/pull/45

andrerom:
@vidarl maybe better if you review here

mnocon:

PR #490, opened 2019-12-13
EZP-30750: Added possibility to set purge_type with environmental variables

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-30750


adamwojs:
CI failure seems to be releated to PR changes.

webhdx:
CI tests using Varnish & Redis seem to be fine: https://github.com/ezsystems/ezplatform-page-builder/pull/453

config/packages/ezplatform.yaml
```diff
@@ -1,19 +1,16 @@
 imports:
+    - { resource: env/generic.php }
+    - { resource: env/platformsh.php }
+    - { resource: cache_pool/cache.memcached.yaml }
```
emodric: Why uncommenting of memcached and redis pool services?

webhdx: Yes it shouldn't be done this way but I didn't know that `env/generic.php` was loading those files based on `CACHE_POOL` value. It wasn't working tho so I fixed it and removed those imports completely.

webhdx:
Ping @mnocon. Ready for another round of testing. I think we should also test if p.sh instances are configured correctly as I had to move/fix some of its configuration.

lserwatka:
Well done @webhdx 🥇

lserwatka:
@webhdx could you merge it horizontally?

webhdx:
Merged

PR #495, opened 2020-01-19
Adapt 'platform variable:create' to changed syntax

reithor:
Syntax in platform CLI according to:
https://docs.platform.sh/tutorials/composer-auth.html#set-the-envcomposerauth-project-variable

### DIFF
--- a/doc/platformsh/INSTALL.md
+++ b/doc/platformsh/INSTALL.md
@@ -42,7 +42,7 @@ This requires more manual steps, but may be more up to date with current develop
        For example set the project variables for your eZ Network installation ID and token:
       `platform variable:create --level project --name env:COMPOSER_AUTH --json true --visible-runtime false --value '{"http-basic": {"updates.ez.no": {"username": "network-id", "password": "token"}}}'`
    4. If you have the need to debug things remotely, set the `SYMFONY_ENV` environment variable to 'dev':
-      `platform variable:update --name env:SYMONY_ENV --value 'dev'`.
+      `platform variable:update env:SYMONY_ENV --value 'dev'`.
 7. Push your branch. The Platform.sh setup wizard provides the command to use. Example:
    `git push -u platform master`  
    This starts the build process. Now, finish the Platform.sh setup wizard.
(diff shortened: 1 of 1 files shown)


lserwatka:
@vidarl could you review this PR?

doc/platformsh/INSTALL.md
```diff
@@ -40,9 +40,9 @@ This requires more manual steps, but may be more up to date with current develop
       Run `platform get <your project id>`
    3. Authentication against Github/Bitbucket/Gitlab/updates.ez.no
        For example set the project variables for your eZ Network installation ID and token:
-      `platform project:variable:create env:COMPOSER_AUTH '{"http-basic":{"updates.ez.no":{"username":"network-id","password":"token-key"}}}' --no-visible-runtime --sensitive true`
+      `platform variable:create --level project --name env:COMPOSER_AUTH --json true --visible-runtime false --value '{"http-basic": {"updates.ez.no": {"username": "network-id", "password": "token"}}}'`
    4. If you have the need to debug things remotely, set the `SYMFONY_ENV` environment variable to 'dev':
-      `platform project:variable:set env:SYMONY_ENV dev`.
+      `platform variable:update env:SYMONY_ENV --value 'dev'`.
```
vidarl, line 45: I am a bit fuzzed by the use of `update` here.. We set the the SYMFONY_ENV in `.platform.app.yaml`. Have you confirmed that setting this command actually works? Honestly, I couldn't find in p.sh docs if environment variables defined in `.platform.app.yaml` has precedence over project/environment variables or not.

vidarl (changes requested):

doc/platformsh/INSTALL.md
reithor: I have checked it - and it overrides the setting.
Syntax is slightly different though: 
```platform variable:update --value 'dev' env:SYMFONY_ENV```
 
But simply setting SYMFONY_ENV to dev won't work anyway  - in .platform.app.yaml we use composer install :
```composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader```
 so dev packages are not installed on p.sh by default

vidarl: yes, I know.. We had a comment about this in `.platform.app.yaml` earlier, but it was removed.
So, maybe add info about this in a  6.5 bullet ?

PR #496, opened 2020-01-23
Update .platform.app.yaml - set php memory limit

damianz5:
Update .platform.app.yaml - set php memory limit


andrerom:
@damianz5 Do we need this even when using imagick? (see #501)
If so then we have some form of memory leak or similar somewhere.

PR #501, opened 2020-02-11
EZP-31377: Use imagick by default on Platform.sh to avoid memory issues

andrerom:
| Question           | Answer
| **JIRA issue**     | [EZP-31377](https://jira.ez.no/browse/EZP-31377)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `2.5`
| **BC breaks**      | no
| **Tests pass**     | N/A
| **Doc needed**     | no

GD uses PHP memory and hence it's better if we can use imagick if it's available to avoid risking running into php memory issues on larger images.

**TODO**:
- [x] Implement feature / fix a bug.
- [x] Implement tests.
- [x] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.

### DIFF
--- a/app/config/default_parameters.yml
+++ b/app/config/default_parameters.yml
@@ -122,3 +122,6 @@ parameters:
     # Varnish invalidation/purge token (for use on platform.sh, eZ Platform Cloud and other places you can't use IP for ACL)
     varnish_invalidate_token: '%env(HTTPCACHE_VARNISH_INVALIDATE_TOKEN)%'
     env(HTTPCACHE_VARNISH_INVALIDATE_TOKEN): ~
+    
+    # LiipImagineBundle - imagine driver
+    liip_imagine.driver: 'gd'
--- a/app/config/default_parameters.yml
+++ b/app/config/default_parameters.yml
@@ -124,4 +124,4 @@ parameters:
     env(HTTPCACHE_VARNISH_INVALIDATE_TOKEN): ~
     
     # LiipImagineBundle - imagine driver
-    liip_imagine.driver: 'gd'
+    liip_imagine_driver: 'gd'
--- a/app/config/env/platformsh.php
+++ b/app/config/env/platformsh.php
@@ -163,7 +163,8 @@ if (!getenv('HTTPCACHE_VARNISH_INVALIDATE_TOKEN')) {
     $container->setParameter('varnish_invalidate_token', getenv('PLATFORM_PROJECT_ENTROPY'));
 }
 
-// Detect if imagick is enabled, if so get imagine to use it to reduce memory use (compared to GD which uses PHP memory)
+// Adapt config based on enabled PHP extensions
+// Get imagine to use imagick if enabled, to ravoid using php memory for image convertions
 if (extension_loaded('imagick')) {
-    $container->setParameter('liip_imagine.driver', '');
+    $container->setParameter('liip_imagine.driver', 'imagick');
 }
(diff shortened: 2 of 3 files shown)


app/config/env/platformsh.php
```diff
@@ -162,3 +162,9 @@
 if (!getenv('HTTPCACHE_VARNISH_INVALIDATE_TOKEN')) {
     $container->setParameter('varnish_invalidate_token', getenv('PLATFORM_PROJECT_ENTROPY'));
 }
+
+// Adapt config based on enabled PHP extensions
+// Get imagine to use imagick if enabled, to ravoid using php memory for image convertions
```
konradoboza:
```suggestion
// Get imagine to use imagick if enabled, to avoid using php memory for image convertions
```

andrerom:
@vidarl On closer look those settings have been deprecated since 5.4:
- https://github.com/ezsystems/ezpublish-kernel/blob/0eed2fe2d8e4f531bb1fc811a32cc54b90a2e7f4/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Configuration.php#L299-L308
- https://github.com/ezsystems/ezpublish-kernel/blob/c849ef16ba5aecb786f61267855e24a5cf62bcf6/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Configuration/Parser/Image.php#L40
- https://github.com/ezsystems/ezpublish-kernel/blob/d4542d36e30022067f788ba6654984851cfb2ef2/eZ/Bundle/EzPublishCoreBundle/Resources/config/default_settings.yml#L171


What made you think this should be used? Are there some doc still referring to it?

andrerom:
Ready for testing again then @micszo

andrerom:
@damianz5 Ok, do you see a way we can set it dynamically then?

damianz5:
Updated - ready for retest

app/config/default_parameters.yml
```diff
@@ -122,3 +122,6 @@ parameters:
     # Varnish invalidation/purge token (for use on platform.sh, eZ Platform Cloud and other places you can't use IP for ACL)
     varnish_invalidate_token: '%env(HTTPCACHE_VARNISH_INVALIDATE_TOKEN)%'
     env(HTTPCACHE_VARNISH_INVALIDATE_TOKEN): ~
+    
+    # LiipImagineBundle - imagine driver
+    liip_imagine.driver: 'gd'
```
andrerom:
```suggestion
    liip_imagine_driver: 'gd'
```

app/config/env/platformsh.php
```diff
@@ -162,3 +162,9 @@
 if (!getenv('HTTPCACHE_VARNISH_INVALIDATE_TOKEN')) {
     $container->setParameter('varnish_invalidate_token', getenv('PLATFORM_PROJECT_ENTROPY'));
 }
+
+// Adapt config based on enabled PHP extensions
+// Get imagine to use imagick if enabled, to avoid using php memory for image convertions
+if (extension_loaded('imagick')) {
+    $container->setParameter('liip_imagine.driver', 'imagick');
```
andrerom:
```suggestion
    $container->setParameter('liip_imagine_driver', 'imagick');
```

andrerom:
Thanks @damianz5 👍 👍 

PS:  I took the liberty of removing the dot in the param name now to make to more clear it's global param.

PR #502, opened 2020-02-14
[Platform.sh] Updated Maria DB version to 10.4

SerheyDolgushev:
Doctrine version handling for Maria DB was fixed in https://github.com/platformsh/symfonyflex-bridge/releases/tag/2.2.0. And it lets to use `10.4` without any errors/issues.


SerheyDolgushev:
@Crell tagging you here just in case :)

Crell:
Looks good from my end.

lserwatka:
@damianz5 could you mere it horizontally?

PR #514, opened 2020-03-26
EZP-31519: Moved app.php rejection snippet from front controller to ezpublish-kernel

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-31519

I've changed the rule in `vhost.template` files in order to unify logic in both places. This way we can block some requests on httpd level.

Rejection logic has been moved to: https://github.com/ezsystems/ezpublish-kernel/pull/3010

### DIFF
--- a/doc/apache2/vhost.2.2.template
+++ b/doc/apache2/vhost.2.2.template
@@ -94,7 +94,7 @@
         RewriteRule ^/(css|js|fonts?)/.*\.(css|js|otf|eot|ttf|svg|woff) - [L]
 
         # Prevent access to website with direct usage of app.php in URL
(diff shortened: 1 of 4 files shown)


glye:
As I recall, this is as intended. The point is to block all direct access to app.php, while not blocking content named ´*app.php*´. If in doubt, better to block too much than too little.

webhdx:
I just moved the logic from [EZP-30716](https://jira.ez.no/browse/EZP-30716). If this should be changed then we need new ticket and this should be done after 3.0 release as this is the highest priority at the moment. I need this to be merged in order to merge it up to `2.5` and `master`.

ping @lserwatka

webhdx:
@mnocon Can I get final approval from you? We need to merge it.

PR #522, opened 2020-04-01
Update comments about HttpCache purge type

adriendupuis:
This may be a minor forgotten non-updated misleading comment. 'http' is a value that can't be used anymore. It triggers the error "No driver found being able to handle purge_type 'http'."

During migration from v2.5.9 to v3.0.0-rc.1, I replace it with 'varnish' which seems OK.

### DIFF
--- a/config/packages/ezplatform.yaml
+++ b/config/packages/ezplatform.yaml
@@ -39,7 +39,7 @@ parameters:
     purge_type: local
 
 ezplatform:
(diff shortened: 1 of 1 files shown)


micszo:
@ViniTou fyi

xrow:
I think an update is needed here: https://doc.ezplatform.com/en/master/guide/http_cache/

andrerom:
@xrow https://github.com/ezsystems/developer-documentation/pull/1068

PR #530, opened 2020-04-07
EZP-31541: Added new CACHE_NAMESPACE env to redis docker stack

ViniTou:
https://jira.ez.no/browse/EZP-31541
followup to: https://github.com/ezsystems/ezplatform/pull/527

### DIFF
--- a/doc/docker/redis.yml
+++ b/doc/docker/redis.yml
@@ -12,6 +12,7 @@ services:
     environment:
      - CACHE_POOL=cache.redis
      - CACHE_DSN=redis:6379
+     - CACHE_NAMESPACE
 
   redis:
     image: ${REDIS_IMAGE}


doc/docker/redis.yml
```diff
@@ -12,6 +12,7 @@ services:
     environment:
      - CACHE_POOL=cache.redis
      - CACHE_DSN=redis:6379
+     - CACHE_NAMESPACE
```
michal-myszka, line 15: Maybe default namespace name should be provided here?

ViniTou: It is in .env, not sure if we should duplicate that

mnocon: For me `.env` is enough

lserwatka:
@ViniTou could you merge it to EE?

PR #536, opened 2020-04-27
platformsh performance tweeks

ITernovtsiy:
Add recommended configs from 
https://symfony.com/doc/current/performance.html 
https://docs.platform.sh/languages/php.html#opcache-preloading

Let me know if we should stay with PHP 7.3 and I'll remove the `opcache.preload` part.

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -43,7 +43,9 @@ variables:
         # The default OPcache configuration is not suited for Symfony applications
         opcache.memory_consumption: 256
         opcache.max_accelerated_files: 20000
-        opcache.validate_timestamps: 0
+        # We recommend enabling the following opcache.validate_timestamps setting in production, but then opcache_reset() must be called every time you clear symfony cache.
+        #opcache.validate_timestamps: 0
+
         # Applications that open many PHP files, such as Symfony projects, should use at least these values
         realpath_cache_size: 4096K
         realpath_cache_ttl: 600
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -40,6 +40,13 @@ variables:
         # Example of setting php.ini config
         #"display_errors": "On"
         memory_limit: 512M
+        # The default OPcache configuration is not suited for Symfony applications
+        opcache.memory_consumption: 256
+        opcache.max_accelerated_files: 20000
+        opcache.validate_timestamps: 0
+        # Applications that open many PHP files, such as Symfony projects, should use at least these values
+        realpath_cache_size: 4096K
+        realpath_cache_ttl: 600
     env:
         # We disable Symfony Proxy (CacheKernel), as we rather use Varnish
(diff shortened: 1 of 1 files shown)


ITernovtsiy:
Removed part related to `opcache.preload` and php7.4, as it's failing on platform.sh with `eZ\Publish\Core\MVC\Symfony\Routing\ChainRouter` preloaded

lserwatka:
@vidarl could you review this PR?

lserwatka:
@ITernovtsiy we need to stay with PHP 7.3. Could you change a branch to 3.0?

.platform.app.yaml
```diff
@@ -40,6 +40,13 @@ variables:
         # Example of setting php.ini config
         #"display_errors": "On"
         memory_limit: 512M
+        # The default OPcache configuration is not suited for Symfony applications
+        opcache.memory_consumption: 256
+        opcache.max_accelerated_files: 20000
+        opcache.validate_timestamps: 0
```
vidarl, line 46: With `opcache.validate_timestamps: 0`, clearing cache kinda won't work, right?
Or do we/symfony call `opcache_reset()` on `cache:clear` ?

ITernovtsiy: this should work fine on platform sh, as they are using read-only filesystem for code.
the only way to update code is to perform deploy, which will redeploy container and clear opcache.
we're using this on few projects and didn't have any issues with code updates.

ITernovtsiy:
@lserwatka sure, rebased to `3.0`

.platform.app.yaml
vidarl: Symfony cache is stored (and is writeable ) in `var/cache`  and consists of `.php` files. For production use I agree that `opcache.validate_timestamps: 0` is a good setting, but I am afraid it will confuse many if we apply that by default

.platform.app.yaml
```diff
@@ -40,6 +40,13 @@ variables:
...
+        # Applications that open many PHP files, such as Symfony projects, should use at least these values
+        realpath_cache_size: 4096K
+        realpath_cache_ttl: 600
```
vidarl, line 49:
```suggestion
        realpath_cache_ttl: 600
        # We recommend enabling the following opcache.validate_timestamps setting in production, but then opcache_reset() must be called every time you clear symfony cache.
        #opcache.validate_timestamps: 0
```

.platform.app.yaml
vidarl, line 46:
```suggestion
        # We recommend enabling the following opcache.validate_timestamps setting in production, but then opcache_reset() must be called every time you clear symfony cache.
        #opcache.validate_timestamps: 0

```

lserwatka:
@mnocon could you merge it up and horizontally?

lserwatka:
Thank you @ITernovtsiy and @vidarl

PR #542, opened 2020-05-12
EZP-31627: add ability to define APP_HTTP_CACHE in .env file

ITernovtsii:
Due to direct `getenv` usage, it's not possible to define APP_HTTP_CACHE value with `.env` file, only by setting the system environment variable.
I think it makes sense to have it consistent with other configuration variables (unless this was intentionally implemented in this way).


andrerom:
@webhdx Is this from upstream and needs to be applied in symfony or is this logic on our side still?

PR #543, opened 2020-05-13
Bring back SYMFONY_TMP_DIR (in another form) for Docker optimization

Plopix:
There is a slack discussion in #ez-launchpad.
But I prefer to open that PR to the community.

Without this, there is no way to tell Docker to save the Symfony cache in the container rather than in the host.

On Mac OS this is a performance killer.

On 2.5 we still have it, why has it been removed?
https://github.com/ezsystems/ezplatform/blob/2.5/app/AppKernel.php#L84

Thanks!

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -46,4 +46,19 @@ class Kernel extends BaseKernel
         $routes->import($confDir.'/{routes}/*'.self::CONFIG_EXTS, '/', 'glob');
         $routes->import($confDir.'/{routes}'.self::CONFIG_EXTS, '/', 'glob');
     }
+
+    // can be removed when bumping to Symfony 5.2
+    public function getCacheDir()
+    {
+        if (isset($_SERVER['APP_CACHE_DIR'])) {
+            return $_SERVER['APP_CACHE_DIR'].'/'.$this->environment;
+        }
+        return parent::getCacheDir();
+    }
+
+    // can be removed when bumping to Symfony 5.2
+    public function getLogDir()
+    {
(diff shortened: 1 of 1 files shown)


SerheyDolgushev:
@Plopix maybe it also worth to add those variables in default `.env` file?

SerheyDolgushev:
I mean just two variables for cache and logs dirs?

Plopix:
note sure about complexifying more the `.env` as to me, there are already too much info in it. but yes that's a good point ;-)

lserwatka:
@Plopix could you rebase for 3.0?

webhdx:
Hello @Plopix. I removed it because we don't want to do any changes to `Kernel.php`. We intent to use Flex in the future so any change here won't be available to installations using Flex to install eZ Platform on existing Symfony projects. I think it would be a good contribution to Symfony.

Plopix:
ok @webhdx https://github.com/symfony/symfony/pull/37114 let's see

Plopix:
it's going to be approved I guess but that would be for Symfony 5.2 anyway. It might make sense to anticipate in eZ for now on. (instead of waiting eZ 3.2 or even 3.3)
what do you think?

Plopix:
Please wait a bit before merge, maybe we need: https://github.com/symfony/symfony/pull/37232

lserwatka:
@Plopix is anything holding us back to merge it?

lserwatka:
@mnocon could you merge it up and sideways?

lserwatka:
Thank you @Plopix 👍

PR #546, opened 2020-05-26
[Platform.sh] Changed default version of Solr to 7.7

mnocon:
Changing the default version of Solr used on Platform.sh to 7.7.

The current Solr instruction for Platform.sh states:
```
# If you wish to use solr, uncomment this service and the corresponding relationship in .platform.app.yaml.
# Also, you need to generate the config using:
# vendor/ezsystems/ezplatform-solr-search-engine/bin/generate-solr-config.sh
# Multi core setup is currently not supported on Platform.sh. Sharding does not work as the cores are
# unable to reach each other
```

But it's not working out of the box, because the `generate-config` script does not support Solr 6.6: https://github.com/ezsystems/ezplatform-solr-search-engine/blob/3.0/bin/generate-solr-config.sh#L31 (which is the version used by default).

After changing to 7.7 everything works as expected.

7.7 is the recommended version to use with solr-bundle 3.0: https://github.com/ezsystems/ezplatform-solr-search-engine/blob/3.0/README.md
and is supported by Platform.sh: https://docs.platform.sh/configuration/services/solr.html

The only issue I see here could be related to upgrades - people upgrading from v2 to v3 will have to adapt their config manually (and refer to https://docs.platform.sh/configuration/services/solr.html#upgrading), should we mention it somewhere?

### DIFF
--- a/.platform/services.yaml
+++ b/.platform/services.yaml
@@ -57,7 +57,7 @@ rediscache:
 # Multi core setup is currently not supported on Platform.sh. Sharding does not work as the cores are
 # unable to reach each other
 #solrsearch:
-#    type: solr:6.6
+#    type: solr:7.7
 #    disk: 512
 #    configuration:
 #        configsets:


.platform/services.yaml
```diff
@@ -57,7 +57,7 @@ rediscache:
 # Multi core setup is currently not supported on Platform.sh. Sharding does not work as the cores are
 # unable to reach each other
 #solrsearch:
-#    type: solr:6.6
+#    type: solr:7.7
```
andrerom, line 60: So one problem here, @adriendupuis mention Solr 7.x is not available on P.sh Dedicated Cluster:
https://docs.platform.sh/configuration/services/solr.html

By default we use Grid for demos, so not a big problem merging this.
But maybe we should mention it inline here above?

vidarl: What do we do about that. eZ Platform 3 requires solr 7...
Is Solr7 support in the pipeline at p.sh?

mnocon:
> Ideally we should also backport bump to 2.5 but there we would then need to bump default solr bundle version to ^2.0 (bundle version)

I'd like to avoid doing it in this PR, 2.5.10 is coming very soon and this is something we'd like to avoid doing a moment before the release (I'm ok with doing it after).

.platform/services.yaml
mnocon: > By default we use Grid for demos, so not a big problem merging this.
But maybe we should mention it inline here above?

If I understand correctly Solr 7.x is the only version working currently with Platform 3.0, so this PR still feels like a step in right direction anyway (but I can add a comment like `Solr 7.x is not available on Platform.sh Dedicated Cluster and therefore cannot be used with eZ Platform 3.0 there`, but no idea how can we overcome it (except maybe by talking with Platform.sh about their plans).

andrerom: > so this PR still feels like a step in right direction anyway 

it does, we should merge this to more clearly move the issue towards p.sh..

andrerom:
> I'd like to avoid doing it in this PR, 2.5.10 is coming very soon and this is something we'd like to avoid doing a moment before the release (I'm ok with doing it after).


+1, we can backport later

.platform/services.yaml
andrerom: So I'm good with keeping this as-is on second thought

lserwatka:
We can’t do it like this. We can’t ship Solr 7.7 in maintenance release. This explode instances. This 2.0 package is not backward compatible as fare as I remember. This change for p.sh does not make sense without updating package in composer.json.

.platform/services.yaml
SylvainGuittard: I started to discuss it with Gauthier yesterday. I will keep you posted

andrerom:
@lserwatka we still need this on 3.x as default config does not work out of the box (see also duplicated or above)

lserwatka (approved):

lserwatka:
@micszo can we run round of QA?

micszo:
Sure, I'll try to cover it as next thing.

andrerom (approved):

.platform/services.yaml
```diff
@@ -57,7 +57,7 @@ rediscache:
...
 #    disk: 512
 #    configuration:
 #        configsets:
```
micszo, line 63: @andrerom question about the next line `mainconfig: !archive "configsets/solr6"` - should version be changed to 7 there as well?

micszo: `generate-solr-config.sh` will use whatever is set in `mainconfig:` but it could be confusing.

micszo: (Or maybe skip version digit there but then there's the issue of config being overwritten.)

micszo: Ok, nevermind. `generate-solr-config.sh` has this... peculiar approach to default parameter values.

micszo (approved):
QA Approved on platform.sh with v3.1.2 OSS, EE & EC.

lserwatka:
Could you merge it up and sideways?

micszo:
Done

PR #549, opened 2020-06-02
Add error routes in dev environment

emodric:
Enables usage of `/_error/xyz` route in dev environments.


adamwojs:
Fixed in #610 but nevertheless thank you for contribution :wink:

PR #558, opened 2020-06-10
Sort composer.json packages

adriendupuis:
fae938a911caba0472106bbb792c531bf2fcad1c wasn't following `"sort-packages": true,` directive.


mnocon:
Thanks @adriendupuis ! I've rebased this PR with 3.0 as the change is useful there as well. I will take care of merging this upstream/sideways.

SerheyDolgushev:
Was going to submit the same PR, before I realized it is fixed already :)

PR #565, opened 2020-06-16
[Composer] Removed beta stability requirement for friends-of-behat/sy…

mnocon:
…mfony-extension


lserwatka:
@mnocon you can merge it up and sideways. The same @beta tag we need to remove from BehatBundle

PR #568, opened 2020-06-19
Remove forcing of Symfony 5.0

emodric:
The issue with Symfony 5.1 was fixed and released in 5.1.1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -138,8 +138,7 @@
     },
     "extra": {
         "symfony": {
-            "allow-contrib": true,
-            "require": "5.0.*"
+            "allow-contrib": true
         },
         "branch-alias": {
             "dev-master": "3.0.x-dev"


lserwatka:
We are not going to remove it. SF 5.0.0 is meant for eZ Platform 3.0.x. We will upgrade to SF 5.1.0 in eZ Platform 3.1.0 that will be out in 2 weeks.

emodric:
For anyone wondering, Symfony 5.1 works fine with eZ Platform 3.0. Just remove the constraint from your project.

emodric:
Additionally, one more reason for not blocking install of Symfony 5.1: since Symfony 5, EOL dates and EOM dates for non-LTS Symfony releases is aligned, which means that in a month, eZ Platform 3.0 will have not only an unmaintained Symfony version, but also potentially unsecure one:

https://symfony.com/blog/symfony-maintenance-changes-for-standard-releases

Since eZ Platform roadmap is not aligned with Symfony one, there will be moments where eZ Platform releases will use potentially unsecure versions of Symfony.

PR #576, opened 2020-07-16
EZP-31744: Added DATABASE_VERSION variable to .env

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-31744

### Description

Added `DATABASE_VERSION` variable to `.env` to avoid server version lookups by doctrine. See https://symfony.com/doc/current/reference/configuration/doctrine.html#doctrine-dbal-configuration


.env
```diff
@@ -67,6 +67,9 @@ DATABASE_HOST=db
 DATABASE_PORT=3306
 DATABASE_PLATFORM=mysql
 DATABASE_DRIVER=pdo_mysql
+# Needed by Doctrine Bundle / ORM to avoid it opening connection during situations where there is no service yet.
+# See: https://symfony.com/doc/current/reference/configuration/doctrine.html#doctrine-dbal-configuration
+DATABASE_VERISON=mariadb-10.3.0
```
ViniTou, line 72: Please also add this also to base-dev.yml and base-prod.yml, and then in ezcommerce replace `DATABASE_SERVER_VERSION` with this?

ViniTou:
> we didn't specify this on purpose so Doctrine can do it for us
And I agree we should aim for this once more but, main reason for this is situation when during installing database is still not available and all would fail on container building (clear:cache, ezcommerce).

adamwojs:
PR updated according to @ViniTou  code review suggestions.

lserwatka:
@adamwojs could you merge it up and horizontally up to ezcommerce?

dabrt:
Documented as part of:
https://github.com/ezsystems/developer-documentation/pull/1122

PR #581, opened 2020-07-21
EZP-31607: Insecure default value of Varnish invalidate token

kaff:
https://jira.ez.no/browse/EZP-31607


lserwatka:
You need to change a target branch to 2.5 if affected. If not, then 3.1.

ViniTou:
@lserwatka 
Should in not be 3.0 as this is marked as security issue?

lserwatka:
It is not as per say. But yes, I meant 3.0. My mistake in previous comment.

PR #585, opened 2020-07-21
.platform.app.yaml: Update comment web→public

adriendupuis:
Update comment about  about (web|public)/var

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -82,7 +82,7 @@ disk: 3072
 # The mounts that will be performed when the package is deployed.
 mounts:
 # PE Cluster Note: By default will set all to shared, so if moving to PE dedicated cluster you should ask platform.sh
-#                  Support to make sure at least cache + logs are local, while you can let web/var be shared if you prefer that over DFS.
+#                  Support to make sure at least cache + logs are local, while you can let public/var be shared if you prefer that over DFS.
     'var/cache':
         source: local
         source_path: cache


adriendupuis:
> Good catch. Could you re-target & rebase it to `3.0`? (or `3.1` since 3.0 is past EOM - up to you)

@alongosz Let say 3.0 in case there is a last 3.0.7. There is no conflict with 3.0 nor 3.1 so it should be easy to propagate upward.

adriendupuis:
> > Good catch. Could you re-target & rebase it to `3.0`? (or `3.1` since 3.0 is past EOM - up to you)
> 
> @alongosz Let say 3.0 in case there is a last 3.0.7. There is no conflict with 3.0 nor 3.1 so it should be easy to propagate upward.

No, there is some files that want to come with. I got to do the rebase. Sorry for the too early ping.

adriendupuis:
> > > Good catch. Could you re-target & rebase it to `3.0`? (or `3.1` since 3.0 is past EOM - up to you)
> > 
> > 
> > @alongosz Let say 3.0 in case there is a last 3.0.7. There is no conflict with 3.0 nor 3.1 so it should be easy to propagate upward.
> 
> No, there is some files that want to come with. I got to do the rebase. Sorry for the too early ping.

@alongosz Rebased on 3.0.

adriendupuis:
@alongosz I properly rebased this PR on 3.1 this time.

lserwatka:
@adamwojs could you merge it up?

PR #587, opened 2020-07-27
Switched base-dev.yml docker-compose configuration for app container to -dev image

kaff:
For better DX I switched base-dev.yml docker-compose configuration for app container to -dev image (which includes Xdebug)

### DIFF
--- a/doc/docker/base-dev-debug.yml
+++ b/doc/docker/base-dev-debug.yml
@@ -1,0 +1,8 @@
+version: '3.3'
+# Single server setup for dev
+
+services:
+  app:
+    image: ${PHP_IMAGE_DEV}
+    environment:
+      - XDEBUG_CONFIG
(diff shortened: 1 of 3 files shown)


kaff:
According to @mnocon suggestion and I switched back base-dev.yaml to the previous image and added new configuration file base-dev-debug.yaml where the image is switched to -dev

wizhippo:
Can we bump the compose version to '3.4'?  It allows for yaml x-properties that helps reduce repeating config.

doc/docker/base-dev-debug.yml
```diff
@@ -0,0 +1,8 @@
+version: '3.3'
+# Single server setup for dev
+
+services:
+  app:
+    image: ${PHP_IMAGE_DEV}
+    environment:
+      - XDEBUG_CONFIG
```
ViniTou, line 8: Please, add new line at end of the file to be consistent.

wizhippo: All env var indentation.  Most of them are in line with the label

```
  app:
    image: ${PHP_IMAGE_DEV}
    environment:
    - XDEBUG_CONFIG
```

Would be nice if they were all consistent too.

mnocon:
Please wait until https://github.com/ezsystems/docker-php/pull/52 is finished - we're adding xdebug available by default to all images (but enabled with a env variable), which might solve this usecase.

doc/docker/base-dev-debug.yml
wizhippo: Also with recent changes 

- ENABLE_XDEBUG

PR #590, opened 2020-08-03
[Docker] Switched image format and added PostgreSQL service

mnocon:
JIRA: https://jira.ez.no/browse/EZP-31620

This PR is the result of work done recently in docker-php, mostly https://github.com/ezsystems/docker-php/pull/52.

I have:
- switched the default PHP_IMAGE to `ezsystems/php:7.4-v2-node10` and dropped `PHP_IMAGE_DEV`, it's no longer used in our Docker stack
- added `ENABLE_XDEBUG` and `XDEBUG_CONFIG` variables to base-dev and base-prod (which should make  https://github.com/ezsystems/ezplatform/pull/587 not needed anymore - xdebug is present in all images now)
- added a new PostgreSQL Docker service
- added a job running Behat tests on PostgreSQL (requires https://github.com/ezsystems/ezplatform-admin-ui/pull/1445 to pass)

Passing PostgreSQL build: https://travis-ci.org/github/ezsystems/ezplatform/builds/715567661

### DIFF
--- a/.env
+++ b/.env
@@ -10,8 +10,7 @@ DATABASE_PASSWORD=SetYourOwnPassword
 DATABASE_NAME=ezp
 
 ## Docker images (name and version)
-PHP_IMAGE=ezsystems/php:7.4-v1-node
-PHP_IMAGE_DEV=ezsystems/php:7.4-v1-dev
+PHP_IMAGE=ezsystems/php:7.3-v2-node10
 NGINX_IMAGE=nginx:stable
 MYSQL_IMAGE=healthcheck/mariadb
 SELENIUM_IMAGE=selenium/standalone-chrome-debug:3.141.59-20200326
--- a/doc/docker/base-dev.yml
+++ b/doc/docker/base-dev.yml
@@ -11,6 +11,7 @@ services:
      - db
     environment:
      - COMPOSER_MEMORY_LIMIT
+     - ENABLE_XDEBUG
      - EZPLATFORM_SITEACCESS
      - PHP_INI_ENV_memory_limit
      - SYMFONY_ENV=${SYMFONY_ENV-dev}
@@ -25,6 +26,7 @@ services:
      - RECOMMENDATIONS_LICENSE_KEY
      - PUBLIC_SERVER_URI
      - HTTPCACHE_VARNISH_INVALIDATE_TOKEN
+     - XDEBUG_CONFIG
     networks:
      - backend
 
--- a/doc/docker/import-dataset.yml
+++ b/doc/docker/import-dataset.yml
@@ -13,8 +13,7 @@ services:
           nocopy: false
 
   import-db:
-    # Using dev for now, because we need mysql client
-    image: ${PHP_IMAGE_DEV}
+    image: ${PHP_IMAGE}
     environment:
      - DATABASE_USER
      - DATABASE_PASSWORD
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,11 +59,11 @@ matrix:
         - SETUP_BEHAT_OPTS="--profile=setup --suite=personas --tags=@setup"
         - BEHAT_OPTS="--profile=adminui --suite=personas"
     - if: type in (cron, api)
-      name: "[PHP 7.1] Admin UI on Clean Platform"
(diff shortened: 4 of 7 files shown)


doc/docker/import-dataset.yml
```diff
@@ -13,8 +13,7 @@ services:
           nocopy: false
 
   import-db:
-    # Using dev for now, because we need mysql client
-    image: ${PHP_IMAGE_DEV}
```
mnocon, line 17: `dev` image has been merged with `prod`, so the mysql client is present in PHP_IMAGE

.env
```diff
@@ -10,8 +10,7 @@ DATABASE_PASSWORD=SetYourOwnPassword
 DATABASE_NAME=ezp
 
 ## Docker images (name and version)
-PHP_IMAGE=ezsystems/php:7.4-v1-node
```
ViniTou, line 13: I am still not sure if we should use 7.4 as default and not the lowest supported version.

andrerom: For defaults we should pick what we recommend, needs for test matrix to make sure we do pass on lowest and highest supported version should be set using ENVs in CI system IMO

doc/docker/base-dev.yml
```diff
@@ -11,6 +11,7 @@ services:
      - db
     environment:
      - COMPOSER_MEMORY_LIMIT
+     - ENABLE_XDEBUG
```
andrerom, line 14:
```suggestion
     - ENABLE_XDEBUG=${ENABLE_XDEBUG-1}
```

doc/docker/base-prod.yml
```diff
@@ -12,6 +12,7 @@ services:
      - db
     environment:
      - COMPOSER_MEMORY_LIMIT
+     - ENABLE_XDEBUG
```
andrerom, line 15: Is there any use cases for us to enable xdebug on prod image? That is kind of what dev is for now

andrerom:
Besides inline comments `db-stack.yml` is using hardcoded mariadb version, unsure why.

vidarl:
I think the reason for that was simply that I only tested docker stack on mariadb... But feel free to generalize it so it can be used with both, or rename it `db-stack-mariadb.yml` and make a `db-stack-postgreql.yml` counterpart if you think that is better

doc/docker/base-dev.yml
wizhippo: I would prefer not.  I usually have it off by default unless I need to dig in deep as the performance is so slow with it enabled especially if dealing with anything that makes the container rebuild.

andrerom: But now that we have arranged this differently, when do you need to use base-dev?

wizhippo: Unless I do not see the change the prod builds the image and the dev mounts the current directory.

wizhippo: So for development I would like to use base-dev that mounts the current workspace without xdebug enabled by default as in most cases the symfony debug gives enough detail to solve most issues without the speed penalty of xdebug.

andrerom: > Unless I do not see the change the prod builds the image and the dev mounts the current directory instead.

true, good point.

.env
mnocon: I've changed it to use PHP7.3 by defult (recommended for Platform 2.5). Travis will run on PHP7.1 and 7.4 for testing.

doc/docker/base-prod.yml
mnocon: Initially I've decided to add it because it's disabled by default anyway and thought that it might be useufl for some people, but given that it's meant for prod maybe we shouldn't encourage it.

I've removed it.

mnocon:
I've decided to keep db-stack as it is, as I haven't used Docker Stack yet and I'm not sure how to approach it (and renaming it to `db-stack-mariadb.yml` without creating a PostgreSQL alternative would just break existing scripts without any benefit for people).

This PR in its current state allows us to run tests on PostgreSQL on Travis, which was my initial goal - IMHO we can work on adding more PostgreSQL support in the future if needed.

PR #591, opened 2020-08-19
EZP-31807: Updated platform.sh command example to set COMPOSER_AUTH

janit:
https://jira.ez.no/browse/EZP-31807

The example command for configuring Platform.sh COMPOSER_AUTH environment variable was no longer working with the latest version of the Platform.sh CLI. This PR includes an updated command that works with Platform.sh CLI 3.62.0.


andrerom:
@ezsystems/documentation-team Is this perhaps already covered by the doc so we can finally get rid of this here?

PR #482, opened 2019-11-25
EZP-31189: Reduce memory required for 1.7 and 1.13 branches to install

mnocon:
JIRA: https://jira.ez.no/browse/EZP-31189

Trying to unblock failing (https://travis-ci.org/ezsystems/ezplatform/builds/617477582) Travis tests by limiting memory required by Composer.

Merging: If this is approved I will prepare a PR to 1.13 with suggested changes and send it to review - if approved I will close that PR and apply the changes when merging.
These changes should not be applied for 2.5 and master.

PR #594, opened 2020-09-07
[Composer] Reduce memory required for branch 1.13 to install

mnocon:
Failing build: https://travis-ci.org/github/ezsystems/ezplatform/builds/723996895

Similar as in https://github.com/ezsystems/ezplatform/pull/482, limiting the minimum version of packages should make Composer's job easier. However, as you can see by the commits, the limiting of packages was not enough - I've decided to increase the available memory for Composer to 4GB as this solves the issue.


tomaszszopinski (approved):

micszo (approved):

bogusez (approved):

.env
```diff
@@ -4,7 +4,7 @@ COMPOSE_DIR=.
 
 # You'll need to adjust this for Windows and XDB Linux systems: https://getcomposer.org/doc/03-cli.md#composer-home
 COMPOSER_HOME=~/.composer
-COMPOSER_MEMORY_LIMIT=3G
+COMPOSER_MEMORY_LIMIT=4G
```
andrerom, line 7: Reduce? ;)

Composer 2.0 is in RC now, so maybe better to start to look into using it as it reduced memory use by a lot. _(usually around 1/3)_

It will anyway probably be on installed on travis and other places as soon as it is stable, so better to prepare and avoid everything from breaking over it then waiting for it to happen.

mnocon: Reduced to 3.2G, looks like it's enough.

Yep, I hope we will be able to get back to topic of Composer 2.0 once 3.2 is released, but this will be a larger task - I treat this PR as a quick fix to unblock the currently failing builds.

PR #602, opened 2020-10-02
EZP-31982: Revert cache setting in Encore config to default

Steveb-p:
JIRA: [EZP-31982](https://jira.ez.no/browse/EZP-31982)

This in practice disables Encore built-in cache on dev environments, following Symfony's default configuration in this matter.

Fixes an issue when adding new entrypoints to webpack.config.js that were not recognized until cache was cleared - which lead
to confusing errors occuring in template files that included those entrypoints.

### DIFF
--- a/config/packages/webpack_encore.yaml
+++ b/config/packages/webpack_encore.yaml
@@ -10,4 +10,4 @@ webpack_encore:
 
     # Cache the entrypoints.json (rebuild Symfony's cache when entrypoints.json changes)
     # Available in version 1.2
-    cache: '%kernel.debug%'
+    # cache: true


alongosz (commented):
Hmm, if Symfony default is "!kernel.debug", then this is indeed fine.

One Developer enablement request from me: we need commits that are being merged to be formed in past simple, since they're about to become  a _history_. To simplify the process for a Maintainer, we're fine with forming a Pull Request title in the past as well, so a Maintainer can merge it as-is.

I know this is not the mainstream approach, just trying to keep consistency within the Product.

adamwojs (approved):

Steveb-p:
@alongosz agreed.

To clarify, Symfony does not set that config value for environments other than `prod` in their template. Then the extension treats it as `false` by default:
```php
->booleanNode('cache')
    ->info('Enable caching of the entry point file(s)')
    ->defaultFalse()
->end()
```

In production config files it is set to `true`, which is the case for us as well.

lserwatka (approved):

lserwatka:
You can merge it up.

adamwojs:
Done.

dabrt:
@Steveb-p 
Hi Paweł, 
when creating a pull request, please always add a table that:
- lists a Jira ticket that the commit originates from, 
- lists branches that the commit is intended for, and 
- provides a short description, incl. documentation input, if needed.

PR #603, opened 2020-10-05
Bumped libraries that are not compatible with Symfony 5

Steveb-p:
Otherwise those bundles caused container to compile successfully due to reliance on obsolete kernel parameters.

This can be checked by running
```bash
composer install --prefer-lowest
```


lserwatka:
You can switch to 3.1 branch.

Steveb-p:
@lserwatka switched over to 3.1 as base.

lserwatka (approved):

PR #605, opened 2020-10-07
EZP-31875: Added firewall rules for JWT auth

bdunogier:
Required by:
- https://github.com/ezsystems/ezplatform-graphql/pull/86
- https://github.com/ezsystems/ezplatform-rest/pull/54

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -44,6 +44,26 @@ security:
             pattern: /user/(forgot-password|reset-password)
             security: false
 
+        ezpatform_rest:
+            request_matcher: EzSystems\EzPlatformAdminUi\REST\Security\NonAdminRESTRequestMatcher
+            user_checker: eZ\Publish\Core\MVC\Symfony\Security\UserChecker
+            anonymous: ~
+            guard:
+                authenticators:
+                    - lexik_jwt_authentication.jwt_token_authenticator
+                entry_point: lexik_jwt_authentication.jwt_token_authenticator
+            stateless: true
+
+        ezpatform_graphql:
+            request_matcher: EzSystems\EzPlatformGraphQL\Security\NonAdminGraphQLRequestMatcher
+            user_checker: eZ\Publish\Core\MVC\Symfony\Security\UserChecker
+            anonymous: ~
+            guard:
+                authenticators:
+                    - lexik_jwt_authentication.jwt_token_authenticator
+                entry_point: lexik_jwt_authentication.jwt_token_authenticator
+            stateless: true
+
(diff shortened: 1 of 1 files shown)


adamwojs (commented):
Hi! Thanks for PR, but IMHO additional firewall should be enabled by user only if he/she really like to use JWT (as usually in case of security features)

lserwatka:
@mnocon Travis is filing here.

mnocon:
@lserwatka REST functional tests are failing with errors such as:
```
1) EzSystems\EzPlatformRestBundle\Tests\Functional\BookmarkTest::testCreateBookmark

Server error message (401): Unauthorized

The User does not have the 'create' 'content' permission with: parentLocationId '2', sectionId '1'
```
So probably related to this PR, maybe they should be adapted somehow when JWT is used, not sure.

bdunogier:
> Hi! Thanks for PR, but IMHO additional firewall should be enabled by user only if he/she really like to use JWT (as usually in case of security features)

Maybe add them commented out then ?

> REST functional tests are failing
Hmm. I think it is because we rely on basic auth for functional tests.
@kmadejski should we add exclusion rules for the matcher when the request includes basic auth / Bearer token (and maybe session ?) auth ?

kmadejski:
I second @adamwojs here. My idea from the beginning was that this is opt-in.

> Hmm. I think it is because we rely on basic auth for functional tests.
@kmadejski should we add exclusion rules for the matcher when the request includes basic auth / Bearer token (and maybe session ?) auth ?

And this is exactly one of the reasons why I want it as an opt-in, at least for now. If you know that you want to use it, then you enable it. I can hardly imagine that you would like to have both JWT & Session & maybe even basic auth enabled in the same time. 

Maybe for the future, we should consider JWT to be a first-choice for us, but as you know it would involve some changes in AdminUI and so on...

adamwojs:
> Maybe add them commented out then ?

Yes, It's good idea.

bdunogier:
Changed to comments, with a short line of explanation. Also fixed the typo, because `patform` is cute, not not very professional.

@mnocon talking about REST integration tests, do you think we could add one specific to JWT auth ? That would maybe enable JWT auth, login, and run a restricted API call ?

lserwatka (approved):

adamwojs (approved):

micszo (approved):

lserwatka:
@mnocon could you merge it sideways?

mnocon:
@lserwatka sure, merged
@bdunogier that's a good idea, I'll try a have a look at it soon

PR #609, opened 2020-10-09
EZP-32017: fix trusted proxy setup for platform.sh

ITernovtsiy:
platform.sh use "ngx_http_realip_module" nginx module, and put client IP into REMOTE_ADDR.
This makes it impossible to configure an application to be accessible from both with and without varnish. For example, serve a website by varnish and admin by Nginx.

If you'll put REMOTE_ADDR in TRUSTED_PROXIES, admin panel will be open to IP address spoof using "x-forwarded-for" header.
And, the user will be able to request your varnish invalidation token.

If you'll put other IP address, or don't specify TRUSTED_PROXIES at all, you'll receive "Unauthorized" from varnish (because of "/_ez_http_invalidatetoken" route checking "$request->isFromTrustedProxy()", which will return false).

Ideally, it should work in a way where the developer can safely use

$request->getClientIp()
I'm not sure if this issue is reproducible on the EZ platform cloud, but I assume it should be there too.

related to https://github.com/symfony/symfony/issues/26006

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -13,7 +13,30 @@ if ($_SERVER['APP_DEBUG']) {
     Debug::enable();
 }
 
-if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
+$trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? '';
+
+// specific for platform.sh deployed to both with and without varnish access, when you should not trust REMOTE_ADDR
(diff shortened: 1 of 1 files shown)


andrerom:
This is not the right place to fix this, ideally maybe https://github.com/platformsh/symfonyflex-bridge

> This makes it impossible to configure an application to be accessible from both with and without varnish

But any reason why admin should omit varnish here? Admin will disable cache of responses

ITernovtsiy:
@andrerom thanks, missed that package, closing this one

ITernovtsiy:
replaced by https://github.com/ezsystems/ezplatform-http-cache/pull/136 
which looks a bit better, as platform.sh designed to emulate state with the app is not behind a proxy.

PR #621, opened 2020-11-04
[Tests] EZP-32103: Adjusted Behat config to behave more like prod

mnocon:
JIRA: https://issues.ibexa.co/browse/EZP-32103

~~Contains https://github.com/ezsystems/ezplatform/pull/623 and temp changes, needs rebase before merging.~~

Things done:
1) I've removed the default config from FriendsOfBehat\SymfonyExtension

We're using the default values for the kernel class and path. The env and debug value is taken from the loaded bootstrap file, which loads environmental variables - there's no change in behaviour if APP_ENV=behat, APP_DEBUG=1 is specified (which we do in all our builds) and it's not harcoded there in case someone wants to change it.

2) I've adjusted `behat` env configuration to be more prod-like: right now it contains all the files from `prod` env, reasoning: our tests should be running in a similar env as the production one (but with Debug enabled, so that errors are not silenced). I've also enabled strict_requirements for routing: https://symfony.com/doc/current/reference/configuration/framework.html#strict-requirements to make sure we're not silencing any errors. I had to do it because HTTP cache behaves in a weird way when `framework.test` setting is enabled - IMHO it's a good opportunity to decide how the `behat` env should behave.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -30,12 +30,7 @@ default:
 
         DMore\ChromeExtension\Behat\ServiceContainer\ChromeExtension: ~
 
-        FriendsOfBehat\SymfonyExtension:
-            kernel:
-                path: src/Kernel.php
-                class: App\Kernel
-                environment: behat
-                debug: true
+        FriendsOfBehat\SymfonyExtension: ~
 
         EzSystems\BehatBundle\BehatExtension: ~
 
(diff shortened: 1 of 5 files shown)


micszo (approved):
I agree with the direction for `behat` env - prod-like with errors exposed.

bogusez (approved):

PR #627, opened 2020-12-08
EZP-31594: Replaced bin/console with original file from symfony/console recipe

webhdx:
> JIRA: https://issues.ibexa.co/browse/EZP-31594

This is complementary PR to https://github.com/ezsystems/ezplatform-kernel/pull/145 which removes custom `Application` class handling `--siteaccess` option. It's no longer needed because the option is now handled by CompilerPass and EventListener.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -1,18 +1,22 @@
 #!/usr/bin/env php
 <?php
+
 use App\Kernel;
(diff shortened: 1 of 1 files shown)


webhdx:
Tests do not pass because it's missing ezplatform-kernel dependency

PR #631, opened 2020-12-10
Revert "EZP-32214: Refactored config/packages/overrides/* as Compiler Passes"

webhdx:
Reverts ezsystems/ezplatform#626

Original solution did not work well with HTTP Cache. Reverting for now and will provide proper fix soon.


lserwatka (approved):

mnocon (approved):
Green build: https://travis-ci.com/github/ezsystems/ezplatform-http-cache/builds/207810844

PR #633, opened 2020-12-16
EZP-32242: Improve default platform.sh config for common usecases

andrerom:
Includes:
- PHP ini tweaks for performance and higher default upload limit
- Log rotation to make sure log does not consume gigabytes of disk space
- Asset caching
- Monthly HTTPS cert renewable in cron before it expires
- Redis config tweaks
- smaller tweaks to inline doc

Log rotation is done by tweaking monolog settings, so this will also benefit non p.sh setups (incl local).

Todo on merge up:
- ~For 3.x add PHP preloading config~ _(will need to start to use separate build dir for that as server starts before deploy hook runs so as long as this is in a mount it is out of date or missing prior to deploy hook runs and clears cache)_

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -32,7 +32,7 @@ doctrine:
                 server_version: '%database_version%'
     orm:
         auto_generate_proxy_classes: '%kernel.debug%'
-        naming_strategy: doctrine.orm.naming_strategy.underscore
+        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
         auto_mapping: true
 
 # If you are not using MySQL, you can comment-out this section
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -37,9 +37,24 @@ relationships:
 
 variables:
     php:
-        # Example of setting php.ini config
+        # Some recommended defaults for php.ini settings, adapt for your project
         #"display_errors": "On"
         memory_limit: 512M
+        post_max_size: 256M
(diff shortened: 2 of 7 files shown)


app/config/config.yml
```diff
@@ -32,7 +32,7 @@ doctrine:
                 server_version: '%database_version%'
     orm:
         auto_generate_proxy_classes: '%kernel.debug%'
-        naming_strategy: doctrine.orm.naming_strategy.underscore
+        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
```
andrerom, line 35: This is just to avoid deprecation warning on using the old setting name afaik (source @mateuszbieniek).

webhdx:
Not sure if this is going to work for flex. Once this gets merged up it needs to be ported to https://github.com/ibexa/recipes. WIth Flex we shouldn't really alter configuration for other packages. At least not by modifying their config files like `monolog.yml`. We should change these settings from our own config file i.e. as a part of `ezplatform.yaml` (ref. https://github.com/ibexa/recipes/blob/master/ibexa/oss/3.3/config/packages/ezplatform.yaml#L91) but I'm not sure if everything could be done this way.

andrerom:
I won't have chance to look at how this should be approached for flex, it indeed needs to be done differently there. I focused on common pitfalls / tweaks to default config for currently supported releases here for Cloud.

andrerom:
I'm assuming this is no longer relevant ;)

PR #639, opened 2020-12-28
[Platform.sh] Update .platform.app.yaml

damianz5:
[Platform.sh] Update url to env variables for composer


lserwatka:
You can merge it sideways

PR #646, opened 2021-01-22
EZEE-3453: Added SQL update scripts

ciastektk:
Related to: https://github.com/ezsystems/ezplatform-kernel/pull/162


tomaszszopinski (approved):
QA approved on Ibexa Commerce 3.2.4-dev with diff & patch.

alongosz (changes requested):
What if someone already had changed e-mails of system users? This upgrade script should be concerned only with a result of clean installation, thus `WHERE email = 'nospam@ez.no`'; Otherwise this would overwrite custom user data.

alongosz (approved):

PR #661, opened 2021-06-22
[Travis] Added logging into Docker Hub

mnocon:
Follow up to https://github.com/ezsystems/ezplatform/pull/644, which was done only for v2.5 or higher (but never for 1.13).

This change is needed in two places, as for external repos `bin/.travis/trusty/setup_from_external_repo.sh` is called, but for metarepo Docker is invoked directly.

Example failures:
https://travis-ci.com/github/ezsystems/ezplatform/jobs/517257672
https://travis-ci.com/github/ezsystems/ezpublish-kernel/jobs/517204307


micszo (approved):
Thanks!

piotrbisaga (approved):

tomaszszopinski (approved):

bogusez (approved):

PR #662, opened 2021-06-24
[Composer] Added conflict with lexik/jwt-authentication-bundle 2.12.0

mnocon:
Travis failure: https://travis-ci.com/github/ezsystems/ezplatform-page-builder/jobs/518410065#L2576

```
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 255
!!  PHP Fatal error:  Class EzSystems\EzPlatformPageBuilder\Security\EditorialMode\TokenManager contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface::decodeFromJsonWebToken) in /var/www/vendor/ezsystems/ezplatform-page-builder/src/lib/Security/EditorialMode/TokenManager.php on line 24
!!  Symfony\Component\ErrorHandler\Error\FatalError {#10767
!!    -error: array:4 [
!!      "type" => 1
!!      "message" => "Class EzSystems\EzPlatformPageBuilder\Security\EditorialMode\TokenManager contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface::decodeFromJsonWebToken)"
!!      "file" => "/var/www/vendor/ezsystems/ezplatform-page-builder/src/lib/Security/EditorialMode/TokenManager.php"
!!      "line" => 24
!!    ]
!!    #message: "Error: Class EzSystems\EzPlatformPageBuilder\Security\EditorialMode\TokenManager contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface::decodeFromJsonWebToken)"
!!    #code: 0
!!    #file: "./vendor/ezsystems/ezplatform-page-builder/src/lib/Security/EditorialMode/TokenManager.php"
!!    #line: 24
!!  }
!!  
Script @auto-scripts was called via post-install-cmd
```

It's caused by the latest release in LexikJWTAuthenticationBundle, please see https://github.com/lexik/LexikJWTAuthenticationBundle/pull/872#issuecomment-866646057 for more information.

In theory we could only limit the dependency for ezplatform-ee, as this is the project where it breaks for us, but I think it's better to use the same bundle version across editions to avoid complications.

PR #663, opened 2021-06-27
Remove conflict rule for lexik/jwt-authentication-bundle

chalasr:
Reverts #662. The BC break has been fixed in 2.12.1.
Note that you may want to implement the new `JWTTokenManagerInterface#parse()` method in your `TokenManager` implementation as it will be mandatory in 3.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,7 +90,6 @@
     },
     "conflict": {
         "doctrine/persistence": "1.3.2",
-        "lexik/jwt-authentication-bundle": "2.12.0",
         "symfony/framework-bundle": "5.1.0",
         "symfony/symfony": "*"
     },


adamwojs (approved):

mikadamczyk (approved):

ViniTou (approved):

ciastektk (approved):

webhdx (approved):

Steveb-p:
I'm unsure if this removal is necessary - we are targeting a specific version, since a fix was expected. This conflict entry will allow `2.12.1` to be installed normally, while preventing `2.12.0`.

chalasr:
Given that composer always install the latest version when a range is required (or lowest with —prefer-lowest) and that requiring a fixed version is not common (if not discouraged), it’s safe to remove it.
Maintaining such a conflict rule is not necessary.

Steveb-p (approved):
We're not doing lowest dependency check (yet), so :+1:.

chalasr:
Is there anything else I can do to help moving forward here?

mnocon (approved):
Thank you for your PR!

PR #660, opened 2021-06-18
[Composer] Added conflict with laminas/laminas-code:4.4.0

mnocon:
Failing build:
https://travis-ci.com/github/ezsystems/ezplatform/builds/229620437

Error:
```
  Compile Error: Could not check compatibility between Symfony\Bridge\ProxyMa  
  nager\LazyProxy\PhpDumper\LazyLoadingValueHolderGenerator::generate(Reflect  
  ionClass $originalClass, Zend\Code\Generator\ClassGenerator $classGenerator  
  ) and ProxyManager\ProxyGenerator\LazyLoadingValueHolderGenerator::generate  
  (ReflectionClass $originalClass, Laminas\Code\Generator\ClassGenerator $cla  
  ssGenerator), because class Zend\Code\Generator\ClassGenerator is not avail  
  able    
```

Steps to reproduce:
`composer create-project ezsystems/ezplatform:^2.5@dev`

I've used PHP 7.4 and suspect that not all supported PHP versions are affects (see green build on PHP 7.1 in https://travis-ci.com/github/ezsystems/ezplatform/builds/229620437)

I've found that the issue is caused by the release of `laminas/laminas-code` 4.4.0 (released yesterday):
https://github.com/laminas/laminas-code/releases/tag/4.4.0

Installing this package version removes two other packages:
```
MacBook-Pro:test_laminas mareknocon$ composer require laminas/laminas-code:~4.4.0
./composer.json has been updated
Running composer update laminas/laminas-code
Loading composer repositories with package information
Updating dependencies
Lock file operations: 0 installs, 1 update, 2 removals
  - Removing laminas/laminas-eventmanager (3.3.1)
  - Removing laminas/laminas-zendframework-bridge (1.2.0)
  - Upgrading laminas/laminas-code (4.3.0 => 4.4.0)
```

Based on the error message it looks like the error is caused by the lack of laminas-zendframework-bridge (https://github.com/laminas/laminas-zendframework-bridge/tree/1.3.x/src) which is adding some autoloading BC layer:
https://github.com/laminas/laminas-zendframework-bridge/blob/1.3.x/src/Autoloader.php#L117

The `Zend` namespace usage seems to come from symfony/proxy-manager-bridge:
https://github.com/symfony/proxy-manager-bridge/blob/3.4/LazyProxy/PhpDumper/LazyLoadingValueHolderGenerator.php#L15

needed by ez-support-tools:
```
[...]

PR #665, opened 2021-07-12
[Composer] Added conflict with laminas/laminas-code:^4.4

mnocon:
For a bigger picture please see first this PR:
https://github.com/ezsystems/ezplatform/pull/660

And the most important part:
```
I've chosen to add a conflict as it's the quickest way to unblock Travis (and installation of development version of our product) - if you see a better way based on the data I've included here please let me know, I know that it will start failing as soon as laminas/laminas-code 4.4.1 is released.
```
which happened this week:
https://github.com/laminas/laminas-code/releases

Given that this issue won't be solved upstream, as indicated in the disscusions:
https://github.com/laminas/laminas-code/issues/89
https://github.com/symfony/symfony/issues/41742

I think there's nothing we can do except for a conflict with all laminas/laminas-code in 4.4.x series and higher.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -83,7 +83,7 @@
     "conflict": {
         "doctrine/dbal": "2.7.0",
         "doctrine/persistence": "1.3.2",
-        "laminas/laminas-code": "4.4.0",
+        "laminas/laminas-code": "^4.4",
         "symfony/symfony": "3.4.9||3.4.12||3.4.16",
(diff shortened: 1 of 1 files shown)


alongosz (approved):
Apart from forking Symfony I don't have any other solution ATM

micszo (approved):
Thanks!

PR #674, opened 2021-11-22
[CI] Added workaround for newer Docker Compose versions

mnocon:
Similar as in https://github.com/ibexa/ci-scripts/pull/23 , but for v2.

Tested in https://github.com/ezsystems/ezplatform-http-cache/pull/162


micszo (approved):
👍🏻

bogusez (approved):

PR #679, opened 2022-03-01
EZP-32437: EZP2.5 can't be installed on node 10

GrabowskiM:
JIRA ref: https://issues.ibexa.co/browse/EZP-32437

Force to use older version of `sockjs-client` as newest release require node12


dew326 (approved):

mnocon (approved):
Tested with Node10, 12 and 14 - installs correctly now

PR #683, opened 2022-07-13
IBX-2921: Disable TRACE/TRACK

glye:
Ref https://issues.ibexa.co/browse/IBX-2921

Disable TRACE/TRACK methods. Browsers have blocked this already for ages now, but explicitly blocking it in vhosts can silence some test warnings. Proposal by @vidarl, I just added case insensitivity.

### DIFF
--- a/doc/apache2/vhost.template
+++ b/doc/apache2/vhost.template
@@ -74,7 +74,7 @@
         RewriteEngine On
 
         # Make sure TRACE and TRACK methods are denied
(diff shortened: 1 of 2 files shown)


glye:
I don't see any way the code can cause the test failures (unless we actually use TRACE/TRACK in tests). Ready for review.

alongosz (approved):

konradoboza (approved):

adamwojs (approved):

PR #691, opened 2023-08-31
IBX-6383: Remove Allure reports integration from CI

micszo:
**JIRA**: [IBX-6383](https://issues.ibexa.co/browse/IBX-6383)

The goal of the task is to remove the whole Allure reports integration from our test code in order to shut the server hosting it down.

In case of v2.5 the PR also contains a back-port of an update of the docker file for Varnish (for the sake of a green CI build).


mnocon (approved):
Thanks! 🙇

bogusez (approved):

